PR #43, opened 2018-01-23
Clean stuff

loeck:


src/main/app.js
```diff
@@ -2,6 +2,8 @@
 
 import { app, ipcMain, BrowserWindow } from 'electron'
 
+process.setMaxListeners(Infinity)
```
meriadec: we may instead put a high value but not infinity, like `100`

PR #46, opened 2018-01-23
Store lastAdress and lastIndex for better sync

loeck:


package.json
```diff
@@ -20,7 +20,11 @@
     "storybook": "start-storybook -p 4444"
   },
   "lint-staged": {
-    "*.js": ["eslint --fix", "prettier --write", "git add"]
+    "*.js": [
```
meriadec: we need to avoid prettifying package.json in our editors, or we are gonna be stuck in an infinite commited lines for nothing

src/components/AccountPage.js
```diff
@@ -39,11 +39,11 @@ class AccountPage extends PureComponent<Props> {
                 <Card title="Balance">{format(accountData.balance)}</Card>
               </Box>
               <Box flex={1}>
-                <Card title="Receive" />
+                <Card title="Receive">{accountData.address}</Card>
               </Box>
             </Box>
             <Card title="Last operations">
-              {accountData.transactions.map(tr => (
+              {accountData.transactions.reverse().map(tr => (
```
meriadec: maybe the reverse should be done earlier

PR #110, opened 2018-02-13
Clean Dev Tools, add main process in CPU Usage

loeck:


src/main/app.js
```diff
@@ -154,6 +153,9 @@ function createDevWindow() {
     window.show()
   })
 
+  // Don't want to use HTML <title>
+  window.on('page-title-updated', e => e.preventDefault())
```
meriadec: :+1: :joy:

PR #198, opened 2018-03-07
Fix build size

loeck:


meriadec:
We may want additional investigations of what happened in the mac build. Ignoring modules one by one don't sound hardly maintainable.

PR #224, opened 2018-03-14
Update storybook

meriadec:


meriadec:
https://circleci.com/gh/meriadec/ledger-wallet-desktop/118

Not the first problem we got with flow-typed suddenly reporting errors because of its non-versioned nature.

To illustrate:

flow-typed


:smile:

PR #225, opened 2018-03-14
qrcode export

gre:
export all accounts via qrcodes.

### DIFF
--- a/src/components/QRCodeExporter.js
+++ b/src/components/QRCodeExporter.js
@@ -1,0 +1,85 @@
+// @flow
+
+import React, { PureComponent } from 'react'
+
+import { connect } from 'react-redux'
+import type { State } from 'reducers'
+import { getVisibleAccounts } from 'reducers/accounts'
+import QRCode from './base/QRCode'
+
+// encode the app state to export into an array of chunks for the mobile app to understand.
+// returned data frames are json stringified array with format: [ datalength, index, type, ...rest ]
+// NB as soon as we have common types we'll move this in a ledgerhq/common project
+function makeChunks(state: State): Array<string> {
+  const chunksFormatVersion = 1
+  const desktopVersion = __APP_VERSION__
+  const data = [
+    ['meta', chunksFormatVersion, 'desktop', desktopVersion],
+    ...getVisibleAccounts(state).map(account => [
+      'account',
+      account.id,
+      account.name,
+      account.coinType,
+    ]),
+  ]
+  return data.map((arr, i) => JSON.stringify([data.length, i, ...arr]))
+}
+
+const mapStateToProps = (state: State) => ({ chunks: makeChunks(state) })
+
+class QRCodeExporter extends PureComponent<
+  {
+    chunks: string[],
+    fps: number,
+    size: number,
+  },
+  {
+    frame: number,
+  },
+> {
+  static defaultProps = {
+    fps: 10,
+    size: 480,
+  }
+
+  state = {
+    frame: 0,
+  }
+
+  componentDidMount() {
+    const nextFrame = ({ frame }, { chunks }) => ({
+      frame: (frame + 1) % chunks.length,
+    })
+    let lastT
+    const loop = t => {
+      this._raf = requestAnimationFrame(loop)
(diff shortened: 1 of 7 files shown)


src/components/base/QRCode/index.js
```diff
@@ -32,7 +33,8 @@ class QRCode extends PureComponent<Props> {
   }
 
   render() {
-    return <canvas ref={n => (this._canvas = n)} />
+    const { hidden } = this.props
+    return <canvas hidden={hidden} ref={n => (this._canvas = n)} />
```
meriadec: would be more DOM friendly to only update opacity style (that would require to "stack" all the qr codes in absolute)

src/components/SettingsPage/Tools.js
```diff
@@ -0,0 +1,60 @@
...
+      onClick={onClose}
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center',
+      }}
+    >
+      <p style={{ padding: '1em' }}>
+        Open Ledger Wallet Mobile App, go to <strong>Settings {'>'} Import Accounts</strong>
```
meriadec: let's add at least a `// TODO: translate`

src/components/SettingsPage/Tools.js
```diff
@@ -0,0 +1,60 @@
...
+  renderQRCodeModal = ({ onClose }: *) => (
+    <ModalBody
+      onClick={onClose}
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center',
+      }}
+    >
```
meriadec: `ModalBody` renders a `Box` so you can directly

```js
<ModalBody onClick={onClose} justify="center" align="center">
```

src/components/QRCodeExporter.js
```diff
@@ -0,0 +1,85 @@
...
+import { getVisibleAccounts } from 'reducers/accounts'
+import QRCode from './base/QRCode'
+
+// encode the app state to export into an array of chunks for the mobile app to understand.
+// returned data frames are json stringified array with format: [ datalength, index, type, ...rest ]
+// NB as soon as we have common types we'll move this in a ledgerhq/common project
+function makeChunks(state: State): Array<string> {
+  const chunksFormatVersion = 1
+  // $FlowFixMe
+  const desktopVersion = __APP_VERSION__
```
meriadec: see https://github.com/LedgerHQ/ledger-wallet-desktop/blob/master/flow-defs/globals.js

src/components/SettingsPage/Tools.js
gre: is there a eslint rule to disallow every in-place texts?

src/components/base/QRCode/index.js
gre: `hidden` is part of DOM right? not sure which is more efficient, opacity or hidden. position:absolute might help not having the browser do a position calc but I'm pretty sure it's premature optim. it would complexity the parent code to do absolute position

src/components/SettingsPage/Tools.js
gre: cool

src/components/SettingsPage/Tools.js
meriadec: hélas no

src/components/base/QRCode/index.js
meriadec: Yeah it's part of the DOM, that's not what I meant. But `hidden` trigger layout update while `opacity` trigger only a paint one. I would not underestimate that difference nor call it premature opti, mostly in that case where we will draw qrcode multiple time / s. Having a relative wrapper with fixed height and positionning qrcodes in absolute is not so complex compared to the perf gain.

relatable:
https://codepen.io/meriadec/pen/NYNrpY 
left square: hidden
right square: opacity

meriadec: Following our discussion, I will do the update myself after the merge. Can you rebase your branch from master?

PR #228, opened 2018-03-15
New Send Modal

meriadec:
send


src/components/modals/Send/Footer.js
```diff
@@ -0,0 +1,52 @@
...
+
+function Footer({ account, amount, t, onNext }: Props) {
+  return (
+    <ModalFooter horizontal align="center">
+      <Box grow>
+        <Label>{t('send:totalSpent')}</Label>
+        <Box horizontal flow={2} align="center">
+          <FormattedVal
+            color="dark"
+            val={amount.left * 10 ** account.unit.magnitude}
```
gre: amount.left should be directly in the correct (sat) value. (also, it's better to have a `Math.floor` of this)

src/components/modals/Send/Footer.js
```diff
@@ -0,0 +1,52 @@
...
+            unit={account.unit}
+            showCode
+          />
+          <Box horizontal align="center">
+            <Text ff="Rubik" fontSize={3}>
+              {'('}
+            </Text>
+            <FormattedVal color="grey" fontSize={3} val={amount.right} fiat="USD" showCode />
```
gre: IMO `fiat` shouldn't exist. should just use `unit={getFiatUnit("USD")}` for consistency
like we do in https://github.com/LedgerHQ/ledger-vault-front/blob/develop/src/components/CurrencyUnitValue.js

src/components/modals/Send/Footer.js
gre: basically idea is we should never have any float currencies, it's always normalized as an integer (in satoshis / cents). float should never exist in our data model (and only exists if we import data from some places that don't support integer style / format it back to a string)

PR #241, opened 2018-03-21
Add new design for OperationsList

loeck:


src/components/OperationsList/index.js
```diff
@@ -91,89 +78,124 @@ const ShowMore = styled(Box).attrs({
...
+
+const Address = ({ value }: { value: string }) => {
+  const addrSize = value.length / 2
+
+  const left = value.slice(0, 10)
+  const right = value.slice(-addrSize)
+  const middle = value.slice(10, -addrSize)
+
+  return (
+    <Box horizontal color="smoke" ff="Open Sans" fontSize={3}>
```
gre: should `Open Sans` be in a constant? same Q for `smoke` 🤔

src/components/OperationsList/index.js
```diff
@@ -91,89 +78,124 @@ const ShowMore = styled(Box).attrs({
...
+  t: T,
   tx: OperationType,
   withAccount?: boolean,
-  minConfirmations: number,
 }) => {
-  const acc = account || tx.account
+  const { unit } = account
   const time = moment(tx.receivedAt)
-  const Icon = getIconByCoinType(get(tx, 'account.currency.coinType'))
+  const Icon = getIconByCoinType(get(account, 'currency.coinType'))
```
gre: why not doing `account.currency.coinType` ? We should try to limit nb of usage of `get` to not lose the (flow) type

src/components/OperationsList/index.js
```diff
@@ -240,43 +272,40 @@ class OperationsList extends Component<Props> {
...
-                minConfirmations={minConfirmations}
-                tx={tx}
-              />
-            ))}
-          </Box>
+        <Box>
+          <Card flow={1} title={title} p={0}>
+            <Box>
+              {operations.map(tx => {
+                const acc = account || tx.account
```
gre: what is this `account || tx.account` ? 🤔

src/components/OperationsList/index.js
loeck: Actually, it's a constant :D
See: https://github.com/LedgerHQ/ledger-wallet-desktop/blob/master/src/styles/theme.js

src/components/OperationsList/index.js
loeck: You can have `account` in 2 ways, so, in operation object (for example, in Dashboard https://github.com/LedgerHQ/ledger-wallet-desktop/blob/master/src/components/DashboardPage/index.js#L67 when we want to display OperationsList with different accounts) and with props (in AccountPage (https://github.com/LedgerHQ/ledger-wallet-desktop/blob/master/src/components/AccountPage/index.js#L160 we pass directly props)

src/components/OperationsList/index.js
loeck: Yup, you're right, I fix that

src/components/OperationsList/index.js
gre: oh yeah I see. makes sense. I guess later we will have to change that if tx.account is gone. like `account || findInAccounts(accounts, tx.accountId)`,

src/components/OperationsList/index.js
gre: holy smoke :D 👍

PR #247, opened 2018-03-22
Fetch installable apps from api instead of currency list

meriadec:
So now we have complete list of installable apps.
Still improvements to do:

- use device firmware for the request instead of hardcoded 1.4 Nano S (see https://github.com/LedgerHQ/ledger-wallet-desktop/commit/1512ec6b234ec31f911e8d15fae0178eca0ca1bb#diff-f2529adb08f0ad05b8c5cf48c8a577f3R39)
- map applications with our icons (either api side or in our side)

2018-03-22_2018x984

### DIFF
--- a/src/components/ManagerPage/ManagerApp.js
+++ b/src/components/ManagerPage/ManagerApp.js
@@ -9,27 +9,39 @@ import Text from 'components/base/Text'
 const Container = styled(Box).attrs({
   align: 'center',
   justify: 'center',
-  m: 1,
+  m: 2,
 })`
   width: 150px;
   height: 150px;
-  background: rgba(0, 0, 0, 0.05);
+  background: white;
 `
+
+// https://api.ledgerwallet.com/update/assets/icons/bitcoin
 
 const ActionBtn = styled(Tabbable).attrs({
   fontSize: 3,
 })``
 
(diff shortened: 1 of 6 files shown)


src/components/ManagerPage/ManagerApp.js
```diff
@@ -2,40 +2,47 @@
...
 export default function ManagerApp(props: Props) {
-  const { currency, onInstall, onUninstall } = props
-  const Icon = getIconByCoinType(currency.coinType)
+  const { name, icon, onInstall, onUninstall } = props
+  const iconUrl = `https://api.ledgerwallet.com/update/assets/icons/${icon}`
   return (
     <Container flow={3}>
-      {Icon && <Icon size={24} />}
-      <Text>{currency.name}</Text>
+      <AppIcon src={iconUrl} />
```
gre: any reason why not making AppIcon accept the icon and move the src inside AppIcon comopnent?

meriadec: As `AppIcon` is only a styled.img, it forwards all its props to `img` tag.
If needed, yep we could change it to stateless component which will take icon as prop, but as the design is just a POC for now, I think we can keep it like that.

PR #250, opened 2018-03-27
Use types, decode, encode from wallet-common

meriadec:
:warning: [this pull request on wallet-common](https://github.com/LedgerHQ/ledger-wallet-common/pull/3) needs to be merged first. (edit: it's ok now)

**BONUS** you will have to remove your accounts.json file

### DIFF
--- a/src/components/OperationsList/stories.js
+++ b/src/components/OperationsList/stories.js
@@ -37,7 +37,7 @@ const operations = [
     address: '5c6ea1716520c7d6e038d36a3223faced3c',
     hash: '5c6ea1716520c7d6e038d36a3223faced3c4b8f7ffb69d9fb5bd527d562fdb62',
     amount: 130000000,
-    date: '2018-01-09T16:03:52Z',
+    date: new Date('2018-01-09T16:03:52Z'),
     confirmations: 1,
     account: account({
       name: 'Account 1',
@@ -47,7 +47,7 @@ const operations = [
     address: '5c6ea1716520c7d6e038d36a3223faced3c',
     hash: '5c6ea1716520c7d6e038d36a3223faced3c4b8f7ffb69d9fb5bd527d562fdb62',
     amount: 130000000,
-    date: '2018-01-09T16:03:52Z',
+    date: new Date('2018-01-09T16:03:52Z'),
     confirmations: 11,
     account: account({
       name: 'Account 1',
@@ -57,7 +57,7 @@ const operations = [
     address: '27416a48caab90fab053b507b8b6b9d4',
     hash: '27416a48caab90fab053b507b8b6b9d48fba75421d3bfdbae4b85f64024bc9c4',
     amount: -65000000,
-    date: '2018-01-09T16:02:40Z',
+    date: new Date('2018-01-09T16:02:40Z'),
     confirmations: 11,
     account: account({
       name: 'Account 2',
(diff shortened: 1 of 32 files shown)


src/components/OperationsList/stories.js
```diff
@@ -39,7 +37,7 @@ const operations = [
     address: '5c6ea1716520c7d6e038d36a3223faced3c',
     hash: '5c6ea1716520c7d6e038d36a3223faced3c4b8f7ffb69d9fb5bd527d562fdb62',
     amount: 130000000,
-    receivedAt: '2018-01-09T16:03:52Z',
+    date: '2018-01-09T16:03:52Z',
```
gre: should be `new Date(..`

src/helpers/btc.js
```diff
@@ -31,21 +30,25 @@ export function computeOperation(addresses: Array<string>) {
...
+      accountId: 'abcd',
+      blockHeight: 0,
     }
   }
 }
 
 export function getBalanceByDay(operations: Operation[]) {
   const txsByDate = groupBy(operations, tx => {
-    const [date] = new Date(tx.receivedAt).toISOString().split('T')
+    const [date] = new Date(tx.date).toISOString().split('T')
```
gre: tx.date is already a Date ?

src/reducers/accounts.js
```diff
@@ -1,42 +1,34 @@
...
 
 export type AccountsState = Account[]
 const state: AccountsState = []
 
+const accountModel = createAccountModel()
+
 function orderAccountsOperations(account: Account) {
   const { operations } = account
-  operations.sort((a, b) => new Date(b.receivedAt) - new Date(a.receivedAt))
+  operations.sort((a, b) => new Date(b.date) - new Date(a.date))
```
gre: same?

btw I think `orderAccountsOperations` can be dropped, we assume account.operations are always correctly sorted in all our current algorithm. maybe you just need to insert at the correct place when you add a new op?

src/reducers/accounts.js
```diff
@@ -98,31 +90,19 @@ export function canCreateAccount(state: State): boolean {
...
-    currency: getCurrencyByCoinType(account.coinType),
-    name: account.name || `${key}`,
-    unit: account.unit || getDefaultUnitByCoinType(account.coinType),
-  }))
+// Yeah. `any` should be `AccountRaw[]` but it can also be a map
+// of wrapped accounts. And as flow is apparently incapable of doing
+// such a simple thing, let's put any, right? I don't care.
+export function serializeAccounts(accounts: any): Account[] {
+  // ensure that accounts are always wrapped in data key
+  if (accounts.length && !accounts[0].data) {
```
gre: probably we could `try {} catch (){}` on the decode code & if it fails, fallback on returning `[]` , it's a safe thing to do if data happen to be corrupted in some way & so it does not brick the loader (i'm sure there are many ways things can go south & maybe we should assume the file we import can't be 100% trust. obviously we'll try to cover all cases via the migrations functions)

src/components/OperationsList/stories.js
meriadec: true

src/reducers/accounts.js
meriadec: good point, but let's move that input to a future commit, as it's not directly related to types changes

src/helpers/btc.js
gre: OH ok

src/reducers/accounts.js
gre: probably we will keep it like this and a bit later you can simplify this code to no longer support migrating old dev version ;)

PR #259, opened 2018-03-29
Various fixes on counter values & pal.

meriadec:
- unify InputCurrency, RequestAmount and CounterValue to make sure they always take satoshi like values, and returns the correct values. Internally they use our common helpers to do the conversion stuff
- instanciate a redux store for the whole storybook. this will highly simplify the integration of components inside it without tricks.
- fix counter value refreshing (reducer stuff)


src/components/CounterValue/index.js
```diff
@@ -43,10 +44,8 @@ export class CounterValue extends PureComponent<Props> {
       getCounterValue,
       ...props
     } = this.props
-
-    const date = moment(time).format('YYYY-MM-DD')
-    const v = getCounterValue(currency, counterValueCode)(value, date)
-
+    const date = new Date(moment(time).format('YYYY-MM-DD'))
```
gre: why not directly passing down date from props?

meriadec: Somewhere we pass it a string (didnt investigated deeply) and it was crashing :D I admit it's non optimal

PR #264, opened 2018-04-03
Rework Receive modal

loeck:


src/components/DeviceMonitNew/index.js
```diff
@@ -134,11 +134,12 @@ class DeviceMonit extends PureComponent<Props, State> {
   }
 
   render() {
+    const { coinType, account, devices, deviceSelected, render } = this.props
     const { status } = this.state
-    const { devices, deviceSelected, render } = this.props
 
     if (render) {
       return render({
+        coinType: (account && account.coinType) || coinType,
```
gre: - according to type, coinType is always here. not a `?number`
- coinType === 0 is actually legit, which is bitcoin. in that case `0 || coinType` might fallback to undefined if coinType not in props ?

src/components/modals/StepConnectDevice.js
```diff
@@ -2,28 +2,29 @@
...
+import type { Account, Currency } from '@ledgerhq/currencies/lib/types'
 import type { Device } from 'types/common'
 
 import DeviceConnect from 'components/DeviceConnect'
 import DeviceMonit from 'components/DeviceMonitNew'
 
 type Props = {
-  currency: Currency | null,
+  account?: Account,
+  currency?: Currency | null,
```
gre: the `T | null` pattern in Flow, should be replaced by `?T`

to be checked but I also think `currency?: Currency` is enough, no need to do `currency?: ?Currency`

src/components/modals/StepConnectDevice.js
```diff
@@ -34,3 +35,10 @@ export default (props: Props) => (
     )}
   />
 )
+
+StepConnectDevice.defaultProps = {
+  account: undefined,
+  currency: undefined,
```
gre: shouldn't define these because it's already the default.
if you do define these, then the prop is actually a 

```
account: ?Account
currency: ?Currency
```

if you don't, it's a 

```
account?: Account
currency?: Currency
```

which allows to not pass-in props and don't define this

src/components/DeviceMonitNew/index.js
loeck: Yep, but `account` in this context are not always here

loeck: In this context, you have account with coinType (for Send and Receive flow), or just coinType (for AddAccount flow)

src/components/modals/StepConnectDevice.js
loeck: Thanks for the tip !

src/components/DeviceMonitNew/index.js
gre: ok so `account ? account.coinType : cointType` maybe?

PR #269, opened 2018-04-10
Receive funds Modal

loeck:
- [x] Add "Wrong address" screen
- [x] Add "check" when step is valid on Breadcrumb


src/components/CurrentAddress/index.js
```diff
@@ -0,0 +1,210 @@
...
+    <Box fontSize={3} ff="Museo Sans|Bold" mt={1}>
+      {label}
+    </Box>
+  </FooterButtonWrapper>
+)
+
+type Props = {
+  account: Account,
+  addressVerified: null | boolean,
+  amount: null | string,
```
gre:
```
addressVerified?: boolean,
amount?: string
```

src/components/DeviceConfirm/index.js
```diff
@@ -0,0 +1,180 @@
...
+    width: 9px;
+    z-index: 1;
+  }
+`
+
+type Props = {
+  notValid: boolean,
+}
+
+const DeviceConfirm = (props: Props) => (
```
gre: minor: It can be good to make this a pure component OR to put the static `<svg>` element into a constant so React don't have to work a lot

src/components/DeviceConfirm/index.js
```diff
@@ -0,0 +1,180 @@
...
+          strokeWidth="2"
+          d="M126.9617746 31.060606c0 .55228475-.4477153 1-1 1H99.7373741c-2.7614237 0-5-2.23857625-5-5v-8.4856683c0-2.7614238 2.2385763-5 5-5h26.2244005c.5522847 0 1 .4477152 1 1V31.060606z"
+        />
+        <path
+          stroke="#142533"
+          strokeWidth="2"
+          d="M94.3535357 25.85229841H83.4242428V19.4170232h10.9292929v6.43527521z"
+        />
+        <path
+          fill="url(#d)"
```
gre: will this have a "global id" problem?

src/components/RequestAmount/index.js
```diff
@@ -81,35 +89,49 @@ export class RequestAmount extends PureComponent<Props> {
     }
   }
 
-  render() {
-    const { t, value, account, rightUnit, getCounterValue } = this.props
+  renderInputs(containerProps: Object) {
```
gre: i would probably recommend `*`  over `Object`, even tho it's probably infered to `any`. best would be to specify the type

src/components/base/Modal/index.js
```diff
@@ -28,11 +28,20 @@ const springConfig = {
 
 const mapStateToProps: Function = (
   state,
-  { name, isOpened }: { name: string, isOpened?: boolean },
-): Object => ({
-  isOpened: isOpened || (name && isModalOpened(state, name)),
-  data: getModalData(state, name),
-})
+  { name, isOpened, onBeforeOpen }: { name: string, isOpened?: boolean, onBeforeOpen: Function },
+): Object => {
```
gre: `*` would likely help more than `Object` because it would likely infer correctly the type

src/components/base/QRCode/index.js
```diff
@@ -28,6 +28,9 @@ class QRCode extends PureComponent<Props> {
     qrcode.toCanvas(this._canvas, data, {
       width: size,
       margin: 0,
+      color: {
+        light: '#ffffff00', // transparent background
```
gre: I've learned something

src/components/base/Select/index.js
```diff
@@ -136,12 +127,31 @@ const IconSelected = styled(Box).attrs({
   opacity: ${p => (p.selected ? 1 : 0)};
 `
 
+const AngleDown = props => (
+  <Box color="grey" alignItems="center" justifyContent="center" {...props}>
+    <svg viewBox="0 0 16 16" width="16" height="16">
```
gre: same, this can be in a constant imo. OR this can be a pure component. (but it's minor ;) )

src/components/modals/Receive/03-step-confirm-address.js
```diff
@@ -0,0 +1,59 @@
...
+const Text = styled(Box).attrs({
+  color: 'smoke',
+})`
+  text-align: center;
+`
+
+type Props = {
+  account: Account | null,
+  addressVerified: null | boolean,
+  device: Device | null,
```
gre: `| null` can be `?:`

src/components/modals/StepConnectDevice.js
```diff
@@ -11,7 +11,7 @@ import DeviceMonit from 'components/DeviceMonitNew'
 
 type Props = {
   accountName?: string,
-  account?: Account,
+  account?: Account | null,
```
gre: the previous was correct to me. if you don't define `account: undefined` as defaultProps.

src/components/DeviceConfirm/index.js
loeck: Good catch

src/components/base/Select/index.js
loeck: Maybe we can do this for all icons

PR #270, opened 2018-04-10
Group operations by day in Dashboard and account page

meriadec:
2018-04-10_1072x761

### DIFF
--- a/src/components/OperationsList/index.js
+++ b/src/components/OperationsList/index.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { Component } from 'react'
+import React, { PureComponent } from 'react'
 import styled from 'styled-components'
 import moment from 'moment'
 import { connect } from 'react-redux'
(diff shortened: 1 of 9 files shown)


src/components/OperationsList/index.js
```diff
@@ -214,77 +221,108 @@ const mapDispatchToProps = {
...
       return true
     }
 
-    if (this._hashCache === null) {
+    if (this.props.canShowMore !== nextProps.canShowMore) {
       return true
     }
 
-    return !isEqual(this._hashCache, this.getHashCache(nextProps.operations))
+    return false
```
gre: what prevents us to use a `PureComponent` ?

src/components/OperationsList/index.js
```diff
@@ -214,77 +221,108 @@ const mapDispatchToProps = {
...
+      withAccount,
+      nbToShow,
+    } = this.props
+
+    if (!account && !accounts) {
+      console.warn('Preventing render OperationsList because not received account or accounts') // eslint-disable-line no-console
+      return null
+    }
+    const groupedOperations = accounts
+      ? groupAccountsOperationsByDay(accounts, nbToShow || 20)
```
gre: 20 is already a defaultProps so you don't need this

src/components/OperationsList/index.js
```diff
@@ -214,77 +221,108 @@ const mapDispatchToProps = {
...
+            </Text>
+          )}
+          {groupedOperations.map(group => {
+            const d = moment(group.day)
+            const isToday = d.isSame(today, 'day')
+            const isYesterday = d.isSame(moment(today).subtract(1, 'day'), 'day')
+            return (
+              <Box flow={2} key={group.day.toISOString()}>
+                <Box ff="Open Sans|SemiBold" fontSize={4} color="grey">
+                  {isToday ? 'Today' : isYesterday ? 'Yesterday' : d.format('D MMMM YYYY')}
```
gre: if you want an already internationalized version, you could do something like this: https://github.com/gre/ledger-wallet-mobile/blob/master/src/screens/Dashboard/SectionHeader.js 
(BTW see how we have chosen to display things like `April 10, 2018 – Today` , maybe we could suggest to design , we have room to display this , even on mobile ^^ )

gre:
🚀

PR #271, opened 2018-04-12
Simple implementation of Macbook TouchBar

loeck:


src/components/CalculateBalance.js
```diff
@@ -33,14 +37,24 @@ function calculateBalanceToState(props: Object) {
 }
 
 class CalculateBalance extends PureComponent<Props, State> {
+  static defaultProps = {
+    onCalculate: noop,
+  }
+
   state = calculateBalanceToState(this.props)
 
+  componentDidMount() {
```
meriadec: I think it's possible to do that in willMount or constructor

gre: willMount is deprecated :trollface: and I think didMount is now the recommended way for **all the things**

gre: IMO a better pattern would be to have a component `<TouchBarUI />` and you pass it whatever you want like

 `<TouchBarUI balance={foo} />`

instead of an imperative callback called in didMount

gre: if you want an analogy: https://facebook.github.io/react-native/docs/statusbar.html

loeck: Yep, clearly, but maybe we need to wait if you want to keep that or improved that before implement nice component.

meriadec: `didMount` is more suitable if we want to access dom, here we just want to call prop function with computed value

PR #283, opened 2018-04-16
Better flow for modal Send

loeck:


src/components/modals/PrevButton.js
```diff
@@ -0,0 +1,37 @@
...
+import type { T } from 'types/common'
+
+import Button from 'components/base/Button'
+import Box from 'components/base/Box'
+
+import IconAngleLeft from 'icons/AngleLeft'
+
+const Wrapper = styled(Button).attrs({
+  fontSize: 4,
+  ml: 4,
```
meriadec: not a big fan of components leaking margins. i think it's better to add the `ml 4` when we use PrevButton rather than always having it by default

loeck: Yep, but just need feedback for design team with right position of this Button (maybe it's changed)

PR #288, opened 2018-04-18
Ability to change settings by currency

meriadec:
sett


src/reducers/settings.js
```diff
@@ -18,6 +18,7 @@ const defaultState: SettingsState = {
     isEnabled: false,
     value: '',
   },
+  currencies: [],
```
gre: you might want to store them by coinId in future for performance.

meriadec: That's right

PR #303, opened 2018-04-25
wip live-112 settings page updates

NastiaS:


src/components/SettingsPage/DisablePasswordModal.js
```diff
@@ -0,0 +1,121 @@
+// @flow
+
+import React, { PureComponent } from 'react'
+import bcrypt from 'bcryptjs'
+import { ErrorMessageInput } from 'components/base/Input'
```
meriadec: just a matter of code style, but let's separate import of vendor package from local files

gre: on mobile eslint would complain, not here? 🤔

NastiaS: No, it didn't complain, but yes, it wouldn't let it go on the mobile. I guess that's why I missed it, relying on tools too much 😃

meriadec: yeah.. it didn't complain because eslint believe that it's a node_module... because of babel-plugin-module-resolver :trollface:

PR #305, opened 2018-04-27
select perfomance improvements, small design fixes, wip

NastiaS:


meriadec:
You may want to add an overflow: hidden on Select items, to fix the visual glitch when typing.

PR #310, opened 2018-04-30
Fix to prevent app crash on the "abandon" seed

MortalKastor:
The case where there's no "from" address on an operation shouldn't happen... but on testnet (sh)it happens :+1:

### DIFF
--- a/src/components/OperationsList/Operation.js
+++ b/src/components/OperationsList/Operation.js
@@ -49,6 +49,10 @@ const OperationRaw = styled(Box).attrs({
 `
 
 const Address = ({ value }: { value: string }) => {
+    if (!value) {
+      return <Box />
+    }
+
   const addrSize = value.length / 2
 
   const left = value.slice(0, 10)


src/components/OperationsList/Operation.js
```diff
@@ -49,6 +49,10 @@ const OperationRaw = styled(Box).attrs({
 `
 
 const Address = ({ value }: { value: string }) => {
+    if (!value) {
```
gre: there is an issue in term of Flowtype at some point. according to flow `!value` should never have happened. which means props probably should be `{value: ?string}` and then we need also to have that reflected in the original Operation type (not exactly sure where, address autorized to be null ?)

anyway, that's for later, we'll just merge this first ;)

src/components/OperationsList/Operation.js
```diff
@@ -49,6 +49,10 @@ const OperationRaw = styled(Box).attrs({
...
+      return <Box />
```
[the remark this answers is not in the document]
MortalKastor: I usually rather neuter a component into a stub, than removing it completely from DOM to prevent side effects. But, yeah, in this context `null` is perfectly safe and would have made more sense.

src/components/OperationsList/Operation.js
gre: https://github.com/LedgerHQ/ledger-live-common/blob/master/src/types/operation.js#L18

PR #312, opened 2018-05-02
locale for number formatting and language derived from the system

NastiaS:


src/components/base/FormattedVal/index.js
```diff
@@ -98,11 +98,13 @@ export function FormattedVal(props: Props) {
     if (withIcon && isNegative) {
       val *= -1
     }
+    const locale = settings ? settings.language : ''
```
meriadec: is there a risk of `settings` not being populated here ? I see it comes directly from `state.settings` that has at least a default state

src/components/AccountPage/index.js
```diff
@@ -97,12 +99,14 @@ class AccountPage extends PureComponent<Props, State> {
...
           currency: formatCurrencyUnit(account.unit, account.balance, {
             showCode: true,
+            locale: settings.language,
           }),
           counterValue: formatCurrencyUnit(
             getFiatCurrencyByTicker(counterValue).units[0],
             data.totalBalance,
             {
               showCode: true,
+              locale: settings.language,
```
meriadec: not sure of the rightness of mixing up `locale` and `language` together, I feel that it brings more confusion on an already confusing theme

src/components/base/FormattedVal/index.js
gre: no, settings shouldn't be `?: settings` I guess it's part of the flow refactoring we need to do

NastiaS: there is no risk, but in this file and some other files we shouldn't do something like 
settings?: Settings, in Props

src/components/AccountPage/index.js
gre: no it should remain "locale" as in the first parameter of `Number.prototype.toLocaleString`. it is supposed to be a locale string like `fr-FR` etc, not only just a "language" (fr / en / ...).

Which means, in the future, it should probably be `${settings.language}-${settings.region}` , unless we have a `settings.locale` that would be this concatenation

meriadec: yeah, my point is not to rename `locale` to `language`, but only the fact that we put *a variable named `language` inside another, named `locale`. If I'm right it's two different things, ye? Like: `fr` and `fr-FR`

gre: yeah I think locale includes both "fr" and "fr-FR" , like the region can be omitted per spec. so locale & language are different concept I feel

PR #313, opened 2018-05-02
remove all user refs and update design of the top bar

NastiaS:


src/components/DashboardPage/index.js
```diff
@@ -131,7 +129,7 @@ class DashboardPage extends PureComponent<Props, State> {
         <Box horizontal alignItems="flex-end">
           <Box grow>
             <Text color="dark" ff="Museo Sans" fontSize={7}>
-              {t('dashboard:greetings', { name: username })}
+              {t('dashboard:greetings')}
```
gre: does the greedings make sense at all now 🤔  ?

src/icons/Lock.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import React from 'react'
+
+export default ({ size, ...p }: { size: number }) => (
+  <svg width={size} height={size} {...p}>
+    <path
```
gre: general practice would be either to do a PureComponent or to at least move `<path>` into an outer const, so we don't render again & again. (this might help a bit react perf if we do that for all icons.)

src/components/DashboardPage/index.js
NastiaS: I asked Brian what should be there and he said 'Good morning'

meriadec: What if it's the night ? Do we plan to update the greeting according to time of the day ? What about displaying meteo infos ? :trollface:

NastiaS: We could say "Greetings!" to still have a greeting or if we have to have morning/evening one we should probably have a story to make a greeting dynamic

PR #315, opened 2018-05-06
finish off locale settings

gre:
target MVP. this should be good for now, we don't spend more time on locale now 🙏 plz


src/renderer/init.js
```diff
@@ -23,6 +23,11 @@ import App from 'components/App'
 
 import 'styles/global'
 
+console.log(process.env)
```
meriadec: eh

PR #318, opened 2018-05-07
Integrate countervalues changes

gre:
related https://github.com/LedgerHQ/ledger-live-common/pull/16


src/components/CounterValue/index.js
```diff
@@ -14,8 +14,9 @@ import type { State } from 'reducers'
 type OwnProps = {
   // wich market to query
   // FIXME drop ticker in favor of currency
-  ticker: string,
+  ticker?: string,
   currency?: Currency,
```
gre: drop ticker & make currency mandatory

src/components/CounterValue/index.js
```diff
@@ -45,9 +49,10 @@ const mapStateToProps = (state: State, props: OwnProps) => {
   const counterValueCurrency = counterValueCurrencySelector(state)
   const counterValue =
     !counterValueCurrency || !currency
```
gre: none of these case should happen now, (currency mandatory / countervalueCurrency always available)

src/components/DashboardPage/AccountCard.js
```diff
@@ -57,35 +57,36 @@ const AccountCard = ({
...
-              <FormattedVal
-                isPercent
-                val={Math.floor((totalBalance - refBalance) / refBalance * 100)}
-                alwaysShowSign
-                fontSize={3}
-              />
+              {balanceStart && isAvailable ? (
+                <FormattedVal
+                  isPercent
+                  val={Math.floor((balanceEnd - balanceStart) / balanceStart * 100)}
```
gre: we need a `DeltaChange` component to factorize the logic (because we actually want to avoid `/ 0` & also if % is too big diverge the rendering)

PR #320, opened 2018-05-13
Integrate exchange page

meriadec:
2018-05-13_1944x1160

There is a problem with the coinhouse logo exported from zeplin, guess it's some svg stuff:

2018-05-13_2634x695


src/components/ExchangePage/logos/coinhouse.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import React from 'react'
+
+export default ({ width }: { width: number }) => (
```
gre: make it pure or external const (can wrap a `<g>`) ;)

src/icons/Exchange.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import React from 'react'
+
+export default ({ size, ...p }: { size: number }) => (
+  <svg viewBox="0 0 16 16" height={size} width={size} {...p}>
+    <path
```
gre: same, can be a const

meriadec: Maybe we should consider doing a pass on all icons to do that.

src/components/ExchangePage/logos/coinhouse.js
meriadec: or use fragment eheh

PR #340, opened 2018-05-18
Auto update for Mac

MortalKastor:


package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "Ledger Live",
   "description": "Ledger Live - Desktop",
   "repository": "https://github.com/LedgerHQ/ledger-live-desktop",
-  "version": "0.1.0",
+  "version": "0.1.0-alpha",
```
meriadec: I think we should use `1.0.0-alpha.x`

src/main/autoUpdate.js
```diff
@@ -16,5 +17,14 @@ export default (notify: SendFunction) => {
...
 
 export function quitAndInstall() {
-  autoUpdater.quitAndInstall()
+  setImmediate(() => {
+    const browserWindows = BrowserWindow.getAllWindows()
+
+    app.removeAllListeners('window-all-closed')
+    browserWindows.forEach(browserWindow => {
+      browserWindow.removeAllListeners('close')
+    })
```
meriadec: curious about what was the problem here?

gre: (add a comment in sourcecode)

MortalKastor:
Successfully built and drafted a release (in this repo) which should self-update on each new `prerelease`.

meriadec:
Naiiiiiiceu

PR #346, opened 2018-05-21
Implement Ethereum

gre:
this implement ethereum. In theory ethereum_classic will also work but i've not tested.

**it should be feature complete for the MVP: scan accounts from device, synchronize operations from blockchain, sign and broadcast a transaction.**

this is good enough to go in for now & we'll polish over time.
There are things to improve a bit everywhere. For instance we might want to introduce more cache on top of the api (typically the `/current` needs to be throttled ). Also, we might want to continue scanning accounts "forever" and just let the user select accounts and continue on the UI (because we can't know if it's "over", especially for Ethereum case)


gre:
@meriadec there is a new `signTransaction` that might be interesting for you to add btc . the folder `signTransactionForCurrency` is meant to be extended to many coins

gre:
@meriadec tomorrow morning , would you mind porting to latest `live-common` on your side? it breaks in btc impl because Operation needs to provide blockHash

meriadec:
TODO: refacto to have a `src/commands` directory, and cleanup the mess inside `src/main/bridge` and `src/internals`

PR #347, opened 2018-05-22
Btc like transaction & fix confirm address

meriadec:


src/components/DeviceCheckAddress.js
```diff
@@ -27,35 +27,34 @@ class CheckAddress extends PureComponent<Props, State> {
     this.verifyAddress({ device, account })
   }
 
-  componentDidUnmount() {
-    if (this.sub) this.sub.unsubscribe()
```
gre: I think we might want either to keep that or to just have a flag to know if it's unmounted and not do a setState if it's gone

meriadec: plus one

meriadec: done

PR #348, opened 2018-05-22
helpers/derivations to derivate different paths

gre:


src/bridge/EthereumJSBridge.js
```diff
@@ -30,8 +30,8 @@ const toAccountOperation = (account: Account) => (tx: Tx): Operation => {
     hash: tx.hash,
     address: sending ? tx.to : tx.from,
     amount: (sending ? -1 : 1) * tx.value,
-    blockHeight: tx.block.height,
-    blockHash: tx.block.hash,
+    blockHeight: tx.block && tx.block.height,
+    blockHash: tx.block && tx.block.hash,
```
gre: note: we need to make all `blockHeight` and `blockHash` a `?` because it can be null

src/bridge/EthereumJSBridge.js
```diff
@@ -149,24 +151,25 @@ const EthereumBridge: WalletBridge<Transaction> = {
...
-          if (r.complete) {
-            complete()
-            break
+        const derivations = getDerivations(currency)
+        const last = derivations[derivations.length - 1]
+        for (const derivation of derivations) {
+          const isStandard = last === derivation
+          for (let index = 0; index < 255; index++) {
+            const path = derivation({ currency, x: index, segwit: false })
+            console.log(path)
```
meriadec: wesh

PR #367, opened 2018-05-25
improve a bit with a tmp design for step import

gre:


src/components/modals/AddAccount/03-step-import.js
```diff
@@ -17,16 +19,28 @@ function StepImport(props: Props) {
...
             onClick={onToggleAccount && !isExisting ? () => onToggleAccount(account) : undefined}
           >
-            {isSelected && `[SELECTED]`} {isExisting && `[ALREADY IMPORTED]`} {account.name}
+            <CheckBox isChecked={!!isSelected} onChange={onToggleAccount} />
+            <Box grow fontSize={6} style={{ paddingLeft: 10 }}>
+              {account.name}
+            </Box>
+            <FormattedVal
+              alwaysShowSign={false}
+              color="warmGrey"
```
NastiaS: not sure we have warmGrey 😁

PR #375, opened 2018-05-25
wording-lockscreen

amougel:


static/i18n/en/common.yml
```diff
@@ -23,6 +23,6 @@ eastern: Eastern
 western: Western
 lockScreen:
   title: Welcome Back
-  subTitle: Your application is locked
```
meriadec: used [here](https://github.com/LedgerHQ/ledger-live-desktop/blob/490d8bf55ef0adc981cb1ff6387903e09aca3b4c/src/components/IsUnlocked.js#L141)

static/i18n/en/common.yml
```diff
@@ -23,6 +23,6 @@ eastern: Eastern
...
-  description: Please enter your password to continue
-  inputPlaceholder: Type your password
+  description_osx: Enter your password or use Touch ID to unlock your Ledger Live session
```
meriadec: unused

gre: also minor note, but if you want a custom message for touch id, it should not be `_osx` (because it's not related to OS, some mac have it, some won't & we should figure out a way to detect if supported)

PR #377, opened 2018-05-26
add a retry mecanism on the open()

gre:
this will fix most of the time you have the error `Failed to open device` but **there is a more fundamental issue** related to this..

---

Initially here is what i wanted to do: debouncing the .close() and reuse Transport instances across call. the issue is that we currently always `kill:true` the devices process.. we probably should not kill it and always have one and only one process that is always up & deal with the devices.

```
import debounce from 'lodash/debounce'

// TODO this should work but unfortunately our process get always re-started.. a device process should always be up IMO...
const transports = {}
const transportsDebouncedClose = {}

console.log('deviceAccess semaphore STARTED')

function takeDevice(path) {
  const t = transports[path]
  const debouncedClose = transportsDebouncedClose[path]
  if (debouncedClose) debouncedClose.cancel() // we use the resource so we need to cancel scheduled close()
  if (t) return t // if there is a transport instance, we just reuse it
  transports[path] = retry(() => TransportNodeHid.open(path)) // otherwise we create it with a retry mecanism because it sometimes fails
  console.log(`Transport#open ${path}`)
  return transports[path]
}

function releaseDevice(path, force) {
  if (!transportsDebouncedClose[path]) {
    // if there is no debounced function for the path, we will create one and always reuse it
    transportsDebouncedClose[path] = debounce(() => {
      const t = transports[path]
      if (t) {
        delete transports[path]
        t.close()
        console.log(`Transport#close ${path}`)
      }
    }, 1000)
  }
  const debouncedClose = transportsDebouncedClose[path]
  debouncedClose()
  if (force) debouncedClose.flush()
}
```

 and with this code in withDevice:

```
    takeSemaphorePromise(sem, async () => {
      const t = await takeDevice(devicePath)
      let error
      try {
        const res = await job(t)
        // $FlowFixMe
        return res
      } catch (e) {
        error = e
        throw e
      } finally {
        await releaseDevice(devicePath, !!error)
      }
    })
```

### DIFF
--- a/src/helpers/deviceAccess.js
+++ b/src/helpers/deviceAccess.js
@@ -1,7 +1,8 @@
 // @flow
 import createSemaphore from 'semaphore'
 import type Transport from '@ledgerhq/hw-transport'
-import CommNodeHid from '@ledgerhq/hw-transport-node-hid'
+import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'
+import { retry } from './promise'
 
 // all open to device must use openDevice so we can prevent race conditions
 // and guarantee we do one device access at a time. It also will handle the .close()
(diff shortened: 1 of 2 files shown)


gre:
i think we'll go mvp with that but there are fundamental issue & we'll re-architecture process from scratch. too much forked & killed process (literally for each command!).

gre:
let's try to migrate all to using "commands" so our code is independent from the underlying event logic, & later we'll refactor for all

PR #378, opened 2018-05-26
Redesign import accounts

meriadec:
improt


src/components/SideBar/index.js
```diff
@@ -138,9 +119,37 @@ class SideBar extends PureComponent<Props> {
...
+      )
+    })}
+  </Fragment>
+))
+
 export default compose(
-  connect(mapStateToProps, mapDispatchToProps, null, {
-    pure: false,
-  }),
+  connect(mapStateToProps, mapDispatchToProps, null, { pure: false }),
```
gre: why do we need to `{pure:false}` ?

src/components/base/Input/index.js
```diff
@@ -68,6 +68,8 @@ type Props = {
   keepEvent?: boolean,
   onBlur: Function,
   onChange?: Function,
+  onEnter?: Function,
+  onEsc?: Function,
```
gre: don't forget `()=>void` is better than Function 🙏  or if it actually take something, put the exact type, if not, we are prone to more blindness on flow types

src/components/base/Input/index.js
```diff
@@ -119,6 +136,11 @@ class Input extends PureComponent<Props, State> {
     onBlur(e)
   }
 
+  handleSelectEverything = () => {
+    this._input && this._input.setSelectionRange(0, this._input.value.length)
```
gre:
```
const {_input}=this;
if (_input) {
  _input.select();
  _input.focus();
}
```

should do the same thing, just a bit more concise

src/components/modals/ImportAccounts/index.js
```diff
@@ -0,0 +1,224 @@
...
+        )}
+      />
+    )
+  }
+}
+
+export default compose(connect(mapStateToProps, mapDispatchToProps), translate())(ImportAccounts)
+
+function idleCallback() {
+  return new Promise(resolve => window.requestIdleCallback(resolve))
```
gre: I'm pretty sure it's an antipattern and is equivalent to `Promise.resolve()` . i used to do same with requestAnimationFrame but fact is, a promise .then runs in the next tick, so either directly use requestIdleCallback but probably don't transform it to a promise

src/components/modals/ImportAccounts/steps/03-step-import.js
```diff
@@ -0,0 +1,168 @@
...
+  startScanAccountsDevice() {
+    const { currency, currentDevice, setState } = this.props
+    try {
+      if (!currency) {
+        throw new Error('No currency to scan')
+      }
+
+      if (!currentDevice) {
+        throw new Error('No device')
+      }
```
gre: these 2 cases shouldn't happen, meaning we should probably have the props type to be directly currency: Currency, etc.. because you don't want to render until these satisfies. 

had same issue on the Send flow, current solution is not so satisfying but still better in types I feel https://github.com/LedgerHQ/ledger-live-desktop/blob/master/src/components/modals/Send/03-step-verification.js#L45

anyway, it's for later

src/components/SideBar/index.js
meriadec: Because of the internal and mysterious behaviour of react-router combined with redux (& react-router-redux..)

see [here](https://github.com/reduxjs/react-redux/blob/master/docs/troubleshooting.md#my-views-arent-updating-when-something-changes-outside-of-redux) for more context

src/components/modals/ImportAccounts/index.js
gre: ultimately, should we instead have a addAccounts action?

PR #379, opened 2018-05-27
ETH self transaction: Receive to appear after Send

gre:


gre:
yeah! I mean, i don't see another way this can be solved easily, maybe sort the Tx first & then don't need to sort the ops? anyway, that's a quick way for now

PR #392, opened 2018-05-29
rework Currencies settings to fix confirmations & rm unused settings

gre:


src/components/SettingsPage/sections/Currencies.js
```diff
@@ -29,30 +31,11 @@ import {
...
-  exchange: '',
-}
-
 type Props = {
   counterValueCurrency: Currency,
   currencies: CryptoCurrency[],
-  settings: Settings,
-  saveSettings: Function,
+  settings: SettingsState,
+  saveSettings: ($Shape<SettingsState>) => void,
```
NastiaS: we have to update all similar `Function` types everywhere in the app, we have it all over the place

PR #394, opened 2018-05-29
Auto update polish

MortalKastor:


package.json
```diff
@@ -92,6 +92,7 @@
     "rxjs": "^6.2.0",
     "rxjs-compat": "^6.1.0",
     "semaphore": "^1.1.0",
+    "secp256k1": "3.4.0",
```
gre: why do we need to add this dependency?

MortalKastor: It's not a new dependency actually (`yarn why secp256k1` on master says `"ethereumjs-tx#ethereumjs-util" depends on it`).

I needed to force the last `secp256k1` version with prebuilds for cross-building (and I got it wrong, hence the new commit)

PR #400, opened 2018-05-30
Rework the process architecture & ipc

gre:
- all commands are in commands folder & you need to list them in commands/index.js
- `createCommand(id, f)` don't take a channel anymore, just an id.
- device code run about 2-3 faster
- deleted some dead or unused code
- logic for updater is preserved, it could be a command technically that run on Main process, but good for now & i didn't want to change it too much without proper testing. that said, later we should be able to run command on a desired process (Main / Internal). code is almost ready for this.

```


                    ipcRendererSendCommand             command-unsubscribe
                                                       command
    +-----------------+            +----------------+            +----------------+
    |                 |    A       |                | +--------> |                |
    |    Renderer     +----------> |                |     A      |    Internal    |
    |    process      |            |    Main        | +--------> |    process     |
    |                 |            |    process     |            |                |
    |  cmd.send(A)    |            |                |            |  cmd.exec(A)   |
    |                 | <----------+                | <--------+ |                |
    +-----------------+            +----------------+            +----------------+

                    ipcMainListenReceiveCommands        CommandEvent
                                                        | NEXT, data
                                                        | ERROR, error
                                                        | COMPLETE


```


src/main/bridge.js
```diff
@@ -1,100 +1,89 @@
...
 
 import setupAutoUpdater, { quitAndInstall } from './autoUpdate'
 
-const { DEV_TOOLS } = process.env
-
 // sqlite files will be located in the app local data folder
 const LEDGER_LIVE_SQLITE_PATH = path.resolve(app.getPath('userData'), 'sqlite')
 
-const processes = []
+let internalProcess
```
meriadec: so that means that we no longer are able to have multiple internal processes. how about keeping the ability to do it?

src/renderer/runJob.js
```diff
@@ -1,34 +0,0 @@
...
-      }
-      ipcRenderer.removeListener('msg', handler)
-      if (type === successResponse) {
-        resolve(data)
-      } else if (type === errorResponse) {
-        reject(data)
-      }
-    }
-  })
-}
```
meriadec: bye :hand:

src/main/autoUpdate.js
```diff
@@ -6,12 +6,12 @@ import { autoUpdater } from 'electron-updater'
 type SendFunction = (type: string, data: *) => void
 
 export default (notify: SendFunction) => {
-  autoUpdater.on('checking-for-update', () => notify('updater.checking'))
-  autoUpdater.on('update-available', info => notify('updater.updateAvailable', info))
-  autoUpdater.on('update-not-available', () => notify('updater.updateNotAvailable'))
-  autoUpdater.on('error', err => notify('updater.error', err))
-  autoUpdater.on('download-progress', progress => notify('updater.downloadProgress', progress))
-  autoUpdater.on('update-downloaded', () => notify('updater.downloaded'))
+  autoUpdater.on('checking-for-update', () => notify('checking'))
```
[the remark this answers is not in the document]
meriadec: oh, are events all gonna go to `updater` channel?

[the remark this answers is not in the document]
gre: yeah they have their own updater channel now

src/main/bridge.js
gre: there is one internal process to run everything. since everything is asynchronous it shouldn't be a hurdle. if there are CPU intensive work, then yeah we need to introduce the concept of having a target process.
like you want to run that command on "internal", or on "main" or on "whatever".

but the previous way was too crazy because it was forking a process for each single event message

gre: what's important tho is that the commands that need the device have to run inside the same process.
and i think all our commands today uses the device so we have to do it like this

PR #417, opened 2018-05-31
Rewrite Select component

meriadec:
For moar perfs.

The old select has been renamed to `LegacySelect` and is still used at some places (actually: 5 places) where it works well (e.g: `InputCurrency`).

select


gre:
good job!!

PR #421, opened 2018-05-31
delete account in account settings

NastiaS:


meriadec:
We should TODO to remove it from sqlite as well if needed

meriadec:
Yep.

PR #426, opened 2018-06-01
Improve errors in send flow

gre:


src/components/RequestAmount/index.js
```diff
@@ -97,12 +100,13 @@ export class RequestAmount extends PureComponent<Props> {
...
 
   renderInputs(containerProps: Object) {
-    const { value, account, rightCurrency, getCounterValue, exchange } = this.props
+    const { value, account, rightCurrency, getCounterValue, exchange, canBeSpent } = this.props
     const right = getCounterValue(account.currency, rightCurrency, exchange)(value) || 0
     const rightUnit = rightCurrency.units[0]
     return (
       <Box horizontal grow shrink>
         <InputCurrency
+          error={canBeSpent ? null : 'Not enough balance'}
```
NastiaS: we should probably create a yml file for all errors so it can be translated too

PR #427, opened 2018-06-01
fix y axis values on graphs and dashboard page

valpinkman:


src/components/BalanceSummary/index.js
```diff
@@ -59,6 +60,9 @@ const BalanceSummary = ({
                   height={250}
                   currency={counterValue}
                   tickXScale={selectedTime}
+                  renderTickY={(val, account) =>
+                    account ? formatCurrencyUnit(account.unit, val) : val
```
gre: it should not be `val` for the else case, it should be `formatCurrencyUnit(counterValue.units[0], val)`

valpinkman: ok I'll fix that

PR #429, opened 2018-06-01
Libcore synchronisation

meriadec:


src/helpers/withLibcore.js
```diff
@@ -0,0 +1,31 @@
...
+type Job = (Object, Object) => any
+
+export default function withLibcore(job: Job) {
+  if (!walletPool) {
+    walletPool = core.instanciateWalletPool({
+      // sqlite files will be located in the app local data folder
+      dbPath: process.env.LEDGER_LIVE_SQLITE_PATH,
+    })
+  }
+  // $FlowFixMe WTF is happening here, dudes.
```
gre: 🤣🤣🤣
BTW it would be great if you could define a parametric type so that we have this:

```js
type Job<A> = (*, *) => Promise<A>
export default function<A> withLibcore(job: Job<A>): Promise<A> {
```

that might help type & also it's probably better for type..

src/helpers/withLibcore.js
```diff
@@ -0,0 +1,31 @@
...
+  queue = queue.then(async () => {
+    try {
+      if (!walletPool) {
+        throw new Error('wallet pool not instanciated. this should not happen')
+      }
+      return job(core, walletPool)
+    } catch (e) {
+      console.log(`withLibCore: Error in job`, e) // eslint-disable-line no-console
+      return Promise.resolve()
```
gre: mmh to satisfy the type shared above you might want to change a bit. maybe this:

```js
const p = queue.then(() => {
  invariant(walletPool, 'wallet pool not instanciated. this should not happen')
  return job(core, walletPool)
})

queue = p.catch(() => {});

return p
```

src/helpers/libcore.js
```diff
@@ -1,21 +1,16 @@
...
 import { withDevice } from 'helpers/deviceAccess'
 import { getCryptoCurrencyById } from '@ledgerhq/live-common/lib/helpers/currencies'
 
 import type { AccountRaw, OperationRaw, OperationType } from '@ledgerhq/live-common/lib/types'
 import type { NJSAccount, NJSOperation } from '@ledgerhq/ledger-core/src/ledgercore_doc'
 
+import * as accountId from 'helpers/accountId'
+
 type Props = {
+  core: Object,
```
gre: prefer a `core: *` for now until we have an actual type 🤔

src/helpers/accountId.js
```diff
@@ -0,0 +1,16 @@
...
+  xpub: string,
+  walletName: string,
+}
+
+export function encode({ type, xpub, walletName }: Params) {
+  return `${type}:${xpub}:${walletName}`
+}
+
+export function decode(accountId: string) {
+  const [type, xpub, walletName] = accountId.split(':')
```
gre: please make this function returning a `Params`. & also typecheck things (so it's actually all typeof string) and throw if it's not valid. because in future if this don't decode, like we have old id, etc.. maybe it's better to have more clear validation.
maybe you could even have a `libcore1` suffix just in case we want to do some logic in future or change the format? 🤔

src/reducers/settings.js
```diff
@@ -63,7 +63,7 @@ const INITIAL_STATE: SettingsState = {
   marketIndicator: 'western',
   currenciesSettings: {},
   region,
-  developerMode: false,
+  developerMode: !!process.env.DEV,
```
gre: why not `__DEV__`

src/reducers/settings.js
```diff
@@ -106,13 +106,15 @@ const handlers: Object = {
...
     ...settings,
+    developerMode: settings.developerMode || !!process.env.DEV,
   }),
   FETCH_SETTINGS: (
     state: SettingsState,
     { payload: settings }: { payload: $Shape<SettingsState> },
   ) => ({
     ...state,
     ...settings,
+    developerMode: settings.developerMode || !!process.env.DEV,
```
gre: no need imo. just initial case is enough

src/reducers/settings.js
meriadec: Ye why not

src/helpers/withLibcore.js
meriadec: honestly I tried to do that. and failed to make it works properly. maybe we'll see that together later.

src/helpers/withLibcore.js
meriadec: in fact the check is totally useless because [it is impossible that we get to this line without wallet pool instanciated](https://github.com/meriadec/ledger-live-desktop/blob/49959943256207894550eae1a1568fbfbae6b1ef/src/helpers/withLibcore.js#L12-L17), but he is here only because flow can't see it. /shrug

gre: a check is never useless ;P

I think what's happening here is as you have defined walletPool as null, it will be infered as `any` because `instanciateWalletPool` is not typed and flow can't tell if it does not return a null too.

gre: you can probably have the invariant before the `const p =` . anyway, flow is weird at first, but everything is perfectly logic as soon as you inspect what it's doing. like what type each variable was inferred too. that's why it's quite interesting to have an IDE like Nuclide to see the inferred type & also the type coverage

src/helpers/withLibcore.js
gre: sent you a PR

src/helpers/accountId.js
```diff
@@ -0,0 +1,22 @@
...
+
+type Params = {
+  type: string,
+  version: string,
+  xpub: string,
+  walletName: string,
+}
+
+export function encode({ type, version, xpub, walletName }: Params) {
+  return `${type}:${version}:${xpub}:${walletName}`
```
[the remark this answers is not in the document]
meriadec: Agree :+1: 

Additionally, it would be useful to have an `export to file <-> import from file` tool (usable from the CrashScreen) to ease support. Can be interesting to allow detailed exports:

```
Export data:
  [x] accounts
    [x] anonymize (randomize names, etc.)
    [ ] include operations
  [ ] settings
```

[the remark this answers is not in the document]
gre: (interesting idea! yeah definitely. & for logs we need a proper logger system on all user actions / redux actions so we can debug appropriately ;) I'll spec out something this week around this subject. it's quite easy to add, i was thinking dom elements could have a "role" attr that expose them to this system & we track&log all onclick. whatever system we can figure out that is enough debuggable but also easy to add (we should not have to console.log at each single onClick is what i mean) )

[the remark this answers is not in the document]
meriadec: going forward on that subject, being able to retrieve apdu calls (and other precise stuff) occuring on a crash could be very useful.

PR #431, opened 2018-06-01
remove sort ops because it should be done by sync only if needed

gre:


src/reducers/accounts.js
```diff
@@ -147,6 +115,7 @@ export function deserializeAccounts(accounts: Account[]) {
     // as account can be passed by main process, the Date types
     // can be converted to string. we ensure here that we have real
     // date
+    // FIXME Account shouldn't be passed by main process. only AccountRaw should be!
```
meriadec: it shouldn't be the case anymore, so I guess we can get rid of that hack

gre: ok, i'll check later ;)

PR #432, opened 2018-06-01
important bugfix on ETH + integrate more currencies unit

gre:


src/components/AdvancedOptions/EthereumKind.js
```diff
@@ -0,0 +1,36 @@
...
+      <Box grow>
+        <Input
+          value={gasLimit}
+          onChange={str => {
+            const gasLimit = parseInt(str, 10)
+            if (!isNaN(gasLimit) && isFinite(gasLimit)) onChangeGasLimit(gasLimit)
+            else onChangeGasLimit(0x5208)
+          }}
+        />
+      </Box>
```
meriadec: Just curiosity but is it necessary to have `Box width 200` > `Label` > `span` > `text` and `Box grow` > `Input`, rather than having `Label width 200` > `text` and `Input grow` ?

src/components/FeesField/EthereumKind.js
```diff
@@ -34,7 +34,7 @@ class FeesField extends Component<Props & { fees?: Fees, error?: Error }, *> {
     return (
       <GenericContainer error={error} help="Gas">
         <InputCurrency
-          defaultUnit={units[1]}
+          defaultUnit={units.length > 1 ? units[1] : units[0]}
```
meriadec: Should we consider again to have `defaultUnit` on currency ?

gre: here it's not really the defaultUnit I want to pick, `unit[0]` is still the default prop.
this one is more the default gas unit to use for Ethereum. not sure it worth introducing this concept.. I think in future, we might actually want to pick one dynamically, like probably the smallest unit where the default value is > 1 ?

for instance, for bitcoin fee per byte, we'll use satoshis, but imagine if btc price is very very high, maybe you know what to use mBTC, etc.. anyway. for now it's ok to hardcode it (for btc, we always take the `last()`.  ETH have way too much magnitude to use `last()` :D

src/components/AdvancedOptions/EthereumKind.js
gre: 🤷‍♂️ no idea. i have not touched much the structure of this

gre: maybe we should factorize the concept of a Field in the future, because there is dups

PR #448, opened 2018-06-05
improve perfs, use flex-wrap on dashboard

gre:


src/components/base/Box/index.js
```diff
@@ -66,8 +68,10 @@ const Box = styled.div`
     `};
 
   > * + * {
-    margin-top: ${p => (!p.horizontal && p.flow ? `${p.theme.space[p.flow]}px` : '')};
-    margin-left: ${p => (p.horizontal && p.flow ? `${p.theme.space[p.flow]}px` : '')};
+    margin-top: ${p =>
+      !p.horizontal && p.flow && !p.flexWrap ? `${p.theme.space[p.flow]}px` : ''};
+    margin-left: ${p =>
+      p.horizontal && p.flow && !p.flexWrap ? `${p.theme.space[p.flow]}px` : ''};
```
meriadec: we may instead not use `flow` and `flexWrap` combined, it makes no sense to have flow + wrap if wrap removes the action of flow

PR #457, opened 2018-06-06
BUG double request to confirm the address on the device in prod

NastiaS:


src/components/DeviceCheckAddress.js
```diff
@@ -22,11 +15,6 @@ class CheckAddress extends PureComponent<Props, State> {
     isVerified: null,
   }
 
-  componentDidMount() {
```
gre: mmh doesn't this change remove the purpose of this component tho?
I would suggest we just remove the component & replace all the usages by directly doing what the `render()` do here?

NastiaS: It is used in 1 place and I think you are right, it doesn't do much at the moment.

meriadec: nice catch lol

PR #460, opened 2018-06-06
Polish code of reducers/accounts

gre:


src/reducers/accounts.js
```diff
@@ -73,41 +71,16 @@ export const accountSelector = createSelector(
   (accounts, accountId) => accounts.find(a => a.id === accountId),
 )
 
-export function decodeAccount(account: AccountRaw): Account {
-  return accountModel.decode({
+export const decodeAccount = (account: AccountRaw): Account =>
```
meriadec: no real need to change here, but it's pretty handy to have named function, so the (eventual :trollface: ) stacktraces contains more information

gre: I always thought babel was making it named ? did this changed? at least it works for React components

PR #462, opened 2018-06-06
bump some deps

gre:


src/bridge/types.js
```diff
@@ -79,10 +79,10 @@ export interface WalletBridge<Transaction> {
   isValidTransaction(account: Account, transaction: Transaction): boolean;
 
   // render the whole Fees section of the form
-  EditFees?: React$ComponentType<EditProps<Transaction>>;
+  EditFees?: *; // React$ComponentType<EditProps<Transaction>>;
```
meriadec: pretty sure that flow will remove that semicolon

src/reducers/accounts.js
```diff
@@ -96,7 +96,7 @@ export function serializeAccounts(accounts: any): Account[] {
   return accounts ? accounts.map(accountModel.decode) : []
 }
 
-export function deserializeAccounts(accounts: Account[]) {
+export function deserializeAccounts(accounts: Account[]): Array<any> {
```
meriadec: I think now we don't send<->receive Accounts types anymore (we always encode them to AccountRaw) so we should be able to keep the type and remove the horrible thing here:

⏬

src/bridge/types.js
gre: not in a type def I think

src/reducers/accounts.js
gre: yeah that'(s fixed in another PR :D

PR #471, opened 2018-06-07
Fix inputs

gre:


meriadec:
I loled.

so `HUGE TODO: investigate deeply`

PR #483, opened 2018-06-08
Improve how Sync happen

gre:
- a priority queue
- don't sync during critical part of the app
- don't sync more than 2 at same time
- sync when opening a specific account
- sync a selected account in the send/receive modals
- refresh countervalue in send

room for improvment around this:

withLibcore() need to lock less often. typically we lock the whole libcore during a http call triggered by a sync. this blocks the Sign Transaction on device to happen. we also need to be able to interrupt a sync, which is not yet implemented.


meriadec:
this is impressive stuff

PR #484, opened 2018-06-08
Fix back button going to the wrong step on send flow

MortalKastor:
🐜


MortalKastor:
I thought it was merged so I deleted my branch, which close the PR 😅

PR #495, opened 2018-06-10
Pluralizing account summary description on the dashboard

mmaton:


gre:
thanks for the PR, but for plurals, we might want to try to use crowdin feature https://blog.crowdin.com/2016/11/09/icu-syntax-in-crowdin/#plural

gre:
I wonder if there is a better way to do plural with Crowdin 🤔

meriadec:
Yep there is a [built-in way to do that](https://www.i18next.com/translation-function/plurals) with i18next.

You can find usage on the Dashboard, for example ([usage](https://github.com/LedgerHQ/ledger-live-desktop/blob/312fbae7e92bbbf7043c8d97276e3dd938f925c1/src/components/DashboardPage/index.js#L106) and [translation](https://github.com/LedgerHQ/ledger-live-desktop/blob/312fbae7e92bbbf7043c8d97276e3dd938f925c1/static/i18n/en/dashboard.yml#L6))

mmaton:
Ah! That looks elegant. Funnily that's the only usage & it doesn't seem to work as stock. I'll look into it though later today. Thanks for the help guys, I love your work on this app!

meriadec:
@mmaton didn't understood at first, but you're right, the built-in way wasn't working :smile: 

fixed [here](https://github.com/LedgerHQ/ledger-live-desktop/commit/e17ca0f716dd0c4d820e4bd4cc1e76ef3cef149c)
relatable: [i18next PluralResolver.js](https://github.com/i18next/i18next/blob/9871516aa0fb0e5b12953bd6ba192a4f7cf2c980/src/PluralResolver.js#L141-L151)

closing this one as it's fixed now

mmaton:
Thanks, good find!

PR #496, opened 2018-06-10
Font fallback

mmaton:
Updated language packs, set fallback font family on style helper for those of us who don't have access to Museo Sans in our development environment


meriadec:
Thanks!

PR #500, opened 2018-06-11
Better error message precision

gre:
- Receive: differenciate between wrong app & wrong account
- unify isCurrencyAppOpened back into getAddress (making it more strong for bitcoin related apps)


src/components/DeviceConnect/index.js
```diff
@@ -311,14 +309,7 @@ class DeviceConnect extends PureComponent<Props> {
...
             <Box shrink selectable>
-              {accountName ? (
-                <Trans i18nKey="deviceConnect:info" parent="div">
-                  {'You must use the device associated to the account '}
-                  <strong>{accountName}</strong>
-                </Trans>
-              ) : (
-                String(errorMessage || '')
-              )}
+              {String(errorMessage || '')}
```
meriadec: so, no more markup inside error message? isn't it a regression?

src/components/EnsureDeviceApp/index.js
```diff
@@ -116,31 +117,42 @@ class EnsureDeviceApp extends PureComponent<Props, State> {
...
+          .catch(e => {
+            if (
+              e &&
+              (e.name === 'TransportStatusError' ||
+                // we don't want these error to appear (caused by usb disconnect..)
+                e.message === 'could not read from HID device' ||
+                e.message === 'Cannot write to HID device')
+            ) {
+              logger.log(e)
+              throw new Error(`You must open application ‘${cur.name}’ on the device`)
```
meriadec: is it intended to use french quotes here?

gre: not really. what do you think we should use?

gre: ““ maybe ?

gre: ''
""
‘‘
““

meriadec: After all, let's keep it like that. It's pretty elegant :nail_care:

src/components/DeviceConnect/index.js
gre: we just let the error flow whatever it is, which allow more error cases to appear. we'll just try to skip errors we don't want but ultimately we have more precise error

gre: BTW in future, we'll make this displaying on only one line and with ellipsis `...`. the error also need to be selectable

PR #501, opened 2018-06-11
Optimize performance of OperationsList

gre:


src/components/OperationsList/AccountCell.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+class AccountCell extends PureComponent<Props> {
+  render() {
+    const { currency, accountName } = this.props
+    const Icon = getCryptoCurrencyIcon(currency)
+    return (
+      <Cell horizontal flow={2}>
+        <Box alignItems="center" justifyContent="center" style={{ color: currency.color }}>
+          {Icon && <Icon size={16} />}
```
meriadec: condition on `Icon` can be put directly on containing Box

gre: also, in future, we probably want to always have an <Icon> and no icon would take same space. also probably make that Icon is directly a component exported by live-common and you give me the currency as prop

gre: note that i've not really touched this part, so probably leave it for now & i think we'll refactor to this bigger idea?

PR #502, opened 2018-06-11
ui for firmware update

valpinkman:


src/components/ManagerPage/FirmwareUpdate.js
```diff
@@ -89,31 +92,49 @@ class FirmwareUpdate extends PureComponent<Props, State> {
...
+                fontSize={4}
+                style={{ marginLeft: 'auto', marginRight: 15 }}
+              >
+                {t('manager:latestFirmware', { version: latestFirmware.name })}
+              </Text>
+              <Button primary onClick={this.installFirmware(latestFirmware)}>
+                {t('manager:installFirmware')}
+              </Button>
+            </Fragment>
+          )}
```
gre: it would be great if you could split things into more components.
I've been started doing this in our project, it helps performance, readability and maintainability ;)

for instance here, it could be a component that takes a `firmware` prop (and `device` prop) so you don't need to create a function each time.

valpinkman: I fixed it, please tell me if this is what you were expecting @gre

gre: mmh i think `UpdateButton` could be in a separate file implmented as a Component and that have a local method  `installFirmware =  async () => {` so that relieve a bit the codebase of this component that don't need `  installFirmware = (firmware: FirmwareInfos) => async () => {`

even more simply, you can provably directly access this.state.latestFirmware so you don't need to do this nested function

valpinkman: Okay, fixes are done @gre

PR #504, opened 2018-06-11
Feature/hard reset

meriadec:
This commit make the libcore erase all sqlite data when doing an hard reset.
This make the reset really trustable, because it brings teh user in the exact same config as the first installation.

Note that you can use the `LEDGER_RESET_ALL` env var to trigger it on start.

hard-reset

### DIFF
--- a/src/commands/libcoreHardReset.js
+++ b/src/commands/libcoreHardReset.js
@@ -1,0 +1,20 @@
+// @flow
+
+import { createCommand } from 'helpers/ipc'
+import { Observable } from 'rxjs'
+import withLibcore from 'helpers/withLibcore'
+
+const cmd = createCommand('libcoreHardReset', () =>
+  Observable.create(o => {
+    withLibcore(async (core, njsWalletPool) => {
+      try {
+        njsWalletPool.eraseDataSince(new Date(0))
+        o.complete()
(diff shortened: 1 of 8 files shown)


gre:
i think this is neat we do it like this, because it's like an instant reboot. it's just potentially a bit longer reboot but it's nice.

src/commands/libcoreHardReset.js
```diff
@@ -0,0 +1,22 @@
...
+
+import { createCommand } from 'helpers/ipc'
+import { Observable } from 'rxjs'
+import withLibcore from 'helpers/withLibcore'
+
+const cmd = createCommand('libcoreHardReset', () =>
+  Observable.create(o => {
+    withLibcore(async (core, njsWalletPool) => {
+      try {
+        njsWalletPool.eraseDataSince(new Date(0))
```
gre: the genesis Date

src/commands/libcoreHardReset.js
```diff
@@ -0,0 +1,22 @@
...
+        o.complete()
+      } catch (e) {
+        o.error(e)
+      }
+    })
+    // no unsubscribe
+    return () => {}
```
gre: no need to return it I think

PR #510, opened 2018-06-11
make sync exec on rAF, improve a bit scroll perf (but still not perfect)

gre:


gre:
let's come back later on this

PR #528, opened 2018-06-13
StickyBackToTop

gre:


src/components/StickyBackToTop.js
```diff
@@ -0,0 +1,94 @@
...
+    this.context.getScrollbar(scrollbar => {
+      scrollbar.scrollTo(0, 0, 500)
+    })
+  }
+
+  releaseListener = () => {}
+
+  render() {
+    const { visible } = this.state
+    const el = document.getElementById('sticky-back-to-top-root')
```
meriadec: can even be put in `index.ejs` and have this query done in file scope rather than in render. what dyou think?

gre: no we can't, because modals

gre: but i agree with you, as soon as we refacotr modal to use portal too.

not before this, otherwise we have issue with z-index

meriadec: I don't get it. modals are using portals, currently

meriadec: anyway let's keep it like that, eh.

PR #529, opened 2018-06-13
Polishes/stuff

meriadec:
- Memoize currencySettingsLocaleSelector result
- Prevent WithFeesAPI settings its state after unmount
- Autofocus on send/receive/import modals
- Fix double click on ActivityIndicator
- Unify modals back/close buttons


src/reducers/settings.js
```diff
@@ -177,12 +177,19 @@ export const getOrderAccounts = (state: State) => state.settings.orderAccounts
 
 export const areSettingsLoaded = (state: State) => state.settings.loaded
 
+// TODO: avoid this memoize hack (i thought reselect was doing it for us)
```
gre: maybe we'll look later at things like re-reselect 🤣

PR #539, opened 2018-06-13
add TopGradient for MainSideBar

gre:


src/components/MainSideBar/TopGradient.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+import React, { PureComponent } from 'react'
+import styled from 'styled-components'
+
+const TopGradientBox = styled.div`
+  width: ${p => p.width}px;
```
meriadec: `width 100%` without having the width prop will do the trick

PR #553, opened 2018-06-14
Better logger: DEBUG_COMMANDS DEBUG_DB

gre:
by default, all logs need to all happen in prod (so we have logs) but not in dev unless DEBUG enabled (otherwise too verbose)


gre:
😭

meriadec:
Ahah!

PR #573, opened 2018-06-14
add genuine check call

valpinkman:
Genuine is call is now working BUT:
- The code is not on production server yet
- We are running a proxy for the genuine check
- The url used by createSocketDialog for this call is the temporary one

\o/ it does work though


gre:
@valpinkman so if not in production we should not merge it correct? if so, please close the PR & reopen when ready

valpinkman:
OK @gre I'll reopen when it's in production

PR #594, opened 2018-06-17
add missing padding in Manager & Settings

gre:


meriadec:
Prefer using `pb` shorthand with our theme values

PR #596, opened 2018-06-18
refactor GrowScroll to be performant

gre:
this provides:

- native scroll instead of the one managed by react-smooth-scrollbar (which is actually not very smooth!)
- big performance improvment, no more frozen scroll because it's a native scroll, it does not run on JS thread

tradeoffs are:

- you don't see the scrollbar
- there is no animation for scroll up. because there not offered by native. https://developer.mozilla.org/fr/docs/Web/API/Element/scrollIntoView was looking promising, but it does not animate, even in "smooth" mode.


gre:
yeah

PR #597, opened 2018-06-18
[STAND-BY] Feature: Device interaction

meriadec:
More scalable way to handle simple step-by-step interaction with the device (or other async stuff that needs to be done sequentially).

Each step can either provide a function that returns a promise, or a custom rendered component. 
All data falls in cascade to the next steps.

interaction


src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,225 @@
...
+  ErrorDescContainer,
+  ErrorContainer,
+} from './components'
+
+export type Step = {
+  id: string,
+  title: string | (any => string),
+  desc: any,
+  icon: React$Element<*> | null,
+  run?: any => Promise<any> | { promise: Promise<any>, unsubscribe: void => any },
```
gre: 👍  for `unsubscribe`

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,225 @@
...
+  SpinnerContainer,
+  IconContainer,
+  SuccessContainer,
+  ErrorDescContainer,
+  ErrorContainer,
+} from './components'
+
+export type Step = {
+  id: string,
+  title: string | (any => string),
```
gre: what is `any` ?

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,225 @@
...
+  desc: any,
+  icon: React$Element<*> | null,
+  run?: any => Promise<any> | { promise: Promise<any>, unsubscribe: void => any },
+  render?: (
+    { onSuccess: any => any, onFail: any => any, onRetry: any => any },
+    any,
+  ) => React$Element<*>,
```
gre: use `React$Node` instead of all your `React$Element<*>` . it encapsulate more cases (meaning for icon you don't even need | null )
see https://github.com/facebook/flow/blob/master/lib/react.js#L12

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,225 @@
...
+
+export type Step = {
+  id: string,
+  title: string | (any => string),
+  desc: any,
+  icon: React$Element<*> | null,
+  run?: any => Promise<any> | { promise: Promise<any>, unsubscribe: void => any },
+  render?: (
+    { onSuccess: any => any, onFail: any => any, onRetry: any => any },
+    any,
```
gre: it would be better to try to not have to use `any` , so we have more flow checking

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,221 @@
...
+  safeSetState = (...args: any) => {
+    if (this._unmounted) return
+    this.setState(...args)
+  }
+
+  run = async () => {
+    const { step, data } = this.props
+
+    if (this.__IS_MOUNTED__THX_FOR_REMOVING_COMPONENTWILLMOUNT__) {
+      this.safeSetState({ status: 'running' })
```
gre: i would prefer not to use this `safeSetState` pattern. this is a hack to hide the bigger problem^^ we don't want to silent react errors.

& we really want to stop the run logic if it's unmounted

second reason is safeSetState does not help flow typecheck.

```
if (this._unmounted) return
this.setState(...args)
```

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,221 @@
...
+        // $FlowFixMe SAME THING, JUST TESTED THE MINMS KEY, BUT EH
+        if (d2 - d1 < step.minMs) {
+          // $FlowFixMe nice type checking
+          await delay(step.minMs - (d2 - d1))
+        }
+      }
+      if (res.promise) {
+        this._unsubscribe = res.unsubscribe
+        const realRes = await res.promise
+        this.handleSuccess(realRes)
```
gre: make sure we don't call handleSuccess if it's unmounted

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,221 @@
...
+      >
+        <IconContainer>{step.icon}</IconContainer>
+        <Box py={4} pr={4} justify="center" grow shrink>
+          <Box color={isActive && !isSuccess ? 'dark' : ''} ff="Open Sans|SemiBold">
+            {title}
+          </Box>
+          {step.desc && step.desc}
+          {step.render && (
+            <Box>
+              {step.render(
```
gre: it could be a component?

```
<Step.Render
  onSuccess={this.handleSuccess}
  onFail={this.handleFail}
  onRetry={onRetry}
  data={data}
/>
```

src/components/DeviceInteraction/index.js
```diff
@@ -0,0 +1,122 @@
...
+    if (!isCurrentStep) {
+      return
+    }
+    const data = { ...prevData, [step.id]: res || true }
+    const isLast = stepIndex === steps.length - 1
+    if (isLast) {
+      if (!waitBeforeSuccess) {
+        onSuccess && onSuccess(data)
+      }
+      this.safeSetState({ isSuccess: true, data, showSuccess: !waitBeforeSuccess })
```
gre: i think it would be better to turn this code into a `setState(updater)`

src/components/EnsureDeviceAppInteraction.js
```diff
@@ -0,0 +1,77 @@
...
+      <DeviceInteraction
+        steps={[
+          {
+            id: 'device',
+            title: 'Connect your device',
+            icon: usbIcon,
+            desc: 'Because it is required',
+            minMs: 300,
+            run: () =>
+              createCancelablePolling(500, () => {
```
gre: each step `run` function should be set on the instance & not recreated each time. 
(especially fact it access `this.props` from inside this can be error prone)

PR #602, opened 2018-06-18
unify all constants

gre:


src/components/base/CopyToClipboard.js
```diff
@@ -1,8 +1,9 @@
 // @flow
+import { STORYBOOK_ENV } from 'config/constants'
 
 let clipboard = null
 
-if (!process.env.STORYBOOK_ENV) {
+if (!STORYBOOK_ENV) {
```
meriadec: will not work with how storybook webpack handle that (that's why i changed it this morning)

src/components/base/Print.js
```diff
@@ -1,11 +1,12 @@
 // @flow
 
+import { STORYBOOK_ENV } from 'config/constants'
 import { PureComponent } from 'react'
 import qs from 'qs'
 
 let BrowserWindow = null
 
-if (!process.env.STORYBOOK_ENV) {
+if (!STORYBOOK_ENV) {
```
meriadec: same

PR #607, opened 2018-06-18
Reload translations mapped to `ctrl + l`

MortalKastor:


static/i18n/en/app.yml
```diff
@@ -360,3 +360,4 @@ crash:
   restart: Restart app
   reset: Reset app files
   createTicket: Create ticket
+testMe: Just a dummy string to test language file reloading
```
meriadec: :p

gre:
make sense

src/components/modals/Debug.js
```diff
@@ -92,7 +92,7 @@ class Debug extends Component<*, *> {
           <ModalBody onClose={onClose}>
             <ModalTitle>developer internal tools</ModalTitle>
             <ModalContent>
-              <Box style={{ height: 60, overflow: 'auto' }}>
+              <Box>
```
gre: i would prefer we keep the scrolling thing because it allow to see the logs at same time

src/components/layout/Default.js
```diff
@@ -32,11 +32,22 @@ const Main = styled(GrowScroll).attrs({
...
 class Default extends Component<Props> {
   componentDidMount() {
+    const kbShortcut = event => {
+      if (event.ctrlKey && event.key === 'l') {
+        this.props.i18n.reloadResources()
+      }
+    }
+
     window.requestAnimationFrame(() => (this._timeout = setTimeout(() => window.onAppReady(), 300)))
+    window.removeEventListener('keydown', kbShortcut) // Prevents adding multiple listeners when hot reloading
```
gre: this is not going to work. `kbShortcut` is a new reference. instead you want to do this in DidUnmount

PR #621, opened 2018-06-19
Fix ripple and other rendering bugs

gre:


src/components/TranslatedError.js
```diff
@@ -18,7 +19,12 @@ class TranslatedError extends PureComponent<Props> {
     const { t, error } = this.props
     if (!error) return null
     if (typeof error === 'string') return error
-    return t(`errors:${error.name}`, error)
+    const translation = t(`errors:${error.name}`, error)
+    if (translation) {
+      return translation
+    }
+    logger.warn('TranslatedError: no transation!', error.name, error)
+    return error.message || error.name || ''
```
meriadec: what about returning "Unknown error" (or similar) instead of empty string

PR #622, opened 2018-06-19
Awesome new icon 😻

MortalKastor:
No need to keep different icon files, the build tool should convert `icon.png` to each OS relevant format.


meriadec:
awesome

PR #623, opened 2018-06-19
initial version of a Placeholder for when no countervalues loaded yet

gre:


src/components/Placeholder.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components'
+
+export const PlaceholderLine = styled.div`
+  background-color: ${p => (p.dark ? '#C2C2C2' : '#D6D6D6')};
```
meriadec: we may have theme vars for that

gre: but it's a placholder

gre: and we don't have these colors

gre: but i can try

PR #625, opened 2018-06-19
Polishes

meriadec:
- Fix empty state text for add account
- Hide top bar vertical bar when no accounts
- Adjust the Selects empty option texts
- Cleanify the AccountPage settings button
- Ability to import accounts even if sync is not finished yet


gre:
> Ability to import accounts even if sync is not finished yet

mmh i'm actually not sure this should be done like implemented. I think we need a first action that is "Stop scanning" before NEXT is actually clickable. otherwise, it's too easy to click next and you think you have everything

PR #630, opened 2018-06-19
Minor fixes branch

NastiaS:


src/components/base/Spoiler/index.js
```diff
@@ -41,15 +41,17 @@ class Spoiler extends PureComponent<Props, State> {
...
     const { isOpened } = this.state
     return (
       <Fragment>
-        <Box horizontal flow={1} color="dark" align="center">
+        <Box horizontal flow={1} color="dark" align="center" {...p}>
           <IconContainer isOpened={isOpened}>
             <IconChevronRight size={12} />
           </IconContainer>
-          <Title onClick={this.toggle}>{title}</Title>
+          <Title {...p} onClick={this.toggle}>
```
gre: this is not a so good practice to explode `...p`, especially if it goes into multiple parts, but that's ok for now

PR #638, opened 2018-06-20
Improve crash screen design and behaviour

meriadec:
2018-06-20_968x599


meriadec:
Ye I know :/ design don't include it, but you're right we can put it anyway (like hard reset button :p)

meriadec:
2018-06-20_892x699

gre:
sounds good. i hope people won't send us this screenshot without the expanded Show Error 🤣 i would prefer to see everything... i wonder what design thing . anyway it's ok for now

gre:
it's ok for now, let's merge this

meriadec:
We can't prevent people from being stupid

PR #647, opened 2018-06-20
Pixelpushgrekhalil

gre:


src/components/base/GrowScroll/index.js
```diff
@@ -11,6 +12,8 @@ type Props = {
 
 export const GrowScrollContext = React.createContext()
 
+const scrollbarWidth = measureScrollbar()
```
meriadec: :'(

src/components/base/GrowScroll/index.js
```diff
@@ -47,7 +50,7 @@ class GrowScroll extends PureComponent<Props> {
 
     const scrollContainerStyles = {
       overflowY: 'scroll',
-      marginRight: `-80px`,
+      marginRight: `-${80 + scrollbarWidth}px`,
```
meriadec: use directly `marginRight: -80 + scrollbarWidth`

gre: not the same thing tho it's more like `-80-scrollbarWidth` , i think it's less readable tho,

PR #649, opened 2018-06-20
Fix hard reset which is async now, update ledger-core to 2.0.0-rc.2

KhalilBellakrid:


gre:
https://github.com/KhalilBellakrid/ledger-live-desktop/pull/2/files

PR #656, opened 2018-06-21
Display error tooltip in ActivityIndicator

meriadec:
error

### DIFF
--- a/src/components/TopBar/ActivityIndicator.js
+++ b/src/components/TopBar/ActivityIndicator.js
@@ -14,6 +14,8 @@ import { BridgeSyncConsumer } from 'bridge/BridgeSyncContext'
 import CounterValues from 'helpers/countervalues'
 
 import { Rotating } from 'components/base/Spinner'
+import Tooltip from 'components/base/Tooltip'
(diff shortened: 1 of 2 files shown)


src/components/TopBar/ActivityIndicator.js
```diff
@@ -123,6 +126,23 @@ class ActivityIndicatorInner extends PureComponent<Props, State> {
...
     )
+
+    if (error) {
+      return (
+        <Tooltip
+          tooltipBg="alertRed"
+          render={() => (
+            <Box fontSize={4} p={2} style={{ maxWidth: 250 }}>
+              <TranslatedError error={error} />
+            </Box>
```
gre: could use a variable as well because the inner don't change

gre: actually maybe you don't need a render function , just a dom , like children ?

gre: it's ok for now, i'll merge

meriadec: +1 we could refactor all tooltips passing nodes directly

PR #683, opened 2018-06-22
add confirm to hard reset on the crash screen

NastiaS:
I tested brining up a confirmation modal for the hard reset by throwing an error from the init and 2 components inside the app. see if there are other scenarios to test

### DIFF
--- a/src/components/RenderError.js
+++ b/src/components/RenderError.js
@@ -14,6 +14,9 @@ import ExportLogsBtn from 'components/ExportLogsBtn'
 import Box from 'components/base/Box'
 import Space from 'components/base/Space'
 import Button from 'components/base/Button'
+import IconTriangleWarning from 'icons/TriangleWarning'
+import ConfirmModal from './base/Modal/ConfirmModal'
+import { IconWrapperCircle } from './SettingsPage/sections/Profile'
 
 type Props = {
   error: Error,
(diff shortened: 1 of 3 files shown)


src/components/RenderError.js
```diff
@@ -22,11 +25,33 @@ type Props = {
...
+> {
   state = {
     isHardResetting: false,
+    isHardResetModalOpened: false,
   }
 
+  handleOpenHardResetModal = () => this.setState({ isHardResetModalOpened: true })
+  handleCloseHardResetModal = () => this.setState({ isHardResetModalOpened: false })
+
+  handleHardReset = async () => {
```
meriadec: you forgot to remove the other `handleHardReset` method :p

NastiaS: I certainly did 🙄and neither lint nor something else complained..

src/components/RenderError.js
```diff
@@ -14,6 +14,9 @@ import ExportLogsBtn from 'components/ExportLogsBtn'
 import Box from 'components/base/Box'
 import Space from 'components/base/Space'
 import Button from 'components/base/Button'
+import IconTriangleWarning from 'icons/TriangleWarning'
+import ConfirmModal from './base/Modal/ConfirmModal'
```
meriadec: better to keep consistency with `components/base/Modal/ConfirmModal` ..

meriadec: lets merge and we'll change them all later

NastiaS: not again 😁

PR #688, opened 2018-06-24
Native deps building failproofing

MortalKastor:


scripts/compile.sh
```diff
@@ -2,11 +2,11 @@
 
 set -e
 
-export GIT_REVISION=`git rev-parse HEAD`
-export SENTRY_URL=https://db8f5b9b021048d4a401f045371701cb@sentry.io/274561
+GIT_REVISION=`git rev-parse HEAD`
```
gre: mmh i'm pretty sure we need `export`

gre: in bash if you define a variable, and you don't export it, is it going to be accessible by commands? I thought it was never accessible until `export` 🤔

meriadec: it will be defined during your script execution, but will not alter your env
I think it's what we want for this case, because for example we don't want to set NODE_ENV globally

.npmrc
```diff
@@ -0,0 +1,3 @@
+npm_config_target=1.8.7
+npm_config_disturl=https://atom.io/download/electron
+npm_config_runtime=electron
```
meriadec: \n

PR #690, opened 2018-06-25
add default version for node with .nvmrc

valpinkman:
active using 
```
nvm use 
```

after you've cd in the project folder


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
meriadec: \n

PR #694, opened 2018-06-25
remove mangling when building rendered in production

valpinkman:


webpack/main.config.js
```diff
@@ -4,7 +4,7 @@ const rules = require('./rules')
 
 const config = {
   mode: __ENV__,
-  plugins: plugins('main'),
+  plugins: [...plugins('main')],
```
meriadec: put back old line :p

valpinkman: done

PR #701, opened 2018-06-25
Add 'Outdated' concept

gre:
this reflect more the reality of the 'up to date' as well as the 'synchronisation' state. we sometimes don't see the spinner even if sync happen in background, i think we always want spinner, even if it's not triggered by user. I also think we need 'up to date' to only appear when it's actually "up to date" meaning the last sync is close in term of timestamp (is close by a block time + some safe delta).

Outdated rarely happen: it's only if you have not yet pulled for a time and you had no network condition. because otherwise, our pull interval is always smaller than the time we consider 'outdated' (set at block time + 5mn to be safe)
In case an error occur during the sync, we won't show it until one of the account is also outdated, which i think will lighten a bit the cases we have errors. we can decide later to be smarter and more verbose (like being able to know which account have a problem, which one is not sync etc..) but we'll need to have a design for it.

so most of the time you have Outdated AND not network, you will see outdated for a few sec, and then "No network" error.

notes from my previous todo:
> Up to date, indicator thing really need to reflect reality. if i do a clean cache, it's quickly showing up to date, yet i see my op list changing over the seconds... it's important we tell the user the truth otherwise he will panic "where are my funds, etc, it's up to date but i don't have them". ALGO:
>  - if there is at least one pending account in sync, you just show a rotating spinner
>  - otherwise if there is at least one error, you show the error
>  - otherwise if upToDate is true, you should "up to date": a true way to calculate it is: upToDate = allAccounts.every(a => Date.now() - a.lastSyncTime < a.currency.blockAvgTime * 1000 + DELTA) (where DELTA is an extra time that can be a constant. set it to like 5min in ms)
>  - in any other case, we show "Outdated!". NB: THIS WILL RARELY HAPPEN, because we always recurrently will try to pull, however if you don't have network for 10mn, this will happen.

### DIFF
--- a/src/components/TopBar/ActivityIndicator.js
+++ b/src/components/TopBar/ActivityIndicator.js
@@ -10,6 +10,7 @@ import type { T } from 'types/common'
 import type { AsyncState } from 'reducers/bridgeSync'
 
 import { globalSyncStateSelector } from 'reducers/bridgeSync'
+import { isUpToDateSelector } from 'reducers/accounts'
 import { BridgeSyncConsumer } from 'bridge/BridgeSyncContext'
 import CounterValues from 'helpers/countervalues'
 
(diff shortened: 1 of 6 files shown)


gre:
this needs some testing to check all cases & also approval from a design so i'm setting a HODL for now

PR #703, opened 2018-06-25
Fix zombie processes

meriadec:
- Put back `yarn dist:dir` for creating only the unpacked release
- Fix zombie processes by hardly terminate them on window close  …
  - To be improved: this commits introduce regression on OSX behaviour when hiding the window: before it was keeping the app opened, and auto locking it. Now it kills it :D
  - I honestly think it's worth it regarding the problem zombies process causes (LOCK on sqlite, RAM usage, etc.) :ghost:
- Make dev env stop when closing window

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "license": "MIT",
   "scripts": {
     "compile": "bash ./scripts/compile.sh",
-    "dist:dir": "yarn dist --dir -c.compression=store -c.mac.identity=null",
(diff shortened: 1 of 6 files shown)


gre:
🙌🙌

gre:
I think closing on close is not a so bad idea for Mac.. i know it's not how most mac apps works, but in context of ledger app, the fact it's sensitive data etc, i feel it makes sense. like if you are in the middle of a sending flow and you close the app, i don't think it make sense to show it again when you reopen.

PR #704, opened 2018-06-26
Prepare for beta.3

gre:
Many polishes and bugfixes on this beta.3.

Some highlights:

## Features

- installer is now signed on Windows!
- Better error reporting when error are thrown from internal process (serialize/deserialize error objects)
- add a confirm when clicking Reset on crash screen
- Many improvement of manager especially regarding error cases.
- Onboarding animation
- Better sync state
- more reliable app list

## Fixes

- major bugfix on a crash at boot on windows: Fixes zombies process. After quitting the app, some process was still running on Windows which was creating a crash at boot because some db locks was still alive 🔒
- minor bugfix: some device connect/disconnect event where returning null in ledgerjs
- fix amount rounding to occur at critical parts. When we have enough space, we better display the true balance of your account instead of trying to round unsignificant digits. this address confusions when trying to empty your account as well.

## Misc

- "Stepper" refactoring
- clean up images / icons in the project
- Failproof native deps building by explicitly set target
- Prevent env pollution
- renderer bundle is no longer mangle for better error stacktrace
- allow to configure window min-height with LEDGER_MIN_HEIGHT env, until we decide on a default one.


gre:
we might want to get #699 in

PR #714, opened 2018-06-26
Device interaction

meriadec:
interaction

This PR include lot of side-stuff impacted by the DeviceInteraction component.

- Rewrite of the GenuineCheck/GenuineCheckModal
- Rewrite of the EnsureDeviceApp component
- Rewrite of the ReceiveModal

As design has been made by *your servitor*, it may causes some iterations with design team. Anyway, let's ship, and polish later.

:rocket:


src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,223 @@
...
+      this.run()
+    }
+  }
+
+  state = {
+    status: 'idle',
+  }
+
+  componentDidMount() {
+    this.__IS_MOUNTED__THX_FOR_REMOVING_COMPONENTWILLMOUNT__ = true
```
gre: which mean you don't actually need to do this here 🤔

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -0,0 +1,223 @@
...
+    const { onFail, step } = this.props
+    this.setState({ status: 'idle' })
+    onFail(e, step)
+  }
+
+  run = async () => {
+    const { step, data } = this.props
+
+    if (this.__IS_MOUNTED__THX_FOR_REMOVING_COMPONENTWILLMOUNT__) {
+      this.setState({ status: 'running' })
```
gre: in the init case, we actually are already in running case 🤔

meriadec: yeah but we don't want to `setState` if we already are running.
nvm refacto incoming

meriadec:
@gre I think we can merge. I pixel push tomorrow w design.

gre:
let's do this

PR #717, opened 2018-06-26
Wording improvements after review 

dasilvarosa:
- Replace funds by crypto assets
- Improved empty state texts
- minor changes in onboarding

### DIFF
--- a/static/i18n/en/app.yml
+++ b/static/i18n/en/app.yml
@@ -318,7 +318,7 @@ settings: # Always ensure descriptions carry full stops (.)
     softResetDesc: Clear the Ledger Live cache to force resynchronization with the blockchain.
     softReset: Clear
     hardResetTitle: Reset Ledger Live
-    hardResetDesc: Erase all Ledger Live data stored on your computer, including your accounts, transaction history and settings. The private keys to access your crypto assetsin the blockchain remain secure on your Ledger device and on your Recovery sheet.
+    hardResetDesc: Erase all Ledger Live data stored on your computer, including your accounts, transaction history and settings. The private keys to access your crypto assets in the blockchain remain secure on your Ledger device and on your Recovery sheet.
     hardReset: Reset
     developerMode: Developer mode
     developerModeDesc: Show developer apps in the Manager and enable testnet apps.
--- a/static/i18n/en/app.yml
+++ b/static/i18n/en/app.yml
@@ -76,7 +76,7 @@ account:
   lastOperations: Last operations
   emptyState:
     title: No crypto assets yet?
-    desc: Make sure the <1><0>{{currency}}</0></1> app is installed to receive crypto assets # replace [cryptocurrency] and make it bold
+    desc: Receive in this accoung with the the <1><0>{{currency}}</0></1> app in the Manager
     buttons:
       receiveFunds: Receive
   settings:
@@ -213,7 +213,7 @@ manager:
     noDashboard: Navigate to the dashboard on your device (TEMPLATED NEEDED)
     noGenuine: Allow the Manager to continue (TEMPLATE NEEDED)
 receive:
-  title: Receive crypto assets
+  title: Receive
   steps:
     chooseAccount:
       title: Choose account
@@ -233,7 +233,7 @@ receive:
       title: Receive crypto assets
       label: Amount (optional)
 send:
-  title: Send crypto assets
+  title: Send
   totalSpent: Total
   steps:
     amount:
--- a/static/i18n/en/app.yml
+++ b/static/i18n/en/app.yml
@@ -76,7 +76,7 @@ account:
   lastOperations: Last operations
   emptyState:
     title: No crypto assets yet?
-    desc: Receive in this accoung with the the <1><0>{{currency}}</0></1> app in the Manager
+    desc: Ensure the <1><0>{{currency}}</0></1> app is installed and start receiving
     buttons:
       receiveFunds: Receive
   settings:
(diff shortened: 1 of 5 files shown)


static/i18n/en/app.yml
```diff
@@ -75,37 +75,37 @@ account:
   receive: Receive
   lastOperations: Last operations
   emptyState:
-    title: No funds yet?
-    desc: Make sure the <1><0>{{currency}}</0></1> app is installed to receive funds. # replace [cryptocurrency] and make it bold
+    title: No crypto assets yet?
+    desc: Receive in this accoung with the the <1><0>{{currency}}</0></1> app in the Manager
```
gre: typo 'accoung'

static/i18n/en/app.yml
```diff
@@ -283,24 +283,26 @@ settings:
...
-    regionDesc: Choose the region in which you’re located to set the application’s time zone.
+    regionDesc: Choose the region in which you’re located to set the Ledger Live's time zone.
     stock: Regional market indicator
     stockDesc: Choose Western to display an increase in market value in blue. Choose Eastern to display an increase in market value in red.
   currencies:
-    desc: Select a cryptocurrency to edit its settings.
+    desc: Select a cryptocurrency to edit its settings
     exchange: Rate provider ({{ticker}})
-    exchangeDesc: Choose the provider of the base currency exchange rates (via BTC).
+    exchangeDesc: Choose the provider of the base currency exchange rates
```
gre: this was a missing info i think we need to add. it's not a direct Crypto->Fiat but we always use the BTC as intermediary. `(via BTC)` . any other way to phrase it?

static/i18n/en/app.yml
```diff
@@ -283,24 +283,26 @@ settings:
     about: Help
   display:
     desc:
-    language: Language
-    languageDesc: Choose the language to display.
+    language: Display language
+    languageDesc: Set the language displayed in Ledger Live.
     counterValue: Base currency
-    counterValueDesc: Choose the currency to display next to your balance and operations. An exchange is used (via BTC pair).
+    counterValueDesc: Choose the currency to display next to your balance and operations.
```
gre: same here, either put it back but to me we miss a description for the exchange that is used. (see on settings screen)

static/i18n/en/app.yml
```diff
@@ -283,24 +283,26 @@ settings:
...
-    confirmationsToSpendDesc: Set the number of confirmations required for your funds to be spendable. # A higher number of confirmations decreases the probability that a transaction is rejected.
+    confirmationsToSpendDesc: Set the number of network confirmations required for your crypto assets to be spendable. # A higher number of confirmations decreases the probability that a transaction is rejected.
     confirmationsNb: Number of confirmations
-    confirmationsNbDesc: Set the number of blocks a transaction needs to be included in to consider it as confirmed. # A higher number of confirmations increases the certainty that a transaction cannot be reversed.
+    confirmationsNbDesc: Set the number of network confirmations for a transaction to be marked as confirmed. # A higher number of confirmations increases the certainty that a transaction cannot be reversed.
     transactionsFees: Default transaction fees
-    transactionsFeesDesc: Select your default transaction fees. The higher the fee, the quicker the transaction will be processed.
+    transactionsFeesDesc: Select your default transaction fees. The higher the fee, the faster the transaction will be processed.
```
gre: (not used? maybe old settings)

static/i18n/en/app.yml
```diff
@@ -309,17 +311,17 @@ settings:
...
     export: Export logs
-    exportDesc: Exporting Ledger Live logs may be necessary for troubleshooting purposes.
+    exportDesc: Exporting Ledger Live logs may be necessary for troubleshooting purposes
     softResetTitle: Clear cache
-    softResetDesc: Clear the Ledger Live cache to force resynchronization with the blockchain.
+    softResetDesc: Clear the Ledger Live cache to force resynchronization with the blockchain
     softReset: Clear
     hardResetTitle: Reset Ledger Live
-    hardResetDesc: Erase all Ledger Live data stored on your computer, including your profile, accounts, transaction history and settings. The private keys that manage your crypto assets remain secure on your Ledger device.
+    hardResetDesc: Erase all Ledger Live data stored on your computer, including your accounts, transaction history and preferences. The private keys that manage your crypto assets remain secure on your Ledger device.
```
gre: charles mentionned it should be at the end something like: `The private keys that manage your crypto assets remain secure on your Ledger device and on your Recovery sheet.`

static/i18n/en/app.yml
```diff
@@ -309,17 +311,17 @@ settings:
...
     hardReset: Reset
     developerMode: Developer mode
-    developerModeDesc: Show developer apps in the Manager and enable testnet currencies.
+    developerModeDesc: Show developer apps in the Manager
```
gre: please undo the revert (or rephrase)

static/i18n/en/app.yml
```diff
@@ -340,31 +342,31 @@ settings:
...
-  warning_2: Warning 2
-  warning_3: Warning 3
-  warning_4: Warning 4
-  errorMessageIncorrectPassword: The password you entered is incorrect.
-  errorMessageNotMatchingPassword: Passwords don't match.
+  warning_0: Very weak
+  warning_1: Weak
+  warning_2: Medium
+  warning_3: Strong
+  warning_4: Very strong
```
gre: These are not used

static/i18n/en/app.yml
dasilvarosa: This is an important topic which was already planned to be addressed. I'll revert, but it will change, possibly with some added disclaimers. I think we need a few extra tooltips or links to documentation to properly explain it.

static/i18n/en/app.yml
dasilvarosa: Changed to faster for consistency.

static/i18n/en/app.yml
dasilvarosa: ouch. fixed

static/i18n/en/app.yml
dasilvarosa: Good point, will add.

static/i18n/en/app.yml
dasilvarosa: Thought they would be in the future, so why not phrase them

static/i18n/en/app.yml
dasilvarosa: ok

static/i18n/en/app.yml
gre: no but i mean, transactionsFeesDesc is not used. there is no longer such settings available. because fees are always at the time you do a transaction.

so you can remove this for now.

static/i18n/en/app.yml
gre: IMO we need to kill today all wording that are not in the app. 
we should not anticipate future potential feature, because it's gonna be asking more things to translators and we are not sure it's gonna happen.

soon we should do a pass on cleaning up all unused things and also all duplicates ;)

static/i18n/en/app.yml
gre: make sense

static/i18n/en/app.yml
gre: ultimately, maybe we need to have 2 settings like we used to have..

PR #725, opened 2018-06-27
Bugfix account.unit to refresh account page when you change it

gre:
this also start splitting a bit the AccountPage component

as CalculateBalance will memoize things, we want inner part to be potentially re-render if they depend on other things, we do this by connecting them to redux. redux idea is to connect leafs instead of connecting the whole tree when possible, so this fit this paradigm to me.

### Type

Bugfix, code improvment

### Context

bug raised by khalil

### Testing / Part affected

Account page

### DIFF
--- a/src/components/BalanceSummary/index.js
+++ b/src/components/BalanceSummary/index.js
@@ -35,6 +35,7 @@ const BalanceSummary = ({
   selectedTimeRange,
 }: Props) => {
   const account = accounts.length === 1 ? accounts[0] : undefined
+  // FIXME This nesting 😱
   return (
     <Card p={0} py={5}>
       <CalculateBalance accounts={accounts} daysCount={daysCount}>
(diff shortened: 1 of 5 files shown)


gre:
@meri @NastiaS  i'm playing a bit with the template here too :D as the commit message will be prepend to the PR message, i actually think we don't need a "### Description" section because it can literally be the beginning. also Testing, usually we have nothing to say, but it is basically what part of the app is affected which is interesting to know

src/components/BalanceSummary/index.js
```diff
@@ -35,6 +35,7 @@ const BalanceSummary = ({
   selectedTimeRange,
 }: Props) => {
   const account = accounts.length === 1 ? accounts[0] : undefined
+  // FIXME This nesting 😱
```
meriadec: XD

PR #732, opened 2018-06-27
Less intrusive apps list loading

meriadec:
### Type

UI Polish

### Context

Manager: apps list

apps


gre:
I think the background of the placeholder should be the same color as when it's loaded. so white-ish.

meriadec:
apps-alt

Opinion?

PR #739, opened 2018-06-28
Onboarding analytics integration

NastiaS:
Integration of the Segment Analytics into the Onboarding flow 

### Type

Feature

### Context

Expanding bootstrapped analytics to more pages

Onboarding flow

As well as there is a little adjustment included on the DeviceIntaraction fort he Genuine Check part to handle errors

### DIFF
--- a/src/components/Onboarding/steps/GenuineCheck/GenuineCheckUnavailable.js
+++ b/src/components/Onboarding/steps/GenuineCheck/GenuineCheckUnavailable.js
@@ -33,7 +33,7 @@ export function GenuineCheckUnavailableFooter({
         <Button
           padded
           disabled={false}
-          event="Skip Genuine Check Onboarding"
+          event="Onboarding Skip Genuine Check"
           onClick={() => nextStep()}
           mx={2}
         >
@@ -64,7 +64,7 @@ export function GenuineCheckUnavailableMessage({
         underline
         onClick={() => {
           handleOpenGenuineCheckModal()
-          track('Genuine Check Unavailable Retry', {
+          track('Genuine Check Retry', {
             flowType: onboarding.flowType,
             deviceType: onboarding.isLedgerNano ? 'Nano S' : 'Blue',
           })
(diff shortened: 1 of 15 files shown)


src/components/Onboarding/steps/GenuineCheck/GenuineCheckUnavailable.js
```diff
@@ -51,7 +58,18 @@ export function GenuineCheckUnavailableMessage({
...
   return (
     <Box align="center" flow={1} color={colors.alertRed}>
-      <FakeLink ff="Open Sans|Regular" fontSize={4} underline onClick={handleOpenGenuineCheckModal}>
+      <FakeLink
+        ff="Open Sans|Regular"
+        fontSize={4}
+        underline
+        onClick={() => {
+          handleOpenGenuineCheckModal()
+          track('Genuine Check Unavailable Retry', {
```
gre: maybe just 'Genuine Check Retry' ? because if you retry, it's because it was unavailable

src/components/Onboarding/steps/GenuineCheck/GenuineCheckUnavailable.js
```diff
@@ -29,7 +30,13 @@ export function GenuineCheckUnavailableFooter({
         {t('app:common.back')}
       </Button>
       <Box horizontal ml="auto">
-        <Button padded disabled={false} onClick={() => nextStep()} mx={2}>
+        <Button
+          padded
+          disabled={false}
+          event="Skip Genuine Check Onboarding"
```
gre: I think better to prefix Onboarding . 
actually i'm even not sure we need to say it's "onboarding". could just be Skip Genuine Check. because we know what is the current "page"

src/middlewares/analytics.js
```diff
@@ -12,6 +12,8 @@ export default store => next => action => {
     if (shareAnalytics) {
       start(store)
     } else {
+      // TODO add user uuid
+      track('Stopping Analytics')
```
gre: I think remove this hack. it does not work because we stop just after. we'll need to figure out a real solution

src/components/Onboarding/steps/GenuineCheck/GenuineCheckUnavailable.js
NastiaS: yes, also it is more general. Retry is also the case if you disattach the device in the middle

src/middlewares/analytics.js
NastiaS: It fires, i can see it in Segment right before it kills the process

PR #740, opened 2018-06-28
Refactor and Polish Add Accounts

gre:
- one spinner at a time, to reflect better the technical reality of the scanning
- Adding a imported account category to avoid confusion when seeing nothing because all is imported (better to show what accounts than a text. for instance, you can also figure out like that which accounts your device have)
- autofocus the first input because we want more engagement on setting an initial name
- input are editable in place. no more 'edit name' editing state
- account name validation! trimmed, max length + handling empty case
- Add more to import more cryptos from step 4
- the new account take the same naming semantic so the naming are consistent if you leave the default ones and you don't end up with all accounts named 'New Account'
- prefer to grey out when you don't select anything to make sure user understand you have to select things to move on.

### Type

- mostly Polish + minor features

- Add Accounts
- Edit Account, because now sharing the logic for account name validation


src/components/base/AccountsList/AccountRow.js
```diff
@@ -11,108 +11,91 @@ import Radio from 'components/base/Radio'
...
-      onClick(updatedAccount)
-    }
+  handleChangeName = (name: string) => {
+    const { onEditName, account } = this.props
+    if (onEditName) onEditName(account, name)
   }
 
-  handlePreventSubmit = (e: SyntheticEvent<any>) => {
-    // prevent account row to be submitted
+  onClickInput = (e: SyntheticEvent<*>) => {
```
meriadec: we can have unified `handleStopEvent` which will be used in place of `handlePreventSubmit` + `onClickInput` (nvm, not so grave)

src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -14,9 +15,22 @@ import Button from 'components/base/Button'
...
 
 import type { StepProps } from '../index'
 
+const LoadingRow = styled(Box).attrs({
+  horizontal: true,
+  borderRadius: 1,
+  px: 3,
+  align: 'center',
+  justify: 'center',
+  mt: 1,
```
meriadec: more a generic consideration, but imo the margins should be set contextually and not inside components
using `<LoadingRow mt={1} />` make it more reusable, because component is not leaking any margin itself

gre: but the component is in place here.. also i just copy what was before in Row

gre: pushing the logic more far, that means we never want to do `styled(Box).attr` right?

meriadec: no, it's only for margins that should not be part of the component style, but only when we _use_ the component in a page

PR #747, opened 2018-06-28
Pixel polish

meriadec:
### Context

- Account settings modal
  - change button style (larger)
  - same width for name input and unit select
- Buttons focus state now listen correctly the `isGlobalTabEnabled` (like Tabbable), to display outline only when user pressed `<Tab>`

- Accounts settings modal
- All buttons

### DIFF
--- a/src/components/CurrentAddress/index.js
+++ b/src/components/CurrentAddress/index.js
@@ -16,7 +16,7 @@ import Box from 'components/base/Box'
 import CopyToClipboard from 'components/base/CopyToClipboard'
 import QRCode from 'components/base/QRCode'
 
-import IconCheck from 'icons/Check'
+import IconRecheck from 'icons/Recover'
 import IconCopy from 'icons/Copy'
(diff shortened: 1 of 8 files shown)


src/components/CurrentAddress/index.js
```diff
@@ -207,11 +225,26 @@ class CurrentAddress extends PureComponent<Props> {
...
+            />
             <CopyToClipboard
               data={address}
               render={copy => (
-                <FooterButton icon={<IconCopy size={16} />} label="Copy" onClick={copy} />
+                <FooterButton
+                  icon={<IconCopy size={16} />}
+                  label={t('app:common.copy')}
+                  onClick={() => {
+                    this.setState({ copyFeedback: true })
```
gre: (future) could this mecanism abstracted out in a component? (using context for instance)

meriadec: yep, totally. idem for the copy hash just added on operation details.
we need generic way

PR #759, opened 2018-06-29
Polish

meriadec:
- Modify amount color on operation detail when < 0
- Open OperationDetail modal after transaction success
- Correct reset send modal state when pressing back/retry
- Apparently its a good idea to hide tx hash from user
- Display more precise error on transaction confirmation

### DIFF
--- a/src/components/modals/Send/index.js
+++ b/src/components/modals/Send/index.js
@@ -49,7 +49,6 @@ type State<Transaction> = {
   transaction: ?Transaction,
   optimisticOperation: ?Operation,
   isAppOpened: boolean,
-  disabledSteps: number[],
   errorSteps: number[],
   amount: number,
   error: ?Error,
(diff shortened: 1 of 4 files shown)


src/components/modals/Send/index.js
```diff
@@ -165,7 +167,14 @@ class SendModal extends PureComponent<Props, State<*>> {
 
   handleChangeAppOpened = (isAppOpened: boolean) => this.setState({ isAppOpened })
   handleChangeTransaction = transaction => this.setState({ transaction })
-  handleRetry = () => this.setState({ error: null, errorSteps: [] })
+  handleRetry = () => {
+    this.setState({
+      error: null,
+      errorSteps: [],
```
gre: need to remove errorSteps and do it in render() so error on the breadcrump happen at the correct place (3 or 4 based on the error)

PR #775, opened 2018-06-29
smoother progress bar

valpinkman:


src/components/base/Button/index.js
```diff
@@ -221,7 +221,7 @@ class Button extends PureComponent<
     const onClickHandler = e => {
       if (onClick) {
         if (event) {
-          track(event, eventProperties)
+          // track(event, eventProperties)
```
gre: why the change?

valpinkman: oh sorry, we needed to comment that for storybook to work, I will fix this

PR #776, opened 2018-06-30
add SyncContinuouslyPendingOperations

gre:
this implement a system that will continuously trigger a sync on accounts that have "pending operations".
Pending operations are basically the operations that just got broadcasted.

> note for improvement: we should have a better management of the pendingOperations list (for better optimistic updates) and also maybe have it to timeout (by block count? / time.).

### Type

Feature

### Context

requested many times

after you broadcast a transaction, a sync will be triggered in the next < 10s  and will be checked again each 10s until it reaches our API.


src/components/SyncContinouslyPendingOperations.js
```diff
@@ -0,0 +1,65 @@
...
+}> {
+  componentDidMount() {
+    this.timeout = setTimeout(this.check, this.props.interval)
+  }
+  componentWillUnmount() {
+    clearTimeout(this.timeout)
+  }
+  check = () => {
+    const { sync, accounts, priority, interval } = this.props
+    setTimeout(this.check, interval)
```
meriadec: means that we always push new checks even if sync is not finished?
orr is it handled at lower level?

PR #777, opened 2018-06-30
Settings refactoring

gre:
This is a complete refactoring of settings for more modularity (first commit) and then a reordering of the settings section according to what was decided.

### Type

Code Quality + polish

### Context

was decided yesterday during a meeting (Lea, @dasilvarosa, @nicolasbigot )
the tradeoff we agree is to move as requested except the countervalues+exchange we can't move it in currencies because it would be too confusing in the fact the settings are per currency in currencies and we must not introduce global settings for now.

Settings


gre:
test

src/components/SettingsPage/AboutRowItem.js
```diff
@@ -0,0 +1,26 @@
...
+import { shell } from 'electron'
+import IconExternalLink from 'icons/ExternalLink'
+import { Tabbable } from 'components/base/Box'
+import { SettingsSectionRow } from './SettingsSection'
+
+export default class AboutRowItem extends PureComponent<{
+  url: string,
+  title: string,
+  desc: string,
+  url: string,
```
meriadec: Only one url prop is enough I think

src/components/SettingsPage/CounterValueExchangeSelect.js
```diff
@@ -0,0 +1,53 @@
...
+          <SelectExchange
+            small
+            from={intermediaryCurrency}
+            to={counterValueCurrency}
+            exchangeId={counterValueExchange}
+            onChange={this.handleChangeExchange}
+            minWidth={200}
+          />
+        ) : null}
+      </Fragment>
```
meriadec: I think you can get rid of Fragment here. Also, I think it would be better to always render a select (in loading state for example, (to be coded)) instead of null, to prevent any blink

src/components/SettingsPage/DisablePasswordButton.js
```diff
@@ -0,0 +1,133 @@
...
+  unlock,
+  cleanAccountsCache,
+  saveSettings,
+}
+
+type Props = {
+  t: T,
+  unlock: () => void,
+  settings: SettingsState,
+  saveSettings: Function,
```
meriadec: Rare case of seeing @gre using `Function` type

src/components/SettingsPage/LanguageSelect.js
```diff
@@ -0,0 +1,61 @@
...
+        <Select
+          small
+          minWidth={250}
+          isSearchable={false}
+          onChange={this.handleChangeLanguage}
+          renderSelected={item => item && item.name}
+          value={currentLanguage}
+          options={this.languages}
+        />
+      </Fragment>
```
meriadec: No need fragment

src/components/SettingsPage/RegionSelect.js
```diff
@@ -0,0 +1,67 @@
...
+      <Fragment>
+        <Select
+          small
+          minWidth={250}
+          onChange={this.handleChangeRegion}
+          renderSelected={item => item && item.name}
+          value={currentRegion}
+          options={regionsFiltered}
+        />
+      </Fragment>
```
meriadec: Same

src/components/SettingsPage/ResetButton.js
```diff
@@ -0,0 +1,82 @@
...
+    )
+  }
+}
+
+export const IconWrapperCircle = styled(Box).attrs({})`
+  width: 50px;
+  height: 50px;
+  border-radius: 50%;
+  background: #ea2e4919;
+  text-align: -webkit-center;
```
meriadec: webkit-center?

src/components/SettingsPage/ResetButton.js
```diff
@@ -0,0 +1,82 @@
...
+      </Fragment>
+    )
+  }
+}
+
+export const IconWrapperCircle = styled(Box).attrs({})`
+  width: 50px;
+  height: 50px;
+  border-radius: 50%;
+  background: #ea2e4919;
```
meriadec: is this colors outside of theme?

src/components/SettingsPage/ResetButton.js
```diff
@@ -0,0 +1,82 @@
...
+              <IconTriangleWarning width={23} height={21} />
+            </IconWrapperCircle>
+          )}
+        />
+      </Fragment>
+    )
+  }
+}
+
+export const IconWrapperCircle = styled(Box).attrs({})`
```
meriadec: No need empty attrs if you don't use it. Just don't put attrs

src/components/SettingsPage/CounterValueExchangeSelect.js
gre: you're right. i know. i know. but we'll keep them because very soon. like in a few days. we'll have to actually `<Track/>` this, so we'll need it :P

agree for the blink tho ;) i'll fix

src/components/SettingsPage/DisablePasswordButton.js
gre: actually all the code of this PR : I have not written a single line. i just dup the section files into a lot of pieces but preserve almost the same. simplified what i could find.

and this part is actually not refactored. it shouldn't take `settings` and `saveSettings` into parameter. this is heresy ! who knows what `settings` structure is ;)

src/components/SettingsPage/LanguageSelect.js
gre: will keep it to add `<Track/>`

src/components/SettingsPage/RegionSelect.js
gre: will keep it to add `<Track/>`

src/components/SettingsPage/ResetButton.js
gre: 🤷‍♂️

src/components/SettingsPage/ResetButton.js
gre: 🤷‍♂️

src/components/SettingsPage/ResetButton.js
gre: ok. but again i have not written this 🤣

src/components/SettingsPage/CounterValueExchangeSelect.js
meriadec: Ooh ok got it :ok_hand:

PR #780, opened 2018-06-30
bump all patch version to fix a case where React is in dup!

gre:
yarn update-interactive # is a BAD idea.. it does not fix the dups !!

this however, does not fix the current bug we have with unexpected remounting

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -22,11 +22,11 @@
   dependencies:
     "@babel/highlight" "7.0.0-beta.44"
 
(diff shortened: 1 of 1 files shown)


meriadec:
What alternative do we have to `yarn upgrade-interactive`? Re-generating `yarn.lock` with `package.json`? Editing it manually? :doge:

gre:
let's not merge anything for now. we'll have to regen the yarn because changing lot of pkg to fix the babel 7 bug

gre:
@meriadec  good point. i think in future we will have to double check we have no dup. i think we should have a check in the compile script that error if there are react dups

PR #782, opened 2018-06-30
Remove UglifyJS and explicitely opt-out from it

gre:
#### TLDR: Uglify ES is to AVOID for all our projects. this create very sneaky bugs with React.

we have finally figured out the sneaky bug we have found weeks ago.. our recent "particle animation" really revealed it and it was easier to investigate. Each time we spawn a confetti, all the siblings were remounting like shown on this gif..

out2

----

We had a very tricky to notice bug that was happening **only in production** because of uglify.

The issue was that in some case, uglify was transforming our clean component to crap like this:

https://gist.github.com/gre/faf39d26c04c5d1408ea88bf6a1cc851

seriously. look at this. **2 levels of this crap:**

```
class extends Component {
  ...
  render() {
    return React.createElement(
      class extends Component {
        ...
        render() {
          return React.createElement(
            class extends Component {
               render() {

            }
          )
      }
  }
}
```

in our sourcecode, the components are flatten not nested like this... seems like Uglify have inlined a bit too much things here..

which means, each time the parent re-render, you actually create **a new Component class**, so of course it's gonna force React to unmount and remount because it's how the reconcilier works (and it's definitely correct tradeoff for the algorithm to be efficient)..

Anyway, just be careful about Uglify and React people, and especially the fact Webpack 4 use it **by default**.

have we gone too far? I think so... I don't expect a build tool to change my logic...

> My opinion: Uglify is not a good idea to use anyway. why would you want to make your code **ugly**, especially for a electron app. JS is like 1% of the executable size 🤣. also we want clear error stacktrace.

### Type

CRITICAL BUG FIX

### Context

weeks of investigation

the whole app

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     }
   },
   "resolutions": {
-    "uglify-es": "3.3.7",
     "webpack-sources": "1.0.1"
   },
   "dependencies": {
@@ -157,7 +156,6 @@
     "prettier": "^1.13.5",
     "react-hot-loader": "^4.3.2",
     "react-test-renderer": "^16.4.1",
-    "uglifyjs-webpack-plugin": "^1.2.6",
     "webpack": "^4.6.0",
     "webpack-bundle-analyzer": "^2.11.1",
     "webpack-cli": "^2.0.14",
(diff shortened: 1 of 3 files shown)


meriadec:
nb: is it intended to have that big -1000/+1000 changes on yarn.lock? seems that it has been entirely re-generated

PR #784, opened 2018-06-30
Send: polishes

gre:
- Ethereum gasLimit field to not fallback on empty
- Bitcoin fees field to handle Custom properly
- Fix many race conditions in Send step1

### Type

bugfixes

### Context

friday meeting from user feedback

send, step 1

### DIFF
--- a/src/components/base/InputCurrency/index.js
+++ b/src/components/base/InputCurrency/index.js
@@ -190,6 +190,16 @@ class InputCurrency extends PureComponent<Props, State> {
     )
   }
 
+  select = () => {
+    // TODO we should fowardRef so InputCurrency ref is on Input
+    this.input && this.input.select()
+  }
+
+  input: ?Input
+  onRef = (input: ?Input) => {
+    this.input = input
+  }
+
   render() {
     const { renderRight, showAllDigits, unit, subMagnitude } = this.props
     const { displayValue } = this.state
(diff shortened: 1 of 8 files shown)


src/components/base/InputCurrency/index.js
```diff
@@ -198,6 +208,7 @@ class InputCurrency extends PureComponent<Props, State> {
       <Input
         {...this.props}
         ff="Rubik"
+        ref={this.onRef}
```
meriadec: is it intended to not use `React.createRef` here, like in `BitcoinKind.js`?
in your opinion, should we refacto all our refs with it?

src/components/modals/Send/steps/01-step-amount.js
```diff
@@ -128,9 +130,9 @@ export class StepAmountFooter extends PureComponent<
 
     try {
       const totalSpent = await bridge.getTotalSpent(account, transaction)
-      if (this._isUnmounted) return
+      if (syncId !== this.syncId) return
       const canBeSpent = await bridge.canBeSpent(account, transaction)
-      if (this._isUnmounted) return
+      if (syncId !== this.syncId) return
```
meriadec: future: dyou think we could make these syncId checks more generic, maybe using context API, or redux? would be awesome to avoid those checks in components

PR #786, opened 2018-07-01
Wording

dasilvarosa:
Error messages conform to title and description format. To be cleaned and reviewed with @gre 

### Type

Wording

Global error messages

### DIFF
--- a/static/i18n/en/errors.yml
+++ b/static/i18n/en/errors.yml
@@ -1,7 +1,7 @@
 # the error codes are alphabetically sorted
 BtcUnmatchedApp:
-  title: 'That's the wrong app'
-  description: 'Open the ‘{{currencyName}}’ app on your device'
+  title: That's the wrong app
+  description: Open the ‘{{currencyName}}’ app on your device
 DeviceNotGenuine:
   title: Possibly not genuine
   description: 'Request Ledger Support assistance.'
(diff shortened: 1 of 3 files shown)


static/i18n/en/errors.yml
```diff
@@ -1,97 +1,97 @@
...
 NetworkDown:
-  title: 'Your internet connection seems down.'
-  description: ''
+  title: Oops, internet seems down
+  description: Please check your internet connection.
 NoAddressesFound:
-  title: 'No accounts were found.'
-  description: ''
+  title: Sorry, no accounts found
+  description:  # Missing context
```
gre: it should not happen normally, but this is when scanning and libcore fails to calculate the addresses 🤷‍♂️

static/i18n/en/errors.yml
```diff
@@ -1,97 +1,97 @@
...
 NotEnoughBalance:
-  title: 'Insufficient funds to proceed.'
-  description: ''
+  title: Oops, not enough balance
+  description: Make sure the account to debit has sufficient balance
 RangeError:
-  title: '{{message}}'
-  description: ''
+  title: '{{message}}' # Missing context
```
gre: these is no context. this is one of javascript error that man happen on numbers

static/i18n/en/errors.yml
```diff
@@ -1,97 +1,97 @@
...
+  description:
 TimeoutError:
-  title: 'The request timed out.'
-  description: ''
+  title: Oops, a time out occurred
+  description: It took too long for the server to respond. #Retry / check status?
```
gre: yeah, usually we will offer a Retry in place. it depends where (ofr instance for Sync it's the sync button etc. etcc)

PR #792, opened 2018-07-02
Fix GrowScroll style (fixes onboarding step 1 centering)

meriadec:
set the inner container to be flex column displayed.
that allow children to grow, so it gives vertical centering + scroll for free to children

### Type

not really a bug fix, not really an improvement

### Context

whole app. but done for onboarding step 1


gre:
ok, let's deeply test this still behave same everywhere ;) and should be good

PR #795, opened 2018-07-02
Prevent loading fs via analytics in storybook

meriadec:
it prevents loading fs via analytics in storybook

### Type

bug fix

### Context

storybook dev & build


gre:
shouldn't we "mock" the analytics/* import for tests/storybook instead of branching inside it?

meriadec:
thought about it but I would choose mocking at the lower level possible (we could even do the branching in `helpers/db`, to avoid electron-storage to be loaded, and so we keep logging etc. like it is). wdyt?

PR #797, opened 2018-07-02
Prevent check recipient valid if empty

meriadec:
And add a catch to fallback + warning if check throwed

### Type

bug fix

### Context

on opening the send modal, a check was done directly to see if recipient is valid. at least on linux, this check - when done by libcore - is making internal process crash (SIGSEGV)

this commit only prevent checking empty recipient, and also add warning if this case occur (it will, in case of bad address)

### DIFF
--- a/src/components/modals/Send/fields/RecipientField.js
+++ b/src/components/modals/Send/fields/RecipientField.js
@@ -3,7 +3,6 @@ import React, { Component } from 'react'
 import type { Account } from '@ledgerhq/live-common/lib/types'
 import type { T } from 'types/common'
 import type { WalletBridge } from 'bridge/types'
-import logger from 'logger'
 import Box from 'components/base/Box'
 import Label from 'components/base/Label'
 import LabelInfoTooltip from 'components/base/LabelInfoTooltip'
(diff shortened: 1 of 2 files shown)


src/components/modals/Send/fields/RecipientField.js
```diff
@@ -39,12 +40,16 @@ class RecipientField<Transaction> extends Component<Props<Transaction>, { isVali
...
-    this.setState({ isValid })
+    try {
+      const isValid = await bridge.isRecipientValid(
+        account.currency,
+        bridge.getTransactionRecipient(account, transaction),
+      )
+      if (syncId !== this.syncId) return
+      this.setState({ isValid })
+    } catch (err) {
+      logger.warn(`Can't check if recipient is valid`, err)
```
gre: please undo this because in this case we want the promise to crash so it gets reported to Sentry / display a crash screen / etc..

gre: basically we should never try catch without reporting, our approach so far is "let it crash" because if we want to do defensive programming, well, we'll have to do this *everywhere* )

gre: ( and react error boundaries can help when needed )

meriadec: as it is done in a callback, initiated by didmount/didupdate, it does not bubble up to crash screen. without warning, it's just a plain ugly "uncaught promise fail" in console.

what dyou think we should do in that case?

meriadec: ok for removing the try/catch
nb: it will fail uncaught on invalid addresses

gre: the unhandled case is handled by Sentry so yeah it's still better i think

PR #811, opened 2018-07-02
properly fix the 'internal become zombie' issue

gre:
To be tested everywhere but i think that's it. I've double check BEFORE and AFTER my fix:

before i was able to `kill -9`  all Ledger* processes and the internal process was still up forever (on mac).

after, when i `kill -9` the internal also disappear. 

I've also renamed Internal to Ledger Live Internal so we can see it nicely

fixed

(This should also fix the remaining "LOCK" issue! I hope 🙏 )

### Type

Critical Bug Fix

### Context

bug raised by @nicolasbigot & continuation of work on 'terminator', good bye terminator.

the whole app when you quit / reopen / when it crash / it should always restore properly


meriadec:
That's interesting, but on my side it's a regression :smile:
Now it doesnt quit anything anymore: main process AND internal process are still up after closing window, in DEV and PROD mode.

e.g 

```bash
# dev mode
$ yarn start
# ... wait for launch
# ... close the window
$ ps aux | grep 'Ledger Live Internal' | grep -v grep
meri      6949  7.8  0.5 1054768 95216 pts/0   Sl+  09:49   0:21 Ledger Live Internal
```

:x: for meh


ezgif-5-a142c60301

gre:
wtf 😭

PR #812, opened 2018-07-03
Use winston to provide better logs and logfiles

gre:
- all the logs from the 3 processes (main, renderer, internal) are redirected to a log file.
- Export logs feature export logs from these 3 processes too!
- the log file get rotated each day in logs folder of the app directory. ( https://github.com/winstonjs/winston-daily-rotate-file )
- in prod, the logs won't appear in the console anymore. unless you enable a special flag LOGS_IN_CONSOLE .
- the Export feature will now just reuse the last log files so we have all the logs from all processes.

### Type

Debugability Improvement

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -51,6 +51,7 @@
     "no-restricted-syntax": 0,
     "no-prototype-builtins": 0,
     "no-void": 0,
+    "no-lonely-if": 0,
     "react/forbid-prop-types": 0,
     "react/jsx-curly-brace-presence": 0,
     "react/jsx-filename-extension": 0,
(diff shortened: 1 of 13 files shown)


.eslintrc
```diff
@@ -51,6 +51,7 @@
     "no-restricted-syntax": 0,
     "no-prototype-builtins": 0,
     "no-void": 0,
+    "no-lonely-if": 0,
```
meriadec: I totally understand why disabling for `logger.js` if/else blocks :) , but that rule is pretty neat I think. What about disabling it locally for particular use cases?

PR #820, opened 2018-07-03
Downgrade broken concurrently version

meriadec:
relatable: https://github.com/kimmobrunfeldt/concurrently/issues/147 :clap: 

This was causing trouble in dev mode, when inputing anything in the shell.


meriadec:
re-generating `yarn.lock` make concurrently silently update :/

PR #827, opened 2018-07-03
Minor fixes branch

NastiaS:
Genuine Check timeout after 30 sec of waiting to be able to Skip or Retry
### Type

Addition
### Context

On windows it could take especially long and this is a way out

Where genuine check is in the app

### DIFF
--- a/src/components/DeviceInteraction/DeviceInteractionStep.js
+++ b/src/components/DeviceInteraction/DeviceInteractionStep.js
@@ -11,12 +11,14 @@ import {
   IconContainer,
   SuccessContainer,
   ErrorContainer,
+  TimeoutContainer,
 } from './components'
 
 export type Step = {
   id: string,
   title?: React$Node | (Object => React$Node),
   desc?: React$Node,
+  timeout?: number,
   icon: React$Node,
   run?: Object => Promise<any> | { promise: Promise<any>, unsubscribe: void => any },
   render?: ({ onSuccess: Object => any, onFail: Error => void }, any) => React$Node,
@@ -37,6 +39,7 @@ type Props = {
   step: Step,
   onSuccess: (any, Step) => any,
   onFail: (Error, Step) => any,
+  onTimeoutClick: () => any,
   data: any,
 }
 
@@ -44,6 +47,7 @@ class DeviceInteractionStep extends PureComponent<
   Props,
   {
     status: Status,
+    isTimedOut: boolean,
   },
 > {
   static defaultProps = {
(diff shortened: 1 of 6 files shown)


src/components/GenuineCheck.js
```diff
@@ -134,6 +134,7 @@ class GenuineCheck extends PureComponent<Props> {
           </Trans>
         ),
         icon: genuineCheckIcon,
+        timeout: 30000,
```
meriadec: let's define that in a constant (if not already the case)

src/components/DeviceInteraction/DeviceInteractionStep.js
```diff
@@ -120,8 +124,21 @@ class DeviceInteractionStep extends PureComponent<
     try {
       const d1 = Date.now()
 
-      // $FlowFixMe JUST TESTED THE `run` 6 LINES BEFORE!!!
-      const res = (await step.run(data)) || {}
+      let res
+
+      if (step.timeout) {
+        const timeoutPromise = new Promise(r => setTimeout(() => r('timeout'), step.timeout))
```
gre: - you can do `r(new TimeoutError())` directly
- the timeout is never cancelled which i think might be a problem. there is two case we could cancel it: if step unmount, if job finishes. 

> BTW it would be much better if later we move the "job" to use observable paradigm, it have all of these out of the box, `.pipe(timeout(..))`

src/components/GenuineCheck.js
gre: so wait, there is something i'm not sure to understand here.

the timeout is not supposed to interrupt the step.

the feature is supposed to just be a warning, but the task will continue.

the genuine timeout currently throw 2min after, but here,  it's just to implement a lil' message after 30s 🤔  

if this is what you have implemented (which i don't see it is 🤔  ) i think it should be renamed. 

we basically want ability to "notify" , maybe each second or at a given rate 🤔

gre:
I don't see this implemented what was required but I might be wrong

gre:
we've discussed with Anastasia and we misunderstood the feature. also the feature is a minor request and i think it needs a wider solution.. for instance, technically we know the "nonce" increment of the genuine check progress. if we know how much total step is takes, we could have some sort of progress bar, or a pourcentage , or whatever.. but it needs a design, so i would definitely postpone this post-release

src/components/GenuineCheck.js
meriadec: So it should not be an error, and should not provide "skip" feature?

PR #828, opened 2018-07-03
Implement new flow of receive

gre:
Re-implement step 3 and 4 of Receive.

✅  pass security requirements
✅  pixel pushed
❌  reworded

### Type

Feature

### Context

Security was 🙅‍♂️  on the previous feature implementation

Send

### DIFF
--- a/src/components/modals/Receive/steps/04-step-receive-funds.js
+++ b/src/components/modals/Receive/steps/04-step-receive-funds.js
@@ -4,24 +4,50 @@ import invariant from 'invariant'
 import React, { PureComponent } from 'react'
 
 import TrackPage from 'analytics/TrackPage'
+import getAddress from 'commands/getAddress'
+import { isSegwitAccount } from 'helpers/bip32'
 import Box from 'components/base/Box'
-import Label from 'components/base/Label'
 import CurrentAddressForAccount from 'components/CurrentAddressForAccount'
-import RequestAmount from 'components/RequestAmount'
+import { WrongDeviceForAccount } from 'components/EnsureDeviceApp'
 
-import type { StepProps } from '../index'
+import type { StepProps } from '..'
 
-type State = {
-  amount: number,
-}
-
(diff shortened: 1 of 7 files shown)


src/components/modals/Receive/steps/04-step-receive-funds.js
```diff
@@ -4,55 +4,69 @@ import invariant from 'invariant'
...
+      transitionTo('receive')
+    } catch (err) {
+      onChangeAddressVerified(false, err)
+      this.props.transitionTo('confirm')
+    }
   }
 
-  handleChangeAmount = (amount: number) => this.setState({ amount })
   handleGoPrev = () => {
+    // FIXME this is not a good practice at all. it triggers tons of setState. these are even concurrent setState potentially in future React :o
```
meriadec: Yeah.. but didn't really find decent alternative for now :/ At first, parent passed its own `setState` method to `StepProps`, but it rapidly appears as weird design (even if it solves the pb). For now it works pretty well because used on small views.

Wdyt of `ReactDOM.unstable_batchedUpdates`? (relatable [facebook/react#10231](https://github.com/facebook/react/issues/10231#issuecomment-316644950))

Dyou think of better approach?

gre: I think we could just pass a prop that represent exactly what this action is about. exactly like on Redux, it would just "one action" and just "multiple reducers treat it". here it can be just one callback as prop, and just one setState that change what needs.

what fundamentally need to be reworked however, is to be able to "control" everything. for instance, the stepper does not allow to have its step controlled controlled. basically it shouldn't have a `transitionTo`, it should just be an internal detail that is called during componentDidUpdate and when stepId changes. 🤔

once this, we can control everything from the "top level" that handle the business logic (the modal body level) , top-down data flow basically and one state at same place.

meriadec: sounds good

PR #829, opened 2018-07-03
Anonymize Sentry

gre:
make sure all data that is sent to sentry get anonymized before sending.

(i know code is very verbose today but this is very safe code. with help of Flow there is no possibility of runtime type error)

### Type

mandatory legal requirement

### DIFF
--- a/src/sentry/install.js
+++ b/src/sentry/install.js
@@ -1,4 +1,6 @@
 // @flow
+import anonymizer from 'helpers/anonymizer'
+/* eslint-disable no-continue */
 
 require('../env')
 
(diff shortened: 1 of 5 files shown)


src/sentry/install.js
```diff
@@ -11,6 +13,60 @@ export default (Raven: any, shouldSendCallback: () => boolean, userId: string) =
     tags: { git_commit: __GIT_REVISION__ },
     environment: __DEV__ ? 'development' : 'production',
     shouldSendCallback,
+    dataCallback: (data: mixed) => {
```
meriadec: :sunglasses: :+1: :rocket:

gre: this is annoying mode of Flow but at least safe. I was looking at alternative like `lodash update` but i have no experience in it and no idea if it's safe

meriadec: yeah. totally agree on using those kind of vanilla/quickly fixable helpers for any sensitive part

PR #830, opened 2018-07-03
centralize all the support urls to a same place

gre:


gre:
this one is for you @dasilvarosa  xD

PR #833, opened 2018-07-04
Make logger && db env dependent - fixes storybook

meriadec:
prevent `fs` from being loaded by storybook

### Type

bugfix

### Context

we need to ensure, each time we include native/file-system-dependent dependency inside a component/helper/whatever that it does not affect the storybook. separating those concerns will help a lot if some day we plan to do a Ledger Live web version.


gre:
is there a way this can be solved at the build level. kinda how React Native do with the .ios and .android extension?

I don't really like the fact it's not a static check but it's a runtime check, also need things to be in bundle.

anyway, it's ok for now

meriadec:
Yeah I think it's possible and definitely better. I think webpack strip plugin may be suitable for those cases

PR #835, opened 2018-07-04
Wording: add accounts and new receive flow 

dasilvarosa:

### DIFF
--- a/static/i18n/en/app.yml
+++ b/static/i18n/en/app.yml
@@ -37,8 +37,8 @@ common:
   verify: Verify
   copy: Copy
   copyAddress: Copy address
-  copied: Copied!
-  addressCopied: Address copied!
+  copied: Copied
+  addressCopied: Address copied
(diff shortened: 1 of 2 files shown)


static/i18n/en/app.yml
```diff
@@ -254,8 +254,8 @@ receive:
       withoutDevice: Don't have your device?
     confirmAddress:
       title: Verification
-      action: Confirm address on device
-      text: To receive cryptoassets, confirm the address on your device. Click the button below to reveal your {{currencyName}} address.
+      action: Verify address on device
+      text: A {{currencyName}} receive address will be displayed on your device. Carefully verify that it matches the address on your computer.
```
gre: the part about "on your computer" is a bit unprecise to me. if you scan the QR code, it's not "on the computer" anymore. is there a way to say "where the address has been pasted/scanned" or something 🤔

PR #847, opened 2018-07-04
About section

meriadec:
move some infos from `Help` to a new `About` section in Settings.
Discussed with design, could "do the taff" until we have that infamous app information window.


meriadec:
@dasilvarosa can you provide wording for description of the `About` tab?

2018-07-04_860x336

meriadec:
merging for now (not a big deal)

PR #849, opened 2018-07-04
use makeSingleInstance to make app a singleton

gre:
we need to test it a bit


gre:
ok @meriadec  will retry this but only with makeSingleInstance

PR #852, opened 2018-07-04
minimal error handling for mvp

valpinkman:


src/components/ManagerPage/FirmwareUpdate.js
```diff
@@ -102,27 +102,26 @@ class FirmwareUpdate extends PureComponent<Props, State> {
...
+    const result = await installOsuFirmware
       .send({ devicePath: device.path, firmware: latestFirmware, targetId: deviceInfo.targetId })
       .toPromise()
-    return success
-  }
 
-  installFinalFirmware = async (device: Device) => {
-    const { success } = await installFinalFirmware.send({ devicePath: device.path }).toPromise()
-    return success
+    return result
```
gre: it looks like you used to return `success` but now you return `result` 🤔  
i guess you don't use the result value anyway?

src/components/ManagerPage/FirmwareUpdate.js
```diff
@@ -102,27 +102,26 @@ class FirmwareUpdate extends PureComponent<Props, State> {
...
+
+  flashMCU = async (device: Device) => installMcu.send({ devicePath: device.path }).toPromise()
 
   handleCloseModal = () => this.setState({ modal: 'closed' })
 
   handleDisclaimerModal = () => this.setState({ modal: 'disclaimer' })
   handleInstallModal = (stepId: StepId = 'idCheck', shouldFlash?: boolean) =>
     this.setState({ modal: 'install', stepId, shouldFlash, ready: true })
 
+  handleDisclairmerNext = () => this.setState({ modal: 'install' })
```
gre: (minor, typo)

src/components/modals/UpdateFirmware/Disclaimer.js
```diff
@@ -69,7 +69,7 @@ class DisclaimerModal extends PureComponent<Props, State> {
                 <GradientBox />
               </ModalContent>
               <ModalFooter horizontal justifyContent="flex-end" style={{ width: '100%' }}>
-                <Button primary onClick={goToNextStep}>
+                <Button primary onClick={() => goToNextStep()}>
```
gre: you can rollback this lil change since now we have a dedicated callback

static/i18n/en/errors.yml
```diff
@@ -62,7 +62,7 @@ ManagerDeviceLocked:
 ManagerNotEnoughSpace:
   title: Sorry, insufficient device storage
   description: Uninstall some apps to increase available storage and try again.
-ManagerUnexpectedError:
+ManagerUnexpected:
```
gre: we might remove it , right? no longer used as we preserve the original error?

src/helpers/firmware/installMcu.js
```diff
@@ -5,6 +5,20 @@ import { WS_MCU } from 'helpers/urls'
...
 import getNextMCU from 'helpers/firmware/getNextMCU'
 import getDeviceInfo from 'helpers/devices/getDeviceInfo'
+import { createCustomErrorClass } from 'helpers/errors'
+
+const ManagerDeviceLockedError = createCustomErrorClass('ManagerDeviceLocked')
+
+function remapSocketError(promise) {
+  return promise.catch((e: Error) => {
+    switch (true) {
+      case e.message.endsWith('6982'):
```
gre: (note for future, better way to do it is using the error field: https://github.com/LedgerHQ/ledger-live-desktop/blob/708eaa5e4492a077c58bda766396a34fedbd3d08/src/commands/getAddress.js#L35-L42

src/helpers/firmware/installOsuFirmware.js
```diff
@@ -43,7 +42,6 @@ export default async (
     await remapError(createDeviceSocket(transport, url).toPromise())
     return { success: true }
   } catch (error) {
-    const result = { success: false, error }
-    throw result
+    throw error
```
gre: surounding try catch can be removed too? note that the logic have changed too (no longer throwing the object, which was a bad idea anyway

gre:
can't merge until we're sure the firmware update still work. i might try a bit tonight to fix things if i can

PR #858, opened 2018-07-04
Add HookDeviceChange + FreezeDeviceChangeEvents

gre:
Add a mechanism that will make the Manager reload if a device unmount or if it changes and only if it's not in a critical modal. It will also trigger when you leave the critical modal and state had changed during the process. This should make the "leaving the upgrade firmware" reload the manager, which is what we want, instead of current behavior that was leaving you in an ugly empty state.

### Type

Improvement

### Context

MVP

Manager

### DIFF
--- a/src/components/ManagerPage/AppsList.js
+++ b/src/components/ManagerPage/AppsList.js
@@ -34,6 +34,7 @@ import Update from 'icons/Update'
 import Trash from 'icons/Trash'
 import CheckCircle from 'icons/CheckCircle'
 
+import { FreezeDeviceChangeEvents } from './HookDeviceChange'
 import ManagerApp, { Container as FakeManagerAppContainer } from './ManagerApp'
 import AppSearchBar from './AppSearchBar'
 
@@ -170,6 +171,7 @@ class AppsList extends PureComponent<Props, State> {
         isOpened={status !== 'idle' && status !== 'loading'}
         render={() => (
           <ModalBody align="center" justify="center" style={{ height: 300 }}>
+            <FreezeDeviceChangeEvents />
             {status === 'busy' || status === 'idle' ? (
               <Fragment>
                 <ModalTitle>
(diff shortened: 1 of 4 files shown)


gre:
to be tested with Firmware update

src/components/ManagerPage/HookDeviceChange.js
```diff
@@ -0,0 +1,69 @@
+// @flow
+
+import { PureComponent } from 'react'
+import { createStructuredSelector } from 'reselect'
+import { connect } from 'react-redux'
+import { getCurrentDevice } from 'reducers/devices'
+import type { Device } from 'types/common'
+
+const hookDeviceChangeInstances = []
```
gre: who needs context anyway

src/components/ManagerPage/AppsList.js
```diff
@@ -34,6 +34,7 @@ import Update from 'icons/Update'
 import Trash from 'icons/Trash'
 import CheckCircle from 'icons/CheckCircle'
 
+import { FreezeDeviceChangeEvents } from './HookDeviceChange'
```
meriadec: we may want it somewhere else to be used in other parts of the app

src/components/ManagerPage/HookDeviceChange.js
```diff
@@ -0,0 +1,69 @@
...
+  render() {
+    return null
+  }
+}
+
+/* eslint-disable react/no-multi-comp */
+class HookDeviceChange extends PureComponent<{
+  device: ?Device,
+  onDeviceDisconnected: () => void,
+  onDeviceChanges: Device => void,
```
meriadec: `onDeviceChange` may be more consistent with other `onChange` etc.

(importance: LOW :p)

src/components/ManagerPage/AppsList.js
gre: but the behavior only happen in Manager 🤔

meriadec: yeah nvm

PR #860, opened 2018-07-04
Systematically replace all `process.cwd()` by just a '.'

gre:


gre:
wait for it. needs to do something better than cwd(). basically need to anonymize the username in the filepath. maybe some regexp.

@meriadec  can you tell me what a `process.cwd()` looks like for you?

meriadec:
it behaves like OSX, it returns the expected path. for ex, printing it on my machine gave me
```
/home/<user>/<some-folders>/ledger-live-desktop
```

gre:
@meriadec  i see. i need to do both a replace for cwd and the config path to be safe i guess

meriadec:
makes sense
you can get it like that

```
import { app } from 'electron'
console.log(app.getPath('userData'))
```

gre:
@meriadec  yeah i know. but there are 4 cases to consider tho lol. (storybook, internal, main, renderer) they each have a different way to do this XD

PR #863, opened 2018-07-05
They will never stop (polish again)

meriadec:
micro polishes. the last ones.


khalilbenihoud:
🙈

meriadec:
we have the cross icon on the modal

PR #868, opened 2018-07-05
manager firware card title and icon dynamic for nano and blue

NastiaS:
Manager Page the Fimrware card had hardcoded title for the device and Nano Icon. now it is different for Blue

### Type

Bug fix, UI polish

Manager page/Firmware card

### DIFF
--- a/static/i18n/en/app.yml
+++ b/static/i18n/en/app.yml
@@ -218,6 +218,8 @@ manager:
     help: Check on your device to see which apps are already installed
   firmware:
     installed: 'Firmware version {{version}}'
+    titleNano: Ledger Nano S
+    titleBlue: Ledger Blue
     update: Update firmware
     continue: Continue update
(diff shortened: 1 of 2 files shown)


static/i18n/en/app.yml
```diff
@@ -218,6 +218,8 @@ manager:
     help: Check on your device to see which apps are already installed
   firmware:
     installed: 'Firmware version {{version}}'
+    titleNano: Ledger Nano S
+    titleBlue: Ledger Blue
```
meriadec: sry didnt understood your question earlier, but if it's not used anywhere we may not put it in translation file

gre: yeah, it's not to be translated.

meriadec: >< oops missed the usage. LGTM

PR #880, opened 2018-07-06
missing some errors

gre:
later we really need to figure out how to always fallback on generic for the error we don't really need to dup in wording


gre:
( ok at this point we can merge this. i think, we have many recent change to test and this is a minor one )

PR #882, opened 2018-07-06
Linux icon

meriadec:
bc some ppl needs icons :trollface: 

2018-07-06-104411_2560x1440_escrotum
2018-07-06-104400_2560x1440_escrotum


src/main/app.js
```diff
@@ -65,6 +66,11 @@ const saveWindowSettings = window => {
 }
 
 const defaultWindowOptions = {
+  // Linux require window to have icon to be correctly displayed
+  // in docks, when using <C-tab>, etc.
+  // see https://github.com/electron-userland/electron-builder/issues/2269<Paste>
+  icon: i('linux-app-icon-512x512.png'),
```
gre: how will this affect windows and mac ? shouldn't it be if linux?

meriadec: it will not affect them, will just not be used.
will it make more sense if I rename it `browser-window-icon-512x512.png` ?

gre: yeah maybe rename it

PR #883, opened 2018-07-06
Reworded update flow, now going to check for Ledger Blue

dasilvarosa:

### DIFF
--- a/static/i18n/en/app.yml
+++ b/static/i18n/en/app.yml
@@ -220,27 +220,27 @@ manager:
     installed: 'Firmware version {{version}}'
     titleNano: Ledger Nano S
     titleBlue: Ledger Blue
-    update: Update firmware
-    continue: Continue update
+    update: Update
+    continue: Continue
     latest: 'Firmware version {{version}} is available'
-    disclaimerTitle: 'You are about to install the latest <1><0>firmware {{version}}</0></1>'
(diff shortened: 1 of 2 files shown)


static/i18n/en/app.yml
```diff
@@ -220,27 +220,27 @@ manager:
...
     titleBlue: Ledger Blue
-    update: Update firmware
-    continue: Continue update
+    update: Update
+    continue: Continue
     latest: 'Firmware version {{version}} is available'
-    disclaimerTitle: 'You are about to install the latest <1><0>firmware {{version}}</0></1>'
-    disclaimerAppDelete: Please note that all the apps installed on your device will be deleted.
-    disclaimerAppReinstall: You will be able to re-install your apps after the firmware update
+    disclaimerTitle: 'You are about to install <1><0>firmware version {{version}}</0></1>'
```
meriadec: to be double-checked, but I think if we change wording containing `<1>some<0>tags</0></1>` we have to change also in the component. @dasilvarosa did you tested it works with `yarn start` ?

valpinkman:
LGTM

static/i18n/en/app.yml
gre: oh interesting. so basically wordings that use `<1>` need to reflect exactly in the code?

PR #884, opened 2018-07-06
Fix recursion anonymizer

gre:


meriadec:
WOULD BE COOL TO PUT TYPE ANNOTATION

PR #897, opened 2018-07-06
exit from mcu flashing state and add wording to loading state

valpinkman:
Flash MCU fixes and wording

### Type

Fixes

### Context

MVP

Firmware Update Modals


gre:
ok, we'll need someone to test the next rc we'll do ;)

PR #902, opened 2018-07-07
Fix error imprecision in all Manager calls

gre:
the code is equivalent except errors thrown don't get decorated back into an 'Error' leading to imprecision
we'll need to double check all is fine but this will *make errors great again*.
this is really an antipattern we have to avoid.


meriadec:
M.E.G.A

PR #919, opened 2018-07-08
Fix #917

### LINKED ISSUE #917: two last point on graph use same countervalue

- Ledger Live **rc.2**
- Platform: **mac**

#### Expected behavior

the two last point countervalues should be different like other points

#### Actual behavior

two last points are identically

a

meriadec:
#917


gre:
@meriadec can you do a PR with just `Correct date format for chart Tooltip d95e832` ? it appears the bug is gone. it must have been a bug in countervalue api 🤔

PR #927, opened 2018-07-09
Put back the original architecture image for now

meriadec:
Let's keep the old architecture image, because its purpose is to rapidly explain the architecture, and despite the fact that this one is less pretty, it's more explainful, and that's its job


gre:
please rollback in the readme too. we'll merge for now

gre:
unless we have a better schema in time

meriadec:
it's the same img name

gre:
AH yes

Issue #932, opened 2018-07-09
UI regression polish


NastiaS:
#### Ledger Live Version and Operating System

- Ledger Live 1.0.0
- Platform: mac

#### Expected behavior

The arrow pointing to the Add Account should be repositioned 
#### Actual behavior

The arrow goes over the text 
#### Steps to reproduce the behavior

Has to be Empty state - no accounts yet to be able to see the initial instruction 

<img width="226" alt="" src="https://user-images.githubusercontent.com/10082039/42443784-bcbcd6c4-836e-11e8-98ac-98a81a5615a5.png">

PR #938, opened 2018-07-09
Fix for issue #932 - UI polish sideBar

NastiaS:
Fix for the UI regression (Arrow over text) after wording adjustment on the Empty State sideBar under Add account
### Type

UI polish regression
### Context

Issue #932 

Empty state - no accounts
screen shot 2018-07-09 at 14 10 03

### DIFF
--- a/src/components/MainSideBar/index.js
+++ b/src/components/MainSideBar/index.js
@@ -68,8 +68,8 @@ class MainSideBar extends PureComponent<Props> {
 
   ADD_ACCOUNT_EMPTY_STATE = (
     <Box relative>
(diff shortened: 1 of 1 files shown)


src/components/MainSideBar/index.js
```diff
@@ -68,8 +68,8 @@ class MainSideBar extends PureComponent<Props> {
 
   ADD_ACCOUNT_EMPTY_STATE = (
     <Box relative>
```
gre: Mmh i think you can just add padding right here and don’t need extra markup

PR #958, opened 2018-07-09
Remove length limitation on password

### LINKED ISSUE #953: Password Length > 20

- Ledger Live **1.0.0**
- Platform: **linux**

#### Expected behavior

Not that much a bug, but it seems that there is a 20 characters limit on the password length (since nothing appears on the screen if I type more). 
So when I type more than 20 characters,  should'nt I be warned?

#### Actual behavior

(At fist I though the (unseen) extra characters were taken into account for the password, but they are not. my bad!) 
Still, I'd like to be warned about the password policy

During the password creation, enter more than 20 characters,

meriadec:
closes #953

let's wait for next release to _really test this_ :doge:

### DIFF
--- a/src/components/base/InputPassword/index.js
+++ b/src/components/base/InputPassword/index.js
@@ -47,7 +47,6 @@ type State = {
 }
 
 type Props = {
-  maxLength: number,
   onChange: Function,
   t: T,
   value: string,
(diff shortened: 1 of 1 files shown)


gre:
we basically need to make sure we don't constraint at any place. we need to test using a big password and see if we can pass all things (unlock, change password, reset a long password)

PR #1024, opened 2018-07-10
One package per plaftorm: Linux/AppImage macOS/dmg Win/nsis

### LINKED ISSUE #972: Clicking "Update now" closes app but does nothing else

- Ledger Live 1.0.0
- Platform: Windows 10

#### Expected behavior

The Portfolio section tells me there is an update available with a link to "Update now" which when clicked should close the app, update the app, and then relaunch the app with the newly updated version.

#### Actual behavior

Clicking "Update now" just closes the app and nothing else seems to happen.

1. Download and unpack ledger-live-desktop-1.0.0-win.zip.
2. Launch Ledger Live.exe and complete initial setup.
3. Go to Portfolio section of app and wait for update notification to appear.
4. Click "Update now" and see the app close itself.
5. ...
6. Nothing seems to be happening.
7. Relaunching Ledger Live.exe manually shows it is still on v1.0.0.

meriadec:
The zip archives caused some trouble with auto-updates

closes #972

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -10,6 +10,7 @@ mac:
   category: public.app-category.wallet
   target:
     - dmg
+    - zip
 
 appImage:
   systemIntegration: doNotAsk
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -10,7 +10,6 @@ mac:
   category: public.app-category.wallet
   target:
     - dmg
-    - zip
 
 appImage:
   systemIntegration: doNotAsk
@@ -34,9 +33,6 @@ win:
     - sha256
   target:
     - target: nsis
-      arch:
-        - x64
-    - target: zip
       arch:
         - x64
 


gre:
@meriadec  is this solving both for mac and windows? isn't the zip used by the autoupdate itself? any documentation on the subject?

meriadec:
pfieew.. that's right, for macOS it's apparently [required](https://www.electron.build/auto-update) for Squirrel.Mac to works properly.

```
zip target for macOS is required for Squirrel.Mac, otherwise latest-mac.yml
cannot be created, which causes autoUpdater error. 
```

fixed.

gre:
and windows it's not required?

meriadec:
nsis is the [default target configuration](https://www.electron.build/configuration/win) for Win, and `dmg + zip` is the default target configuration for macOS, so it should be optimal config for auto update

PR #1026, opened 2018-07-10
More robust implementation of Error fallback

gre:
- in future we won't anymore see "unworded" error cases
- it will always fallback to the generic case if error is not worded which is what we want (like the previous Error wording)

### Type

improvement

### Context

this is to avoid this kind of weird error:

image 2 2

all errors


meriadec:
my eyes just exploded

gre:
@meriadec  not my fault is folks out there are literally **screen** - **shotting** 🤣

PR #1060, opened 2018-07-11
Always patch operations if some contained `blockHeight: null`

### LINKED ISSUE #1000: Incominc BTC confirmed TX looks unconfirmed

- Ledger Live 1.0.1
- Platform: windows 7 enterprise SP1

#### Expected behavior

I got an incoming transaction on a BTC account that already has more than 60 confirmations, verified on external web explorers and on Ledger Chrome App

#### Actual behavior

Ledger live app says "Not confirmed (0)" since more rthan 10 hours

meriadec:
closes #1000

### DIFF
--- a/src/bridge/LibcoreBridge.js
+++ b/src/bridge/LibcoreBridge.js
@@ -28,6 +28,8 @@ type Transaction = {
 const decodeOperation = (encodedAccount, rawOp) =>
   decodeAccount({ ...encodedAccount, operations: [rawOp] }).operations[0]
 
+const hadSomeNullBlockHeight = ops => ops.some(op => op.blockHeight === null)
+
 const EditFees = ({ account, onChange, value }: EditProps<Transaction>) => (
   <FeesBitcoinKind
     onChange={feePerByte => {
@@ -135,6 +137,7 @@ const LibcoreBridge: WalletBridge<Transaction> = {
 
             const hasChanged =
               accountOps.length !== syncedOps.length || // size change, we do a full refresh for now...
+              hadSomeNullBlockHeight(accountOps) || // a bit berzerk, but in the future we will want to patch ops 1 by 1
               (accountOps.length > 0 &&
                 syncedOps.length > 0 &&
                 (accountOps[0].accountId !== syncedOps[0].accountId ||


src/bridge/LibcoreBridge.js
```diff
@@ -135,6 +137,7 @@ const LibcoreBridge: WalletBridge<Transaction> = {
 
             const hasChanged =
               accountOps.length !== syncedOps.length || // size change, we do a full refresh for now...
+              hadSomeNullBlockHeight(accountOps) || // a bit berzerk, but in the future we will want to patch ops 1 by 1
```
gre: I don't think this is necessary.

```
accountOps[0].blockHeight !== syncedOps[0].blockHeight ||
```

to set after line 143 is enough

gre: also, blockHeight being null is a legit case (it happens on ETH, it is before a node confirm you) and we don't want to always replace again and again if the last one remains blockHeight=null.

meriadec: I don't get how it will solves the problem if ever the incorrect transaction index is `> 0`. IMO it's not worth the risk. Anyway, just made the change ¯\_(ツ)_/¯

src/bridge/LibcoreBridge.js
```diff
@@ -138,7 +138,8 @@ const LibcoreBridge: WalletBridge<Transaction> = {
               (accountOps.length > 0 &&
                 syncedOps.length > 0 &&
                 (accountOps[0].accountId !== syncedOps[0].accountId ||
-                  accountOps[0].id !== syncedOps[0].id)) // if same size, only check if the last item has changed.
+                  accountOps[0].id !== syncedOps[0].id)) || // if same size, only check if the last item has changed.
+              accountOps[0].blockHeight !== syncedOps[0].blockHeight
```
gre: this needs to be inside the 

```
accountOps.length > 0 &&
                 syncedOps.length > 0 &&
```

src/bridge/LibcoreBridge.js
gre: true. but it's a bit less likely. we just want light checks for now before we have a "better algorithm". another approach would be "for each block that was null before, check if the new one, by id, is still null, if not, need a refresh" , which is different from "one is null" which always trigger a refresh of everything at each sync

src/bridge/LibcoreBridge.js
meriadec: oops. rebased thx
was convinced that flow could catch [this kind of access](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzADcwBeMAbQF0KAWABioDoEn8mpUg)

alfonmga:
When you guys are planning to release this fix to production?

PR #1065, opened 2018-07-11
Fixes #1061

### LINKED ISSUE #1061: Show last 5 release notes in the update modal
Just after an update, or when click on the "Details" button
image
a modal with release notes is shown. It should show the last 5 release notes instead of just the last one.

gre:
current release note would display:

capture d ecran 2018-07-11 a 19 16 32

### Type

improvement

### Context

#1061 @EricLarch 

release notes

### DIFF
--- a/src/components/base/Markdown/index.js
+++ b/src/components/base/Markdown/index.js
@@ -11,6 +11,7 @@ export const Notes = styled(Box).attrs({
   fontSize: 4,
   color: 'smoke',
   flow: 4,
+  mb: 6,
 })`
   ul,
   ol {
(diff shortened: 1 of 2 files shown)


src/components/base/Markdown/index.js
```diff
@@ -11,6 +11,7 @@ export const Notes = styled(Box).attrs({
   fontSize: 4,
   color: 'smoke',
   flow: 4,
+  mb: 6,
```
meriadec: margins are contextual and should not leak from component.
prefer using `mb={6}` on `<Notes />` directly inside `ReleaseNotesBody` component

PR #1067, opened 2018-07-11
Fix a bug with error.message not exposed to t()

gre:
This fix is really weird, but some Errors, like the one returned by ripple don't have `message` to be enumerable, meaning `t()` don't work.

to reproduce, simply try to import XRP accounts without network.

**before this PR**

capture d ecran 2018-07-11 a 20 18 47

**after this PR**

capture d ecran 2018-07-11 a 20 18 27

### Type

bugfix

### DIFF
--- a/src/components/TranslatedError.js
+++ b/src/components/TranslatedError.js
@@ -30,14 +30,16 @@ class TranslatedError extends PureComponent<Props> {
       }
       return null
     }
+    // $FlowFixMe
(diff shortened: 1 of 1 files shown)


gre:
Mmh title is supposed to be used as mandatory part. Description is the optional one. We want to display message in « title » because it’s the most used one. Most of the time we just show the title not the description. Maybe we need a better semantic

PR #1091, opened 2018-07-12
Fixes #1085

### LINKED ISSUE #1085: Add a button to open the User Data folder in Settings/Help
a button to open the User Data folder in Settings/Help would be really useful in the future when we ask our user to manually go delete a folder. like the logs folder, etc..

it needs to work on all OS. basically like a `open $FOLDER`

meriadec:
closes #1085

2018-07-12_703x146


meriadec:
@dasilvarosa we may need wording/desc polish here

src/commands/getUserDataFolder.js
```diff
@@ -0,0 +1,15 @@
...
+import { fromPromise } from 'rxjs/observable/fromPromise'
+
+import { createCommand, Command } from 'helpers/ipc'
+
+type Input = {}
+
+type Result = string
+
+const cmd: Command<Input, Result> = createCommand('getUserDataFolder', () =>
+  fromPromise(new Promise(resolve => resolve(process.env.LEDGER_CONFIG_DIRECTORY))),
```
gre: does it really need to go through command tho?
you can directly do `electron.app.getPath('userData')` from renderer side

it's what is done for the logger https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/helpers/resolveLogsDirectory.js

meriadec: Ah! :joy:  my bad, I first used `import { app } from 'electron'` (which was obviously not working in renderer), but I forgot about `import { remote } from 'electron'` then `remote.app`. Good catch. Fixing.

gre:
we can review wording tomorrow

PR #1093, opened 2018-07-12
Fixes #904

### LINKED ISSUE #904: ZEC overwinter support
related https://github.com/LedgerHQ/ledger-wallet-chrome/pull/109

- [x] libcore integration
- [x] ledgerjs integration
- [x] live integration

meriadec:
fixes #904

### DIFF
--- a/src/commands/libcoreSignAndBroadcast.js
+++ b/src/commands/libcoreSignAndBroadcast.js
@@ -77,7 +77,9 @@ async function signTransaction({
   hasTimestamp: boolean,
 }) {
   const additionals = []
+  let expiryHeight
   if (currencyId === 'bitcoin_cash' || currencyId === 'bitcoin_gold') additionals.push('bip143')
+  if (currencyId === 'zcash') expiryHeight = Buffer.from([0x00, 0x00, 0x00, 0x00])
   const rawInputs = transaction.getInputs()
 
   const inputs = await Promise.all(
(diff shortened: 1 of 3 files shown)


src/commands/libcoreSignAndBroadcast.js
```diff
@@ -119,7 +121,7 @@ async function signTransaction({
 
   const changePath = output ? output.getDerivationPath().toString() : undefined
   const outputScriptHex = Buffer.from(transaction.serializeOutputs()).toString('hex')
-  const lockTime = transaction.getLockTime()
+  const lockTime = undefined // TODO: transaction.getLockTime()
```
gre: why is this commented out? could you explain a bit what it means? wouldn't this create a regression on some coin that need it? 🤔

meriadec: seen with the libcore team: this field is not handled yet on their side, it always returns a default value, and we realized that it was what make Zcash tx fail :thinking: as it is not necessary for other coins, we choose to comment it for all, but if you think it's safer, we can set it `undefined` just for Zcash. nb: I tested on other coins (stratis, dash, ..)

gre: so is ZCash working right now? 🤔

meriadec: [yep](https://explorer.zcha.in/transactions/667e86a8b696a5cb60753d66ac1b900078e4e909f32da89e333afee3433984c3)

meriadec: will add a comment in the PR. hodl on

Issue #1055, opened 2018-07-11
Add the possibility to launch the onboarding once again


EricLarch:
A user can have more than one device. Relaunching the onboarding is necessary to either check the integrity of further devices, or get help to restore/initialize another device.

This option could be available from a button located in the settings.

EricLarch:
In the section **Help** of the app settings, add at the bottom a new section.

**Onboarding**
Launch again the onboarding to add a new device to the Ledger Live application

Button: **Launch**

gre:
@NastiaS can you assess the implication of such a feature?
I suspect strange thing will happen on Password screen..

IMO such a feature should not have the two last steps (password, analytics), do you concur @EricLarch ? 🤔

NastiaS:
Yes, it does seem a bit awkward to ask the same user about Password and Analytics again in the context of the same app. Some steps of the onboarding are not relevant to the device. I am looking into filtering some steps for this case.

PR #1101, opened 2018-07-13
Fix/1055

NastiaS:
This PR adds the possibility to re-launch the onboarding from the Settings/Help page if anyone wants to repeat the onboarding with another device. 
For this case it lets the user to skip the Password and Analytics pages because they are not related to the device. It also distinguishes between a new device vs already initialized device.
 
### Type

Enhancement 
### Context

Github issue #1055 

Settings, Onboarding


meriadec:
wip commit :scream:

PR #1150, opened 2018-07-16
use explicit params to avoid pass-in useless get params

gre:


gre:
this is general cleanup. not mandatory but we probably need to do a more general cleanup in not propagating all the useless get params

gre:
not tested enough / other remaining part that need this cleanup / nothing really important right now, (so that's why the hold :P )

PR #1171, opened 2018-07-17
ETH/ETC: always scan 5 empty accounts ahead for better discoverability

gre:
fix discoverability of ETH/ETC address that was manually used from MEW in a wrong order (not respecting the top-down scanning)

this will allow more eth accounts to be discovered. if users have left "empty pages" of address, we won't support this, you will have to move your funds from MEW to a new address.

proof:

capture d ecran 2018-07-17 a 15 31 37

before this commit, account 5 was discovered, but not account 10.

### Type

polish

### Context

people still complaining about not seeing their account. 
typically because this was not respected:

out

eth scan account

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -1,5 +1,6 @@
 // @flow
 import { Observable } from 'rxjs'
+import logger from 'logger'
 import React from 'react'
 import FeesField from 'components/FeesField/EthereumKind'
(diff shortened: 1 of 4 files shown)


gre:
ok can't ship this yet. the `emptyCount` value need to be reset to 0 at each derivation. because we want for instance, for ETC, scan both eth and etc address and with this system on for both...

gre:
also need to review the logic again. i think when it's `== 0` by default, this actually unecessarily scan an extra account.

PR #1173, opened 2018-07-17
Fixes #1153

### LINKED ISSUE #1153: Sorting order is not updated after importing accounts

- tested on Ledger Live **1.0.2**
- Platform and version: **Mac 10.13.5**

#### Expected behavior

Account imported should automatically be sorted by the selected preference.

#### Actual behavior

Accounts are imported in the order they are found, and added at the bottom of the list. Changing views doesn't trigger a reorder. The only way to sort correctly the accounts is to manually select the "sort by" option.

- import accounts
- they'll be in the order of importation and not the order of balance

image

gre:
- trigger a reorder on Portfolio page access. 
- trigger reorder in step 4 of add accounts (before AND after)

also don't do a reorder if countervalues are not ready

### Type

polish

### Context

#1153

sorting of accounts

### DIFF
--- a/src/components/modals/AddAccounts/steps/04-step-finish.js
+++ b/src/components/modals/AddAccounts/steps/04-step-finish.js
@@ -6,6 +6,7 @@ import styled from 'styled-components'
 import TrackPage from 'analytics/TrackPage'
 import Box from 'components/base/Box'
 import Button from 'components/base/Button'
+import RefreshAccountsOrdering from 'components/RefreshAccountsOrdering'
 import IconCheckFull from 'icons/CheckFull'
 import { CurrencyCircleIcon } from '../../../base/CurrencyBadge'
 import type { StepProps } from '../index'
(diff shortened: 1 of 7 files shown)


src/components/modals/AddAccounts/steps/04-step-finish.js
```diff
@@ -30,6 +31,9 @@ const Text = styled(Box).attrs({
 function StepFinish({ currency, t, checkedAccountsIds }: StepProps) {
   return (
     <Box align="center" py={6}>
+      <RefreshAccountsOrdering onMount />
+      {/* the unmount case is useful to trigger a second refresh because most of the time countervalues were not yet loaded... */}
+      <RefreshAccountsOrdering onUnmount />
```
meriadec: what about having `onMount` and `onUnmount` on the same `RefreshAccountsOrdering`

gre: aha yeah 😅

Issue #1110, opened 2018-07-13
Use a BigInt or BigNumber implementation on JS side


gre:
this is will prepare the future and is required to implement features like #905 reliably. This is a general improvement of the app but will take some times to get it everywhere.

PR #1183, opened 2018-07-18
Migrate the app to BigNumber.js

gre:
this completely migrate the app to BigNumber.js for better number precision and preparing need to implement "MAX" feature.

this is to go with live-common: https://github.com/LedgerHQ/ledger-live-common/pull/53

it does not however, benefit from it yet as there are remaining future work:

- ethereum js bridge implementation: as our API returns numbers, we can't be more precise than JS Number. In future, the c++ libcore will take over.
- formatter of live-common still use `toLocaleString`. it's actually a bit tricky to reimplement this JS native feature, but we'll eventually have to do.

### Type

code quality, precision of numbers

### Context

#1110 

we need to intensively test the app and all features related to balance / sending funds. I've migrate things carefully but we might have regression, so we need to test.


gre:
@meriadec  ok , i guess we can do it now then :) i'm trying to figure out the formatter now. it's a bit more hairy but at least this code merged as it is should be a 1-1 in term of feature

gre:
formatter is on live-common side anyway so this should be good for this side

gre:
NB: the data will normally migrate, but you can't go back to previous version (moving from this new to a 1.0.3 with same data won't work)

Issue #1201, opened 2018-07-19
Add ability to copy balances to the clipboard


Urgo:
TLDR: Please add the ability to right click on balances (of each coin and all together) and copy them to the clipboard.

Longer Explanation: Ledger Live is a great step towards managing my portfolio at a glance, but still only tracks balances of accounts stored on my nano s.  This is fine and probably the right choice, but the feature I'd love to have is pretty simple.  It would be great if you could either right click on balances (or really any place where a coin amount is shown) and copy them to the clipboard so I can copy and paste how much btc, eth, etc I have into a 3rd party app (i.e. delta).  Right now once a day or so I have to manually type in the numbers. Copy and pasting would save a lot of time.

#### Part of the application

Ledger Live

juan-cortes:
I had some time to play with this for a while, making it selectable like @NastiaS did on https://github.com/LedgerHQ/ledger-live-desktop/pull/1215 doesn't work because the balances use the `FlipTicker` component which is in charge of the flip animations on price changes and whatnot. Selecting the text results in the selection of _all_ the text inside the flip ticker. Think of the `FlipTicker` like an analog odometer from a car, made of wheels containing 0-9, when you select, you are selecting the whole wheel.

### Only making it selectable, no good.
![ori](https://user-images.githubusercontent.com/4631227/44177961-73313500-a0f0-11e8-82d8-e4ca44eca388.gif)


### Approach 1
The first solution that is easily implemented (done if needed) would be to intercept the `oncopy` event and override the copied value with the formatted amount or whatever value deemed appropriate. I'm not totally convinced with it since the selection is visible and you might be copying only a part of the amount but still get the full amount.
![sol1](https://user-images.githubusercontent.com/4631227/44175034-43c8fb00-a0e5-11e8-8a89-07bf4e6b2d7a.gif)

### Approach 2
Another option I whipped together is extending the component and showing a copy icon on hover, that will remove the need to manually select the text and _maybe_ feel more natural. Clicking on the hover element replaces the clipboard's content with the value to the left. A bit hacky since we need to generate a temp element to select the text from and all that, but nothing too noisy. This would of course require someone with an eye for design to make it look nice, I don't mind working together with someone else if needed.
![sol2](https://user-images.githubusercontent.com/4631227/44177632-11240000-a0ef-11e8-8019-d1ae9e312ce5.gif)

@gre I know this is super low priority but I'm just warming up familiarising myself with the code, hope you guys don't mind

[...]

PR #1215, opened 2018-07-19
fixes #1210

### LINKED ISSUE #1210: Enable copy and paste for text/instructions across the app
There were some requests and remarks on having a possibility to copy and paste text/instructions across the app for translation purposes. People who are not English-speakers, while there is no their language support should be able to copy and paste instructions into translators, etc. 

Currently the text selection is not available 
#### Part of the application

The information-rich text blocks that have instructions/warnings/explanations

NastiaS:
It should give non-English speakers a quick way to translate info chunks by copying and pasting the text
### Type

Enhancement 
### Context

Github issue #1210 

- Onboarding screens with all instructions 

- Application settings 

- All modals, including warnings, confirmation modals, modals with additional info during onboarding

- Manager: both Dashboard and Genuine check 

- Exchanges 

All elements (like cards) that were clickable and had text inside retain their behavior. You might copy the text using shortcuts but it would still navigate as before.

### DIFF
--- a/src/components/Onboarding/index.js
+++ b/src/components/Onboarding/index.js
@@ -202,6 +202,7 @@ class Onboarding extends PureComponent<Props> {
 const Container = styled(Box).attrs({
   bg: 'white',
   p: 60,
+  selectable: true,
 })`
   position: fixed;
   top: 0;
--- a/src/components/base/Modal/ModalBody.js
+++ b/src/components/base/Modal/ModalBody.js
@@ -77,6 +77,7 @@ const Body = styled(Box).attrs({
   bg: p => p.theme.colors.white,
   relative: true,
   borderRadius: 1,
+  selectable: true,
 })`
   box-shadow: 0 10px 20px 0 rgba(0, 0, 0, 0.2);
 `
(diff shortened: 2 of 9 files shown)


src/components/Onboarding/index.js
```diff
@@ -209,6 +209,8 @@ const Container = styled(Box).attrs({
   right: 0;
   bottom: 0;
   z-index: 25;
+  user-select: text;
+  cursor: text;
```
meriadec: is there a reason to not use the `selectable` attribute here?

src/components/base/Modal/ModalBody.js
```diff
@@ -79,6 +79,8 @@ const Body = styled(Box).attrs({
   borderRadius: 1,
 })`
   box-shadow: 0 10px 20px 0 rgba(0, 0, 0, 0.2);
+  user-select: text;
+  cursor: text;
```
meriadec: same comment

src/components/Onboarding/index.js
NastiaS: Actually not really. I guess if we decide to not use Box and use something else in those wrappers usage of selectable would be a very small issue compare to not using Box at all😁
Will change it

gre:
does it address #1201 too btw ? or not yet

NastiaS:
No, @gre, it doesn’t resolve completely the 1201

PR #1216, opened 2018-07-19
Fixes #1186

### LINKED ISSUE #1186: Ripple bug: 'not exactly one'..
i've been reported this bug from one user that tried to send a big amount of XRP (e.g. 1k)

"works with 100 but not 1k" 

img_4304

might just be related to ripple-lib.

gre:
Fixes #1186

app was failing to send more than 1000 XRP because was formatting numbers as 1,000.000 to the ripple-lib / API during the transaction preparation.

### Type

bug

### Context

 #1186

i had to buy 1000 XRP to reproduce the problem 🤣


meriadec:
capture d ecran 2018-07-20 a 08 45 15

SFYL

PR #1224, opened 2018-07-20
Fixes #1057 - Detect "Cant open device" errors and display link to help center

### LINKED ISSUE #1057: Detect & document Linux USB connection issues
Some Linux distributions (all?) require adding custom udev rules to correctly communicate with the device.

If the rules are not present, the device can be detected but interacting with it (e.g: checking if an app is opened) will fail. We may be able to detect the problem, and at minimum display an understandable error, and better: provide a link to [the help page with full instructions](https://support.ledgerwallet.com/hc/en-us/articles/115005165269-Connection-issues-with-Windows-or-Linux).

meriadec:
### Type

ui polish

### Context

#1057

no we can define errors help urls in `config/urls.js` and DeviceInteraction will display a link to it. Also we display err description if exists.

err

### DIFF
--- a/src/commands/getAddress.js
+++ b/src/commands/getAddress.js
@@ -6,10 +6,7 @@ import { fromPromise } from 'rxjs/observable/fromPromise'
 import { withDevice } from 'helpers/deviceAccess'
 import getAddressForCurrency from 'helpers/getAddressForCurrency'
 
-import { createCustomErrorClass } from 'helpers/errors'
-
-const DeviceAppVerifyNotSupported = createCustomErrorClass('DeviceAppVerifyNotSupported')
-const UserRefusedAddress = createCustomErrorClass('UserRefusedAddress')
+import { DeviceAppVerifyNotSupported, UserRefusedAddress } from 'config/errors'
 
 type Input = {
   currencyId: string,
--- a/src/components/DeviceInteraction/components.js
+++ b/src/components/DeviceInteraction/components.js
@@ -2,8 +2,13 @@
 
 import React from 'react'
 import styled from 'styled-components'
+import { translate } from 'react-i18next'
+
+import type { T } from 'types/common'
 
 import { radii } from 'styles/theme'
+import { openURL } from 'helpers/linking'
+import { urls } from 'config/urls'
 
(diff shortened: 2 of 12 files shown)


gre:
need a pass with design. I wonder if the description should be rendered in a second row, instead of center-alignining things like this

meriadec:
2018-07-20_501x586

meriadec:
@gre something like that?

2018-07-20_433x349

src/commands/getAddress.js
```diff
@@ -39,6 +36,8 @@ const cmd: Command<Input, Result> = createCommand(
           if (e.statusCode === 0x6985) {
             throw new UserRefusedAddress()
           }
+        } else if (e && e.message && e.message.match(/cannot open device with path/)) {
+          throw new CantOpenDevice()
```
gre: this is not the correct place where this error should be thrown

please do it at the `deviceAccess` level so we benefit this from all over the places.

i think it's basically the failure case of `open()`

src/commands/getDeviceInfo.js
```diff
@@ -6,13 +6,21 @@ import { withDevice } from 'helpers/deviceAccess'
...
 
 type Input = {
   devicePath: string,
 }
 
 const cmd: Command<Input, DeviceInfo> = createCommand('getDeviceInfo', ({ devicePath }) =>
-  fromPromise(withDevice(devicePath)(transport => getDeviceInfo(transport))),
+  fromPromise(
+    withDevice(devicePath)(transport => getDeviceInfo(transport)).catch(e => {
+      if (e && e.message && e.message.match(/cannot open device with path/)) {
```
gre: which means you can also undo this.

meriadec:
@dasilvarosa no that's incorrect, the error you just described is when device is not detected as connected. this one is related to #1057.

meriadec:
I mean: disconnecting/reconnecting will not help you

meriadec:
@dasilvarosa here is the updated wording for error. is that OK? if yes, can you approve

```
  "CantOpenDevice": {
    "title": "Device not reachable",
    "description": "Device detected but connection failed. Please retry and get help if the problem persists."
  }
```

PR #1229, opened 2018-07-20
use new icons model from api

valpinkman:
Switch Icon retrieving for apps to new api

### Type

Improvement

Manager


gre:
@valpinkman what is the status of this?

valpinkman:
It alls depends on Alix and the deployment of the update API with the icons integrated directly in the model. We would retrieve a url directly from model, but I don't know why it's still not deployed

gre:
after asking. we'll come back to this later after the API refactoring.

PR #1230, opened 2018-07-20
anonymise one more edge case

gre:
replace the app path by '.' , replace the user data by '$USER_DATA', even when it's part of an "URI" like in a few rare error cases

### DIFF
--- a/src/helpers/anonymizer.js
+++ b/src/helpers/anonymizer.js
@@ -11,9 +11,25 @@ const configDir = (() => {
 
 const cwd = typeof process === 'object' ? process.cwd() || '.' : '__NOTHING_TO_REPLACE__'
 
+// all the paths the app will use. we replace them to anonymize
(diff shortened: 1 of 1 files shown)


src/helpers/anonymizer.js
```diff
@@ -11,9 +11,25 @@ const configDir = (() => {
...
+// all the paths the app will use. we replace them to anonymize
+const basePaths = {
+  $USER_DATA: configDir,
+  '.': cwd,
+}
+
 function filepathReplace(path: string) {
-  if (!cwd) return path
-  return path.replace(cwd, '.').replace(configDir, '$USER_DATA')
+  if (!path) return path
```
valpinkman: if no path return path ?

gre: yeah. that was what was supposed to be the previous code. there is no point replacing an empty string.

PR #1250, opened 2018-07-23
add account error step in modal

NastiaS:
The add account modal had inconsistent error message display (red color) there were inconsistencies in Add Button display
Also there were no Customer Support button. It is added now, it opens a form to contact the support.

addaccountnew
### Type

Bug fix, Ui polish

Add account modal/error case


src/components/modals/AddAccounts/index.js
```diff
@@ -230,6 +236,7 @@ class AddAccounts extends PureComponent<Props, State> {
       onClickAdd: this.handleClickAdd,
       onCloseModal: this.handleCloseModal,
       setScanStatus: this.handleSetScanStatus,
+      customerSupport: this.handleCustomerSupport,
```
gre: I don't think this need to be contextual prop passed-in

src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -353,25 +366,34 @@ export const StepImportFooter = ({
...
-        </Button>
+        <Fragment>
+          <Button
+            mr={2}
+            outlineGrey
+            onClick={() => customerSupport()}
+            style={{ borderRadius: '4px' }}
+          >
+            {t('app:common.support')}
+          </Button>
```
gre: actually, I think this should be made generic to be a `ContactButton` that would just propagate props to Button but the onClick and children would be implemented

```
          <ContactButton
            mr={2}
            outlineGrey
            style={{ borderRadius: '4px' }}
          />
```

this would be beneficial not only to sharing more code but also the day we'll refactor this to trigger a modal

static/i18n/en/app.json
```diff
@@ -5,6 +5,7 @@
     "apply": "Apply",
     "confirm": "Confirm",
     "cancel": "Cancel",
+    "support": "Contact Support",
```
gre: to be confirmed with @dasilvarosa  but I think it's a "Contact us" wording today . at least it's the case in receive.confirmAddress.support so i think we probably want to share this (which would be out of the box if we could have a reused "ContactButton")

dasilvarosa: Yep, either:
- Contact us --> support.ledgerwallet.com/hc/requests/new
- Get support --> support.ledgerwallet.com

src/components/modals/AddAccounts/steps/03-step-import.js
NastiaS: 👍

src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -353,25 +367,27 @@ export const StepImportFooter = ({
     <Fragment>
       {currency && <CurrencyBadge mr="auto" currency={currency} />}
       {scanStatus === 'error' && (
-        <Button mr={2} onClick={() => setScanStatus('scanning')}>
-          {t('app:common.retry')}
-        </Button>
+        <Fragment>
+          <ContactButton mr={2} label={t('app:common.support')} url={urls.faq} />
+          <Button primary onClick={() => setScanStatus('scanning')}>
```
NastiaS: maybe we should do the same to the Retry button

src/components/base/RetryButton.js
```diff
@@ -0,0 +1,27 @@
...
+  t: T,
+  onClick: Function,
+  event?: string,
+}
+
+class RetryButton extends PureComponent<Props> {
+  render() {
+    const { t, onClick, event, ...props } = this.props
+    return (
+      <Button onClick={onClick} {...props} event={event}>
```
gre: both event and onClick can be left as part of `props` and no need to define neither in Props type

src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -353,25 +368,25 @@ export const StepImportFooter = ({
     <Fragment>
       {currency && <CurrencyBadge mr="auto" currency={currency} />}
       {scanStatus === 'error' && (
-        <Button mr={2} onClick={() => setScanStatus('scanning')}>
-          {t('app:common.retry')}
-        </Button>
+        <Fragment>
+          <ExternalLinkButton mr={2} label={t('app:common.getSupport')} url={urls.faq} />
+          <RetryButton primary onClick={setScanStatus('scanning')} />
```
gre: it looks like you're accidentally calling setScanStatus instead of giving a function!

src/components/modals/AddAccounts/steps/04-step-finish.js
```diff
@@ -50,13 +50,10 @@ function StepFinish({ currency, t, checkedAccountsIds }: StepProps) {
...
-export const StepFinishFooter = ({ onCloseModal, onGoStep1, t }: StepProps) => (
+export const StepFinishFooter = ({ onGoStep1, t }: StepProps) => (
   <Fragment>
-    <Button mr={2} outline onClick={onGoStep1}>
+    <Button mr={2} primary onClick={onGoStep1}>
       {t('app:addAccounts.cta.addMore')}
     </Button>
-    <Button primary onClick={onCloseModal}>
-      {t('app:common.close')}
-    </Button>
```
gre: why is is gone?

static/i18n/en/app.json
```diff
@@ -5,6 +5,7 @@
     "apply": "Apply",
     "confirm": "Confirm",
     "cancel": "Cancel",
+    "getSupport": "Get Support",
```
gre: we want this third wording @dasilvarosa  ? Contact us / Contact Support / Get Support 🤔

src/components/base/ExternalLinkButton.js
```diff
@@ -0,0 +1,25 @@
...
+  url,
+  event,
+  ...props
+}: {
+  label: string,
+  url: string,
+  event?: string,
+}) {
+  return (
+    <Button onClick={() => openURL(url)} {...props} event={event}>
```
gre: same for event . can be propagated as part of props

src/components/modals/AddAccounts/steps/04-step-finish.js
gre: @bscholving ok with this?

NastiaS: @khalilbenihoud requested for it

meriadec: would be great if it was tracked/written somewhere, we have github we may use it

NastiaS: yes, because we already have a cross ot close it and you can close it by clicking outside, it is like 3rd time and it was the main focus - primary button

gre: ok . make sense 👌

Issue #1244, opened 2018-07-23
High memory usage


vidaru:
#### Ledger Live Version and Operating System

- tested on Ledger Live v1.0.3
- Platform and version: Arch Linux
#### Expected behavior
Normal memory usage

#### Actual behavior

On a 32GB system, 70% of the total memory was being consumed by ledger live.

#### Steps to reproduce the behavior
Leave ledger live running for an hour or so, system had been hibernating and reloading, had received one transaction and sent a transaction during that time. Ledger live had been running for just over an hour.

Here is a screenshot of my `top` output

![screenshot from 2018-07-23 15-47-56](https://user-images.githubusercontent.com/38734836/43084307-5420b112-8e90-11e8-8c6a-4fcf8b6454c9.png)

gre:
is it happening when you see "Synchronizing" or does it say "Synchronized" ?
it looks like the Internal process is working so it might be related to sync (libcore)

vidaru:
It often says 'Synchronized failed' on my system, however it does seem to synchronize successfully as prices and transactions update

gre:
if you "hover" the error, what does it says? "timeout error"? I think we have some sort of queue getting accumulate in some cases and i'll try to tackle this in #1108

vidaru:
"Oops, a time out occurred" 
![screenshot from 2018-07-23 16-16-59](https://user-images.githubusercontent.com/38734836/43085810-d9b4ab96-8e93-11e8-8c4b-59a01d521aa5.png)

gre:
BTW as a workaround (/second thing to test) is to kill the "Ledger Live Internal" process when it happens, if my intuition is correct it would fix the issue.

it's totally ok to kill it at any time, the app will always be able to recreate it and is resilient to it being killed. it's purpose is exactly to encapsulate complex work (device access, c++ backend) so we might want to ping it to check if it respond and kill it from the main thread.

vidaru:
The app still responds fine and I killed it using the standard window close button (x). I only discovered the memory usage because I couldn't open up any other app on my system as gnome was complaining about a lack of memory.

KhalilBellakrid:
I didn't know there were cases where libcore is hanging for ever, there was an issue related to BCH about a hanging synchro but I could not reproduce after trying with user's xpub.
It would be nice to have scenario to investigate !

gre:
next time it happens again, it would be useful if you could Export your logs (Ctrl+E) .

[...]

PR #1253, opened 2018-07-24
Fix LibcoreBridge#synchronize to be properly unsubscribed

gre:
simplification of code and also potentially address problems around memory leak / timeout : the `.timeout()` we do on a sync will "finish" the job to exec, however the previous code was potentially .toPromise() it forever without the unsubscribe mecanism to work.

Now it's not sure this actually fix the bug #1244 , but i think it's something along this path: if the sync actually blocks in the internal thread. the render will eventually reach a timeout, but things might not be terminated properly still. so that's why we would also need another "kill if not responding" workaround too.

### Type

improvment

### Context

#1244

sync of btc like accounts


gre:
@meriadec wdyt putting this in 1.0.4 ? i think it might help #1244 and it's a light refactoring to just allow to unsubscribe it through

meriadec:
sounds GTM :ok_hand:

PR #1283, opened 2018-07-25
More information on clear cache modal

dasilvarosa:
More user education on the clear cache modal.
### Type

Wording polish.

Settings > Help > Clear cache

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -399,7 +399,7 @@
     },
     "softResetModal": {
       "title": "Clear cache",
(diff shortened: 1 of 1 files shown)


gre:
> Your settings and accounts are not affected

technically accounts get trashed so you temporary won't see operations (especially if you have no network during that)

but i guess it's to say that it does not affect the coins?

dasilvarosa:
It means that your accounts stay present in your portfolio: you don't have to add them again, unlike the hard reset.

PR #1288, opened 2018-07-25
modify wording on add account modal

NastiaS:
Modify wording on the add account modal

| Old        | New
| ------------- |:-------------
|screen shot 2018-07-25 at 16 07 13|screen shot 2018-07-25 at 18 27 04

### Type

UI polish 

### Context

None

Add account modal (trying to create second account without funding first one)

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -196,7 +196,7 @@
     "successDescription_plural": "Your accounts have been created.",
     "createNewAccount": {
       "title": "Add a new account",
-      "noOperationOnLastAccount": "There are no transactions on your last created <1><0>{{accountName}}</0></1> account. You must first receive a transaction on that account before you can add a new one.",
+      "noOperationOnLastAccount": "There are no transactions on your last created <1><0>{{accountName}}</0></1> account. You must first receive crypto assets on that account before you can add a new one.",
       "noAccountToCreate": "No <1><0>{{currencyName}}</0></1> account was found to create"
     },
     "cta": {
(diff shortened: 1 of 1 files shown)


static/i18n/en/app.json
```diff
@@ -196,7 +196,7 @@
     "successDescription_plural": "Your accounts have been created.",
     "createNewAccount": {
       "title": "Add a new account",
-      "noOperationOnLastAccount": "No transactions found on your last new account <1><0>{{accountName}}</0></1>. You can add a new account after you've started transacting on that account.",
+      "noOperationOnLastAccount": "There are no transactions on your last created <1><0>{{accountName}}</0></1> account. You must fund it before creating another account.",
```
dasilvarosa: There are no transactions on your last created <1><0>{{accountName}}</0></1> account. You must first receive a transaction on that account before you can add a new one.

gre: i feel the expression `receive a transaction` is wrong, no? shouldn't it be more `cryptoassets` ?

you receive funds / cryptoassets via a transaction 🤔


especially the first sentence already say "transaction" so it would be clear to also say you need to receive funds

(or whatever alternative you have to say "funds")

gre:
i feel the expression `receive a transaction` is wrong, no? shouldn't it be more `cryptoassets` ?

you receive funds / cryptoassets via a transaction 🤔


especially the first sentence already say "transaction" so it would be clear to also say you need to receive funds

(or whatever alternative you have to say "funds")

dasilvarosa:
Go ahead with: _You must first receive crypto assets on that account before you can add a new one._

PR #1297, opened 2018-07-26
Rework persistence data layer

### LINKED ISSUE #1235: rework the Persistence
both for robustness (see #1184 ) and performance, we need to rework completely how data are persisted.

it needs to be async, atomic and resilient.

maybe it just one json file that is saved in background or we need a better database 🤔  what's sure if we need to rework this. we need first a pragmatic solution with small changes to what we have today. first challenge is to make everything async. if we can achieve it next week. would be great. 
also need to make sure we can migrate things from old data.
to my mind the simplest solution: we could literally merge all files to one & just refactor the way this file is saved, it would be one big object with keys being the store name and value behind the data, accounts field would be potentially encrypted like before. there would be one big save that happen reccurently & debounced. we just recurrently sync, but in one atomic time. maybe the save happen on another thread, or at least it happen asynchronously

meriadec:
Closes #1235

This commit introduce a pretty big change on how app data files are
managed. Instead of having multiple `accounts.json`, `settings.json`,
`user.json`, `countervalues.json`, it now store everything in
`app.json` (filled by renderer process) and `windowParams.json` (filled by main process). And it does it in an *async* way, so it should prevent some
frame loss.

Migration will be seamless, and will keep `accounts.json` encryption if set.
So no user action is involved.

This changes comes also with some simplification on password handling
(e.g: no more need to store password hash).

/!\ Disclaimer

During development, I ran about a weird issue (~ 3 or 4 times) when my
data was simply "erased" in the file, so back to onboarding, no more settings,
etc. I suspect race/write condition, something with write-file-atomic, but I
didn't managed to reproduce, everytime I tried. Anyway, I feel not 100% confident
with it, so if you guys can test on your side, with your data (with encryption
key or not) it will help a lot!


src/migrations/migrations.js
```diff
@@ -0,0 +1,53 @@
...
+      await db.setKey('app', 'settings.password', undefined)
+      const windowParams = await db.getKey('app', 'settings.window')
+      await db.setKey('app', 'settings.window', undefined)
+      await db.setNamespace('windowParams', windowParams)
+      await Promise.all(legacyKeys.map(key => fsUnlink(path.join(dbPath, `${key}.json`))))
+    },
+  },
+]
+
+async function getLegacyData(filePath) {
```
gre: i guess you could inline this function in the migration it's used, because it's really going to be used and run once

meriadec: true that. d'you think it's worth to even separate each migration to dedicated file?

PR #1304, opened 2018-07-27
Optimize a bit libcoreSignAndBroadcast and libcoreSyncAccount perf

gre:
no need to send the whole account object from renderer to internal thread if we only need a few params.

### Type

performance


src/bridge/LibcoreBridge.js
```diff
@@ -120,41 +120,48 @@ const LibcoreBridge: WalletBridge<Transaction> = {
...
-            patch.pendingOperations = [] // For now, we assume a change will clean the pendings.
-          }
-
-          return {
-            ...account,
-            ...patch,
+    libcoreSyncAccount
+      .send({
+        accountId: account.id,
+        freshAddressPath: account.freshAddressPath,
```
meriadec: why do we need `freshAddressPath` to sync?

gre: to extract out some info: segwit? split? etc..

PR #1307, opened 2018-07-27
Add a way to get app version and log it if there is an error

gre:
this is not displayed to the UI yet for now, but at least we'll see the app version in logs when an error happen in AddAccount/Send/Receive


src/components/modals/Receive/steps/03-step-confirm-address.js
```diff
@@ -12,6 +12,7 @@ import ExternalLinkButton from 'components/base/ExternalLinkButton'
 import RetryButton from 'components/base/RetryButton'
 import type { StepProps } from '../index'
 import TranslatedError from '../../../TranslatedError'
+import DebugAppInfosForCurrency from '../../../DebugAppInfosForCurrency'
```
meriadec: ../../../../../../../../../../../../../../../

:doge:

PR #1308, opened 2018-07-27
switching to use str for providers

amougel:
Needs : 
- https://github.com/LedgerHQ/ledger-app-store-api/pull/21 to be in prod
- modifications have to be made on the API to match providers name and support 1.2 devices

DO NOT MERGE


gre:
@amougel what is the status of this?

gre:
after asking. we'll come back to this later after the API refactoring.

PR #1325, opened 2018-07-30
Add more timeout on the sync to avoid "forever running" issues

gre:


src/helpers/promise.js
```diff
@@ -65,6 +66,17 @@ export function createCancelablePolling(
   return { unsubscribe, promise }
 }
 
+export const timeoutTagged = <T>(tag: string, delay: number, promise: Promise<T>): Promise<T> =>
+  new Promise((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      reject(new TimeoutTagged('timeout', { tag }))
+    }, delay)
+    promise.then(resolve, e => {
```
meriadec: don't we need to clear the timeout in case of success?

gre: @meriadec good point, it's not mandatory because reject would just get ignored but i can still improve it

PR #1332, opened 2018-07-30
fixes sent modal for ethereum/litecoin when no balance was set

valpinkman:
Fix some issues on the send modal, allow zero to be set in fees, also fixed the trigger on ethereum to continue to next step

### Type

Bug Fix

### Context

Bug/Regression found while using the app. Some of these issues are a side effect of switching to BigNumber

Send Modal

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -403,7 +403,10 @@ const EthereumBridge: WalletBridge<Transaction> = {
       ? Promise.resolve()
       : Promise.reject(new NotEnoughBalance()),
 
-  getTotalSpent: (a, t) => Promise.resolve(t.amount.plus(t.gasPrice.times(t.gasLimit))),
+  getTotalSpent: (a, t) =>
+    t.amount.isGreaterThan(0) && t.gasPrice.isGreaterThan(0) && t.gasLimit.isGreaterThan(0)
+      ? Promise.resolve(t.amount.plus(t.gasPrice.times(t.gasLimit)))
+      : Promise.resolve(BigNumber(0)),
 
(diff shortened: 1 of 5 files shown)


src/bridge/EthereumJSBridge.js
```diff
@@ -403,7 +403,10 @@ const EthereumBridge: WalletBridge<Transaction> = {
       ? Promise.resolve()
       : Promise.reject(new NotEnoughBalance()),
 
-  getTotalSpent: (a, t) => Promise.resolve(t.amount.plus(t.gasPrice.times(t.gasLimit))),
+  getTotalSpent: (a, t) =>
+    t.amount.isGreaterThan(0) && t.gasPrice.isGreaterThan(0) && t.gasLimit.isGreaterThan(0)
+      ? Promise.resolve(t.amount.plus(t.gasPrice.times(t.gasLimit)))
+      : Promise.resolve(BigNumber(0)),
```
gre: (note: this is to be consistent with Libcore implementation. the zero value is used to block the "NEXT" on the UI)

src/components/FeesField/BitcoinKind.js
```diff
@@ -85,10 +87,10 @@ class FeesField extends Component<Props & { fees?: Fees, error?: Error }, State>
     return { items, selectedItem }
   }
 
-  componentDidUpdate() {
+  componentDidUpdate({ fees: prevFees }: OwnProps) {
     const { feePerByte, fees, onChange } = this.props
     const { items, isFocused } = this.state
-    if (fees && feePerByte.isZero() && !isFocused) {
+    if (fees && fees !== prevFees && feePerByte.isZero() && !isFocused) {
```
gre: (note: we only want to force the "reset to default fees" if new fees was loaded and value was still zero. because zero is still allowed)

PR #1340, opened 2018-07-31
Use new USB device detection from ledgerjs

MortalKastor:
### Type

Performance improvement

### Context

LedgerHQ/ledgerjs#191

### Parts of the app affected

Device detection


src/helpers/deviceAccess.js
```diff
@@ -39,7 +40,7 @@ export const withDevice: WithDevice = devicePath => job => {
     busy = true
     refreshBusyUIState()
     try {
-      const t = await TransportNodeHid.open(devicePath).catch(mapError)
+      const t = await retry(() => TransportNodeHid.open(devicePath), { maxRetry: 1 }).catch(mapError).catch(mapError)
```
meriadec: is double catch intended?

MortalKastor: Of course, it's safer.

meriadec:
I think we'll need

```
apt-get install -y libudev-dev
```

on CI config (see [here](https://github.com/poanetwork/circleci-playground/blob/f8178278fe137db012eb67a6a18025c78f005385/.circleci/config.yml#L38-L41))

meriadec:
Tested on Linux and I had to put `maxRetry: 2` in `withDevice` to avoid displaying the error (and stop pinging)

PR #1342, opened 2018-07-31
Fixes #1337

### LINKED ISSUE #1337: Setting wrong expectations on confidentiality
Running 1.0.7 on Arch Linux.

During the onboarding, the user is told that if their password is compromised, an attacker could get hold of their account information. Some users are likely to get the impression that any private information on the file system is encrypted. Account information is, however, by default saved in clear-text in logs and Cache.

As long as this is the case, can we change the text on this in the onboarding to avoid misunderstandings?

gre:
remove log files when the app closes.

### Type

privacy

### Context

#1337

log files of the app and of libcore


src/main/app.js
```diff
@@ -153,7 +153,7 @@ async function createMainWindow() {
   return window
 }
 
-app.on('window-all-closed', () => {
+app.on('window-all-closed', async () => {
```
meriadec: :eyes:

src/helpers/log.js
```diff
@@ -0,0 +1,17 @@
...
+export const resolveLogsDirectory = () => {
+  const { LEDGER_LOGS_DIRECTORY } = process.env
+  if (LEDGER_LOGS_DIRECTORY) return LEDGER_LOGS_DIRECTORY
+  const electron = require('electron')
+  return path.resolve((electron.app || electron.remote.app).getPath('userData'), 'logs')
+}
+
+export const cleanUpBeforeClosingSync = () => {
+  rimraf.sync(resolveLogsDirectory(), { disableGlob: true })
+  rimraf.sync(path.resolve(resolveUserDataDirectory(), 'sqlite/*.log'))
```
meriadec: not a big deal but I'm curious about why we can't do it "async" and we run both rimraf in parallel

gre: we'll keep like this for now, not such a big deal because it's at the time you press the X (or CMD+Q) and we somehow want to block the "close" event

PR #1366, opened 2018-08-03
Fix a JS bug when you have no txs but have balance

### LINKED ISSUE #1362: ETH Accounts: Cannot match against 'undefined' or 'null'.

- tested on Ledger Live 1.1.0
- Platform and version: Mac 10.13.6

#### Expected behavior

When adding an ETH Account, it should find accounts that have only received ETH issued by smart contracts..

#### Actual behavior

Step 3:
- Synchronizing...
- Error: Cannot match against 'undefined' or 'null'.

gre:
+ Fixes #1362 
+ fix visibility of Ethereum wallets that only receive funds from contract

we have to do a patch release this morning

### Type

bug

### Context

#1362

ETH scanning if one of your account have contract-only funds, scan will crash

### DIFF
--- a/src/components/modals/AddAccounts/steps/03-step-import.js
+++ b/src/components/modals/AddAccounts/steps/03-step-import.js
@@ -10,6 +10,7 @@ import uniq from 'lodash/uniq'
 import { urls } from 'config/urls'
 import ExternalLinkButton from 'components/base/ExternalLinkButton'
 import RetryButton from 'components/base/RetryButton'
+import isAccountEmpty from 'helpers/isAccountEmpty'
 
 import { getBridgeForCurrency } from 'bridge'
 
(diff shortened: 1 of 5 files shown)


src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -10,6 +10,7 @@ import uniq from 'lodash/uniq'
 import { urls } from 'config/urls'
 import ExternalLinkButton from 'components/base/ExternalLinkButton'
 import RetryButton from 'components/base/RetryButton'
+import isAccountEmpty from 'helpers/isAccountEmpty'
```
meriadec: is the file pushed? Don't see it

gre: done

PR #1375, opened 2018-08-03
Improvements on persistence layer

meriadec:
We had a few reported cases of problem occuring after v1.1.0 migration, more precisely some cases where migration is ran but files has already be migrated (basically: a nonce failure).

To resolve the problem it adds a second migration that puts the nonce _inside_ the `app.json` file so we are sure they are consistent (should have been done in the first :joy: ).

It also improve robustness of first migration, because.. eh, especially on Windows we can have all exotic cases (eating EBUSY & sh..), and having resulting non-deleted/still present files. Migration should not be affected by this HORRIBLENESS anymore.


gre:
tested and working

PR #1393, opened 2018-08-07
eip55 validation for ethereum recipient addresses

MortalKastor:
### Type

Bug Fix for ETH recipient addresses not being correctly validated

Send flow

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -9,6 +9,7 @@ import throttle from 'lodash/throttle'
 import flatMap from 'lodash/flatMap'
 import uniqBy from 'lodash/uniqBy'
 import type { Account, Operation } from '@ledgerhq/live-common/lib/types'
+import eip55 from 'eip55'
 import { apiForCurrency } from 'api/Ethereum'
 import type { Tx } from 'api/Ethereum'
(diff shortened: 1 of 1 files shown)


gre:
Are we ok doing and have we studied the general support of that  e.g. among exchanges / other wallets / eth community? the risk is that potentially people will be stuck trying to send funds to address that is not eip formatted.
To mitigate the problem you could ignore (aka keep doing basic validation like before) if and only if there are either ONLY lower case or ONLY upper case.

gre:
https://www.reddit.com/r/ledgerwallet/comments/95hqh5/trying_to_exchange_ethereum_classic_on_shapeshift/?utm_source=reddit-android

gre:
I've checked and for instance Wirex give a lower case address so I would recommend testing this (does eip55 lib support this case, I see it's only suggested in the spec GitHub comments) or implement the ignore "all lower or all upper"

meriadec:
What about having an additional warning box or a "I confirm this address is valid" cta if we can't pass the eip55 check (if mixed lower/upper)? Can be useful to inform user of eventual typo (bad copy paste/bad qr code scan)

meriadec:
follow up https://github.com/LedgerHQ/ledger-live-desktop/pull/1400

PR #1400, opened 2018-08-08
Display warning instead of rejecting ETH address without checksum

### LINKED ISSUE #1397: ETH addresses without checksum should only display a warning instead of rejecting the transaction

- tested on Ledger Live 1.1.4
- Platform and version: All

#### Expected behavior

When sending ETH to an address without checksum, the Live should display a warning instead of rejecting the transaction. This because the checksum has been introduced late in the coin life cycle and many addresses don't have one.

3 cases:

- address with a correct checksum: no warning
- address without checksum: warning
- address with an incorrect checksum: transaction rejected

#### Actual behavior

The transaction is rejected.

Try to send ETH to an address without checksum.

meriadec:
Closes #1397

This commit introduce a new `warning` prop on Input component. `error`
prop will always takes precedence over it if present, btw.

We don't reject anymore recipient validity if address is either full
lowercase OR full uppercase, to support legacy formats: having a warning
will allow user to go to next step.

Obviously the fix is available for ETC as well.

Wording to be reviewed :)

eth

***edit: wording has been updated to `Auto-verification not available: carefully verify the address`***

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -16,10 +16,8 @@ import { getDerivations } from 'helpers/derivations'
 import getAddressCommand from 'commands/getAddress'
 import signTransactionCommand from 'commands/signTransaction'
 import { getAccountPlaceholderName, getNewAccountPlaceholderName } from 'helpers/accountName'
-import { NotEnoughBalance } from 'config/errors'
+import { NotEnoughBalance, ETHAddressNonEIP } from 'config/errors'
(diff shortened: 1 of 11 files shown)


meriadec:
@khalilbenihoud there were no warning color so I took `#f57f17` (from Google Material's palette). Feel free to discuss that.

meriadec:
squashed :ok_hand:

gre:
'legacy address' wording is a bit weird. Otherwise 👌

meriadec:
Yep, it has been updated ;)

Issue #1395, opened 2018-08-07
Replace qrcode-reader by jsQR


meriadec:
[qrcode-reader](https://github.com/edi9999/jsqrcode) is deprecated and not maintained anymore, it has to be removed and replaced by [jsQR](https://github.com/cozmo/jsQR) (or similar modern and **tested** library).

After that, we can re-enable the QR code scanning from the send flow.

PR #1404, opened 2018-08-08
Replace unmaintained QR reading lib

MortalKastor:
The unmaintained `qrcode-reader` has been replaced by `jsQR`

### Type

- Bugfix
- Code Quality Improvement

### Context

Github issue #1395

The QR code scanning in the send flow, after being re-enabled.


MortalKastor:
Should I also re-enable the QR reading feature @fdv ?

PR #1409, opened 2018-08-10
Added Auto-Lock functionality

juan-cortes:
Added  the auto-lock functionality that when having a password lock set in the settings of the app, will allow the user to set an idle timeout that will trigger the auto-lock of the app. This was implemented by adding a simple component `Idler` that will monitor the last action (mouseover and keypressed currently) and check the difference between said last action and the current time.

Events are debounced in order not to trigger the call needlessly.
Defaults to 10min and hides/disables when no password is set in the app.

aug-10-2018 20-54-16

### Type

Feature

### Context
None

Settings page, Display Section now has a new entry and accesses the `hasPassword` setting to check whether to render or not the setting. Also had to add the above mentioned `Idler` component to avoid adding functionality elsewhere.

As for testing, I don't mind writing a test case for this if needed. I tested it with logs and waiting, it's simple enough.


.gitignore
```diff
@@ -10,3 +10,12 @@
...
 /build/linux/arch/*.tar.xz
+.idea/workspace.xml
+.idea/vcs.xml
+.idea/modules.xml
+.idea/misc.xml
+.idea/ledger-live-desktop.iml
+.idea/inspectionProfiles/Project_Default.xml
+.DS_Store
+.gitignore
+yarn.lock
```
meriadec: Remove all those lines :)
- consider using global gitignore for editor/environment/os specific files
- .gitignore and yarn.lock shouldn't be ignored

src/components/Idler.js
```diff
@@ -0,0 +1,65 @@
...
+  lock
+}
+
+class Idler extends PureComponent<Props> {
+  constructor(){
+    super();
+    this.checkForAutoLock.bind(this)
+    this.idleTimeHandler.bind(this)
+  }
+
```
meriadec: No need to bind (arrow functions have parent `this`)

src/components/Idler.js
```diff
@@ -0,0 +1,65 @@
...
+  componentDidMount() {
+    window.addEventListener('keydown', this.debounceOnChange)
+    window.addEventListener('mouseover', this.debounceOnChange)
+    setInterval(this.checkForAutoLock, 10000)
```
meriadec: Remove the listeners & clear the interval in `componentWillUnmount` (even if never unmounted for now)

src/components/Idler.js
```diff
@@ -0,0 +1,65 @@
...
+    this.setState({lastAction: Date.now()})
+  }
+
+  render() {
+
+    return null
+  }
+}
+
+export default translate()(
```
meriadec: No need `translate()`

src/components/SettingsPage/PasswordAutoLockSelect.js
```diff
@@ -0,0 +1,64 @@
...
+    { value: "60", label: `1 ${this.props.t('app:time.hour')}` },
+    { value: "-1", label: this.props.t(`app:common.never`) }
+  ]
+
+  render() {
+    const { autoLockTimeout } = this.props
+    const currentTimeout = this.timeouts.find(l => l.value === autoLockTimeout)
+
+    return (
+      <Fragment>
```
meriadec: No need `Fragment`

meriadec:
Looks good! :+1: 
Can you rebase your commits into one, to keep atomicity ? Thanks!

@gre @nicolasbigot WDYT of the functionality? Do we put it inside v1.2.0 ?

juan-cortes:
@meriadec done, I think

gre:
neat

src/components/Idler.js
```diff
@@ -0,0 +1,73 @@
...
+
+  componentWillUnmount() {
+    window.removeEventListener('keydown', this.debounceOnChange)
+    window.removeEventListener('mouseover', this.debounceOnChange)
+    clearInterval(this.interval)
+  }
+
+  interval: IntervalID
+
+  debounceOnChange = debounce(_ => this.idleTimeHandler(), 1000)
```
gre: this debounce is not "free" at unmount time. not sure if we have such unmount case, but for the beauty of it, you could `this.debounceOnChange.cancel()` at unmount to be sure the inner function won't get called after.

src/components/Idler.js
```diff
@@ -0,0 +1,73 @@
...
+    if (this.props.hasPassword && timeout && +timeout !== -1) {
+      if (Date.now() - (this.state.lastAction + timeout * 60000) > 0) {
+        this.props.lock()
+      }
+    }
+  }
+
+  idleTimeHandler = _ => {
+    this.setState(_ => ({
+      lastAction: Date.now(),
```
gre: technically it wouldn't need to be a state. it could literally be a property living on the comopnent instance. but i think it's ok as this render() don't do much anyway

src/components/SettingsPage/PasswordAutoLockSelect.js
```diff
@@ -0,0 +1,63 @@
...
+  handleChangeTimeout = ({ value: timeoutKey }: *) => {
+    this.props.setAutoLockTimeout(+timeoutKey)
+  }
+
+  timeouts = [
+    { value: '1', label: `1 ${this.props.t('app:time.minute')}` },
+    { value: '10', label: `10 ${this.props.t('app:time.minute')}s` },
+    { value: '30', label: `30 ${this.props.t('app:time.minute')}s` },
+    { value: '60', label: `1 ${this.props.t('app:time.hour')}` },
+    { value: '-1', label: this.props.t(`app:common.never`) },
```
gre: kinda minor feedback here, but it should be possible to make `value` a `number` , that way you don't need to `+()` cast it. it might just be about providing a render item function for `<Select>`

juan-cortes:
@gre followed your changes, current pr should address all above

src/components/Idler.js
juan-cortes: Added it to the current pr

Arnaud97234:
Tested on MacOS and It works !

dasilvarosa:
Auto-lock
Automatically lock Ledger Live after inactivity

PR #1416, opened 2018-08-13
Put back QR code address scanning and enforce scan security

### LINKED ISSUE #1406: Enforce the checksum on the QR code decoding
When using the QR code, we should refuse any address not having a checksum instead of just displaying a warning. 

The message should be "This address doesn't have automated checking and therefore can't be used with the QR code scanner" or something like this. Lea + support to display the warning.

### LINKED ISSUE #1412: Bring the QR Code back
But only after #1406 is done and well tested.

meriadec:
This commit prevent scanning non-eip55 addresses (without checksum),
or more generally prevent scanning address that generate warning for
recipient (currently the only possible warning is for non-eip55 address
for ETH/ETC).

Here is what appear on this case:

2018-08-13_458x90

If the warning-address is entered with keyboard instead of scanning, it
will still display the warning (as now), the behaviour doesnt change.

Some test material:

| eip55           | non-eip55  |
| :-------------: | :-------------: |
| `0x94381955F4028159A477a107510618aDb6B79Eb7` | `0xdecd15c7c00d2db819ded845b5c4d4bd81cb7d66` |
| eip      |  non-eip |

Closes #1412
Closes #1406

### DIFF
--- a/src/components/QRCodeCameraPickerCanvas.js
+++ b/src/components/QRCodeCameraPickerCanvas.js
@@ -156,7 +156,7 @@ export default class QRCodeCameraPickerCanvas extends Component<
         .catch(e => {
           if (this.unmounted) return
           this.setState({
-            message: String(e.message || e),
+            message: String(e.message || e.name || e),
           })
         })
     }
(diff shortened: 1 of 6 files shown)


MortalKastor:
Built production version on my Mac and tested EIP55, non-EIP55 and invalid ETH address 👉🆗

PR #1423, opened 2018-08-15
Avoid stack overflow when calling execute method of NJSExecutionContext

KhalilBellakrid:
This will avoid an 'Internal Process Error (null)' caused by a stack overflow (Maximum call stack size exceeded on JS side) due to recursive calls to execute method of NJSExecutionContext

### Type

Bug Fix

### Context

Many users are facing unexplained libcore errors of type 'Internal Process Error (null)' when computing fees, this occurs when users try to send an important amount which requires picking a lot of UTXOs (more than 150).

All runnables from libcore side are now called through a 'setImmediate' method

### DIFF
--- a/src/helpers/init-libcore.js
+++ b/src/helpers/init-libcore.js
@@ -24,7 +24,8 @@ const stringToBytesArray = str => Array.from(Buffer.from(str))
 const NJSExecutionContextImpl = {
   execute: runnable => {
     try {
-      runnable.run()
+      const runFunction = () => runnable.run()
+      setImmediate(runFunction)
     } catch (e) {
       logger.log(e)
     }


gre:
won't that substantially increase the exec time or it is very negligible? it's weird we reach JS max call which is 10k in Chrome I think.

KhalilBellakrid:
Actually, I started by writing a test in C++ to reproduce the bug but it was always succeeding that's how I realised it was specific to JS or in this case to nodeJS binding. 
I used 'setImmediate' (instead of a setTimeout(,0)) to reduce the loss of performance, so I think it will be negligible but we could benchmark that once we have a tool to do so.
And I agree this was also surprising to me, but if I remember we are running most of our calls to libcore in an async promise-based calls which are requiring promise contexts (which are not negligible) each time, if we think about that in a recursive use case it can be a mess (specially in a child process).
Another possible cause would be the fact that we are dealing with a native addon, so perhaps the bottleneck is from v8 side, I know for sure that node has some options to tweak the size of stack v8 is allowed to use (and by default it's less than 1MB).
Those 2 theories can be verified by a profiling tool that watches those indicators.
I'm no JS expert, but to me those are the only possible explanations for this issue.

Issue #1448, opened 2018-08-23
isValidTransaction never called on 01-step-amount


juan-cortes:
#### Ledger Live Version and Operating System

- tested on Ledger Live 1.1.7
- Platform and version: Mac 10.14 Beta (18A365a)

#### Expected behavior

A bridge implementation should be able to define additional validation rules that should prevent the form from being submitted. For example, while implementing the Memo system from Stellar, I need to be able to check those fields too.

#### Actual behavior
`isValidTransaction` is never called so we can get to step 2 with potentially invalid transactions

#### Steps to reproduce the behavior
If needed I'll reproduce this again in the main branch but hopefully the screenshot of the current integration suffices to validate/dismiss this bug report. **I can make a PR to solve this no problem**, it's simply adding the call to `bridge.isValidTransaction` and including it in resolving `canNext`

**Update** Checked out master and tested this on the Ripple bridge, with the same result, a zero amount transaction that would break according to the code, let's you through, I proceed to make the PR

**Update 2** Decided we should drop `isValidTransaction` in favor of `checkCanBeSpent` which will be renamed to `checkValidTransaction` to be a bit more generic. PR updated

gre:
yeah I think this is currently a dead code. 
`checkCanBeSpent` is the actual call being used instead of isValidTransaction that was too limited. we probably should drop isValidTransaction

juan-cortes:
Updated PR @gre

PR #1449, opened 2018-08-23
isValidTransaction dropped in favour of checkCanBeSpent for custom validations

juan-cortes:
Simple fix for https://github.com/LedgerHQ/ledger-live-desktop/issues/1448 Maybe I misunderstood the function of `isValidTransaction` but since it wasn't being called I figured it's probably a bug.

### DIFF
--- a/src/bridge/types.js
+++ b/src/bridge/types.js
@@ -76,15 +76,16 @@ export interface WalletBridge<Transaction> {
 
   getTransactionRecipient(account: Account, transaction: Transaction): string;
 
-  isValidTransaction(account: Account, transaction: Transaction): boolean;
-
   // render the whole Fees section of the form
   EditFees?: *; // React$ComponentType<EditProps<Transaction>>;
 
   // render the whole advanced part of the form
   EditAdvancedOptions?: *; // React$ComponentType<EditProps<Transaction>>;
 
-  checkCanBeSpent(account: Account, transaction: Transaction): Promise<void>;
+  // validate the transaction and all currency specific validations here, we can return false
+  // to disable the button without throwing an error if we are handling the error on a different
+  // input or throw an error that will highlight the issue on the amount field
+  checkValidTransaction(account: Account, transaction: Transaction): Promise<boolean>;
 
   getTotalSpent(account: Account, transaction: Transaction): Promise<BigNumber>;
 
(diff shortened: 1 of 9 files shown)


gre:
mmh I think we should probably be able to just use `checkCanBeSpent` instead (and probably provide a PR that drops `isValidTransaction` that was way too limited. the fact it's sync and also just return a boolean. checkCanBeSpent is its "async" version that also throw custom errors – that can be translated)

juan-cortes:
After our talk @gre I've changed the PR to drop `isValidTransaction` and renamed `checkCanBeSpent` to a more generic `checkValidTransaction`.

src/bridge/types.js
```diff
@@ -76,15 +76,16 @@ export interface WalletBridge<Transaction> {
...
   // render the whole Fees section of the form
   EditFees?: *; // React$ComponentType<EditProps<Transaction>>;
 
   // render the whole advanced part of the form
   EditAdvancedOptions?: *; // React$ComponentType<EditProps<Transaction>>;
 
-  checkCanBeSpent(account: Account, transaction: Transaction): Promise<void>;
+  // validate the transaction and all currency specific validations here, we can return false
+  // to disable the button without throwing an error if we are handling the error on a different
+  // input or throw an error that will highlight the issue on the amount field
```
gre: Please detail why we need the case of returning "false"? I think we always want to give feedback to user with an error (in failing promise) to inform why it's blocking.
and if we don't, we probably just should ignore a given error to not display it.

NB: i'm currently in progress of refactoring all of these functions in our mobile implementation in order to unify things a bit better between the 2 projects for the future, so we might want to hodl a bit on this.

juan-cortes: If I throw an exception on the `checkValidTransaction` method, the field that gets the feedback and message is the amount field. I wanted to show the error on a different field but still disable the _next_ button if the validation failed.

  ```
    checkValidTransaction: async (a, t) => {
      if (t.memoError) {
        return false
      }
      if (t.amount.isLessThanOrEqualTo(a.balance.plus(t.fee))) {
        return true
      }
      throw new NotEnoughBalance()
    }
```
I'm not against changing this at all, I just needed a way to throw an error on my own input from the `AdvancedOptions` without it being linked to the amount field. So with the code in this pr, the validation of an invalid form, failing because of a memo issue, would give us this

image


If instead of relying on the false case like I'm doing I throw an error if `t.memoError` I get this sort of validation which is not really true because I have no issue with the value of amount but rather the value of the memo. Might be a better way that I'm failing to see?

image

meriadec:
@gre I think we can merge this for now? If needed we can iterate later for error messages precision.

gre:
we can yes. we need to reiterate around this part for next release so i'll do more testing on these subject anyway. (in context of improving how fees are loaded)

PR #1450, opened 2018-08-24
Misc build scripts improvements

MortalKastor:
Some various build scripts polish/improvements:
- Prevent a race condition issue when drafting a release on GitHub
- Fix OS and version not shown on Mac at start
- Have a meaningful error for when Museo is missing and we're not able to get it (eg on Windows)

### Type

Build system improvement

Build scripts

### DIFF
--- a/scripts/create-draft-release.js
+++ b/scripts/create-draft-release.js
@@ -1,0 +1,85 @@
+#!/usr/bin/env node
+
+/* eslint-disable no-console */
+
+const util = require('util')
+const exec = util.promisify(require('child_process').exec)
+const octokit = require('@octokit/rest')()
+
+const repo = {
+  owner: 'LedgerHQ',
+  repo: 'ledger-live-desktop',
+}
+
+async function getTag() {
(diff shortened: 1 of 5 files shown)


scripts/create-draft-release.js
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env node
+
+/* eslint-disable no-console */
+
+const util = require('util')
```
meriadec: is `util` a peerDependency? Because I don't see it in package.json neither in `src/` directory. Maybe use `promisify` from our `helpers/promise` (would require to register Babel, so not 100% sure it's worth)

MortalKastor: `util` is [part of Node.js](https://nodejs.org/dist/latest-v8.x/docs/api/util.html)... but your comment made me realise I forgot to add the `@octokit/rest` dev dependency to this PR 😅

meriadec: oh nice :D
note to myself: replace our promisify by util.promisify

Issue #1444, opened 2018-08-22
XRP qrcode scanning should also support the 'dt' tag id parameter


gre:
https://wiki.ripple.com/Ripple_URIs

`dt` is part of the uri scheme of XRP. so if we scan it, we should also prefill the tag id. i would expect that when this happen, we also force open the advanced options spoiler. or maybe it shouldn't be a spoiler at all https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/components/AdvancedOptions/RippleKind.js#L17 regarding how important the tag id is for ripple.

for instance, as in the example: `rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn?dt=123` should be accepted. For instance this is actually what you see when using ShapeShift. I would expect copy pasting this whole thing in the address field would write `rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn` in address and `123` in tagId field.

PR #1454, opened 2018-08-25
Auto tag/amount from qrcode/pasting on ripple addresses

juan-cortes:
Solves https://github.com/LedgerHQ/ledger-live-desktop/issues/1444 which involves removing the wrapper spoiler for the ripple tag to make it more prominent and parsing the address for a valid tag following the pattern described in the ripple urls https://wiki.ripple.com/Ripple_URIs

### Type
enhancement

### Context
 https://github.com/LedgerHQ/ledger-live-desktop/issues/1444
aug-25-2018 20-01-35

Since the QRCode part simply alters the input value and this still triggers the onchange handler there was no need to do anything on that part. This affects only ripple so the implementation was encapsulated inside the bridge attempting to extract the params from the url string so while I was there also dragged the amount from the url if present.

Haven't added it to the gif to not make it too long but it's there.


src/bridge/RippleJSBridge.js
```diff
@@ -463,10 +463,29 @@ const RippleJSBridge: WalletBridge<Transaction> = {
 
   getTransactionAmount: (a, t) => t.amount,
 
-  editTransactionRecipient: (account, t, recipient) => ({
-    ...t,
-    recipient,
-  }),
+  editTransactionRecipient: (account, t, recipient) => {
+    const params = new URLSearchParams(recipient)
+    recipient = recipient.split('&')[0]
```
gre: my fault in the #1444 but it's actually a split on `?`

e.g.

`rwfGzgd4bUStS9gA5xUhCmg1J86TMtmGMo?dt=51489346`

src/bridge/RippleJSBridge.js
```diff
@@ -463,10 +463,29 @@ const RippleJSBridge: WalletBridge<Transaction> = {
...
+
+    // Extract parameters we may need
+    for (const [key, value] of params.entries()) {
+      switch (key) {
+        case 'dt':
+          t.tag = +value
```
gre: I think it would be preferable to use `parseInt(value, 10)` and check for potential NaN too

src/bridge/RippleJSBridge.js
juan-cortes: Oh right, that makes more sense I blindly tested using your string, I'll update the PR with those

Issue #936, opened 2018-07-09
Support for Stellar (XLM) accounts


nunoarruda:
- [x] I have checked this feature was not yet requested.

#### Part of the application

Accounts

#### Description

There's already a Stellar app for the Ledger Nano S, now it would be handy if Ledger Live supported Stellar accounts:
<img width="748" alt="screen shot 2018-07-09 at 11 19 17" src="https://user-images.githubusercontent.com/1189149/42447604-fe1ace0a-8369-11e8-8484-3be8b5add17d.png">

chrisplusplus:
I will be contributing to the development of adding Stellar to account management.  I could really use some stellar sneks to lend a hand!

kingofsevens:
Where can I follow your work on it @chrisplusplus?

sphinxc0re:
@chrisplusplus I would also like to see stellar integration in ledger live

juan-cortes:
you guys know about my pr? there's ready a working o yegtatiom done, just wait a couple weeks for the guys to have time to get it merged

gre:
Thanks for the issue.
We're unifying integration requests into [#1560](https://github.com/LedgerHQ/ledger-live-desktop/issues/1560).

Please follow [#1560](https://github.com/LedgerHQ/ledger-live-desktop/issues/1560) for updates on what will be the process.

atreyyu:
I've just upgraded to the latest version firmware (1.4.2) and latest stellar app in ledger live. Since then, log in to stellar.org/account-viewer is not working anymore. The ledger option is greyed out.
The tutorial says that "browser support" should be on, however since the firmware upgrade there is no "browser support" anymore. Instead you can select hash signing  yes or no (no idea what that means, anyway, is not working with neither of the 2 choices).
After googling this issue I see that there is a lot of talk abou that on reddit and no solution (since this summer !?).
Any news/solutions on this thing ?

PR #1458, opened 2018-08-26
Stellar Integration

juan-cortes:
### Screenshots
image
image
image
image

### List of things that are working
  - Detecting accounts from the device with funds, through derivative path
  - Adding accounts, renaming, deleting, etc
  - Live updates and pretty charts
  - Sending a payment to an account with or without a MEMO
  - Working with a locally created account derived from our seed that is not created yet
  - Detect that we are sending XLM to a non-existing account (but valid) in order to change
    the operation from a payment to a create_account one. With minimum startingBalance check
  - Synchronisation of accounts which updates balances and operations
  - UI for memos, Memos, and memo types, with validations
  - Some generic adaptations, mainly wrapping in `Ellipsis` when the address was overflowing

  - Displaying the memo of an existing transaction, maybe on the detail view until we get tags on ledger live
  - Filtering out transactions that are lower than X amount, due to spam, the same way account-viewer does
  - You tell me

### How to set-up the environment to test this
Simply clone the repo and run `yarn install` and `yarn start`

### Tests
~~Although it will locally pass the CI it will fail online since it won't find anything related to Stellar on  [LedgerHQ/ledger-live-common](https://github.com/LedgerHQ/ledger-live-common) and it relies on that to function. Once (if) the PR for that repo is accepted then this could start to move forward.~~
With the latest version of live-common now including Stellar and the dependencies that relied on it using that version, we can now pass all tests successfully.

### Using the Testnet
If you don't have lumens you can use the testnet by setting the `STELLAR_USE_TESTNET` flag in your `.env` file and topping up some test coins at the lab using [the Friendbot](https://www.stellar.org/laboratory/#account-creator?network=test)

### DIFF
--- a/src/bridge/StellarJSBridge.js
+++ b/src/bridge/StellarJSBridge.js
@@ -1,0 +1,349 @@
+// @flow
+import { BigNumber } from 'bignumber.js'
+import { Observable } from 'rxjs'
+import React from 'react'
+import type { Account, Operation } from '@ledgerhq/live-common/lib/types'
+import { getDerivations } from 'helpers/derivations'
+import getAddress from 'commands/getAddress'
+import last from 'lodash/last'
+import signTransaction from 'commands/signTransaction'
+import { parseAPIValue, getServer, formatAPICurrencyXLM } from 'api/Stellar'
+import { Transaction as StrTransaction, Asset } from 'stellar-sdk'
+import { StrKey } from 'stellar-base'
+import AdvancedOptionsStellarKind from 'components/AdvancedOptions/StellarKind'
+import { getAccountPlaceholderName } from 'helpers/accountName'
+import { NotEnoughBalance } from 'config/errors'
+import type { WalletBridge, EditProps } from './types'
+
+type Transaction = {
+  amount: BigNumber,
+  recipient: string,
+  fee: BigNumber,
+  memoError?: ?string,
+  memoType?: { label: string, value: ?string },
+  memo?: ?string,
+}
+
+const paymentsToOperation = async (id, payments, currency, account) => {
+  const parsedPayments = []
+  for (const payment of payments) {
+    let from
+    let to
+    let value
+
+    if (payment.type === 'create_account') {
+      from = payment.funder
+      to = payment.account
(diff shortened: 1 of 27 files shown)


sphinxc0re:
@juan-cortes This looks like really good work. Thank you!

defunctzombie:
Related feature request: https://github.com/LedgerHQ/ledger-live-desktop/issues/936

defunctzombie:
@juan-cortes I believe the dependencies (countervalue and common) have both been released with support for stellar.

juan-cortes:
@defunctzombie I didn't have time to check it worked out of the box after rebasing the current `develop` version, today I did and it made things a lot easier for anyone trying to test it since it no longer requires extraneous forked dependencies. Thanks for pointing it out, I didn't even notice my pr was merged for common, or if I did I forgot!

defunctzombie:
@sphinxc0re What is this PR lacking before it can be merged?

juan-cortes:
@gre @meriadec let me know if anything else needs changing in order to be merged, I know you guys are busy with the mobile app so no rush from my end, but if anything needs more work I'm ready to deal with it.

src/bridge/StellarJSBridge.js
```diff
@@ -0,0 +1,349 @@
...
+        memoType: (t.memoType && t.memoType.value) || 'MEMO_NONE',
+        amount,
+      },
+    })
+    .toPromise()
+  const signedTransaction = new StrTransaction(signedTransactionXDR)
+
+  if (!isCancelled()) {
+    onSigned()
+    const transactionResult = await server.submitTransaction(signedTransaction)
```
bartekn: Please handle timeout errors: https://www.stellar.org/developers/horizon/reference/endpoints/transactions-create.html#timeout Otherwise users may think that tx fail, but it actually succeeded.

In case of a timeout error you should inform user about it and let them try to resubmit the same transaction again (with the same seqnum) to check it's final status.

defunctzombie:
@gre What would allow this work to continue?

gre:
we'll keep this PR as reference and backport the code in live-common (potentially it will be solved on libcore side, depending who do it first^^)

PR #1464, opened 2018-08-27
Bump electron to 1.8.8

meriadec:
Following [WebPreferences electron fix](https://electronjs.org/blog/web-preferences-fix), let's bump electron version to 1.8.8. nb: it was not affecting the app, but it removes the github annoying header :)


meriadec:
(tested on Linux)

PR #1474, opened 2018-08-30
drop the libcore db on clear cache and hard reset

gre:
hard reset and 'clear cache' will drop the libcore db that we can fully restore

### Type

improvement

### Context

having to ask users to manually sqlite shouldn't be necessary after this because a clear cache will do it.

clear cache and hard reset


src/components/SettingsPage/CleanButton.js
```diff
@@ -33,11 +31,15 @@ class CleanButton extends PureComponent<Props, State> {
 
   close = () => this.setState({ opened: false })
 
+  actionTriggered = false
   action = async () => {
-    this.props.cleanAccountsCache()
-    await delay(500)
-    db.cleanCache()
-    remote.getCurrentWindow().webContents.reload()
+    if (this.actionTriggered) return
```
meriadec: I think it would be better to put `isLoading` on corresponding `Button` (which disable it + display a spinner) like it's [done in ConfirmModal](https://github.com/LedgerHQ/ledger-live-desktop/blob/37ee4525915fb0926a1826b6125e498bfdf1a9d5/src/components/base/Modal/ConfirmModal.js#L81) because it provide feedback to user.

Anyway, I understand we have to deliver fast this release so it can wait but those kind of micro interactions are :heart:

src/helpers/reset.js
```diff
@@ -0,0 +1,33 @@
...
+  console.log(`Deleting ${dbpath}`)
+  rimraf.sync(dbpath, { glob: false })
+}
+
+export async function hardReset() {
+  resetLibcoreDatabase()
+  disableDBMiddleware()
+  db.resetAll()
+  await delay(500)
+  window.location.href = ''
```
meriadec: I know it's legacy moving code, but let's have a single way to reload the page :D
(softReset use electron currentWindow, so we can use either the one or the other)

PR #1486, opened 2018-09-05
Prevent calling setState if unmounted on CurrentAddress

meriadec:
this removes annoying warning


src/components/CurrentAddress/index.js
```diff
@@ -143,6 +143,10 @@ class CurrentAddress extends PureComponent<Props, { copyFeedback: boolean }> {
     copyFeedback: false,
   }
 
+  componentWillUnmount() {
+    this._isUnmounted = true
+  }
```
gre: can we do a clearTimeout instead (and putting the inner closure in a method?)

meriadec: yep! better. thx

PR #1497, opened 2018-09-07
Drastically increment sync timeouts to address long sync issues

meriadec:
Those values had been tested against user xpub that was reported failing to sync (reproduced), and this fixes the problem.

More a quick win than an optimal solution.

### DIFF
--- a/src/helpers/libcore.js
+++ b/src/helpers/libcore.js
@@ -555,7 +555,7 @@ export async function syncAccount({
   unsub()
 
   const query = njsAccount.queryOperations()
-  const ops = await timeoutTagged('ops', 30000, query.complete().execute())
+  const ops = await timeoutTagged('ops', 5 * 60 * 1000, query.complete().execute())
   const njsBalance = await timeoutTagged('getBalance', 10000, njsAccount.getBalance())
 
   const syncedRawAccount = await buildAccountRaw({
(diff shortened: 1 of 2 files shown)


src/helpers/libcore.js
```diff
@@ -555,7 +555,7 @@ export async function syncAccount({
   unsub()
 
   const query = njsAccount.queryOperations()
-  const ops = await timeoutTagged('ops', 30000, query.complete().execute())
+  const ops = await timeoutTagged('ops', 5 * 60 * 1000, query.complete().execute())
```
gre: @meriadec  can you provide another PR to connect this SYNC_TIMEOUT ? ultimately i'm not sure these timeouts make sense because if we don't interrupt the underlying .execute() we probably shouldn't even timeout?
it would be great if we had a way to interrupt.. or we'll need to refactor our app so a sync is not blocking the rest of the accounts, just one account is blocked

gre: the paradigm is, do we want to interrupt things at all , or just assume everywhere that a sync potentially never end and just treat this nicely in our app.. tricky question

gre: typically , we want to interrupt syncs during critical parts (e.g. if you scan accounts, sign transactions, .. we don't want to be busy on doing other background syncs)

PR #1498, opened 2018-09-07
Polish error messages and transaction verification instructions

dasilvarosa:
### Type

UI Polish

### Context

- Education: app uninstallation does not affect assets
- Reinforce wording for receive address verification 

Manager
Receive flows


meriadec:
@dasilvarosa repo settings now require commits to be signed ;) more inf here https://help.github.com/articles/signing-commits-using-gpg/

PR #1510, opened 2018-09-11
remove the nonce mecanism because don't work for DGB

gre:
also reinforce some failsafe:
- try catch the command execution
- mecanism to flush the cache in case an account was not found in libcore

### Type

bugfix + code quality


meriadec:
+1 for grouping `isSegwit` & pals

PR #1525, opened 2018-09-19
 Add the metadata to build the snap

### LINKED ISSUE #1431: Snap Packaging for Linux Distributions
To the Ledger Wallet Team:

I'm requesting that you create a Snap Package downloadable version for Linux distributions.

ADVANTAGES:

1. will work with all Linux distributions because dependencies are included in package
2. upgrades automatically
3. is sand-boxed for security
4. Is listed and advertised in Ubuntu Software 
5. makes it easier and more visible for Linux Distro. users who are usual crypto. early adopters
6. etc...

You can find instructions at https://snapcraft.io/

Thanks for your time,
Mike D.

elopio:
This uses electron-builder to build the ledger-live snap.

### Type

Feature. 

### Context

Fixes #1431.

yarn dist will generate the snap. It just requires snapcraft installed (`sudo apt install snapcraft`)
To install the generated snap: `sudo snap install dist/*.snap --dangerous`
To connect the raw-usb interface: `sudo snap connect ledger-live-desktop:raw-usb`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ yarn start
 
 ```bash
 # Build & package the whole app
-# Creates a .dmg for Mac, .exe installer for Windows, or .AppImage for Linux
+# Creates a .dmg for Mac, .exe installer for Windows, or .AppImage and .snap for Linux
 # Output files will be created in dist/ folder
 yarn dist
 ```
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -24,6 +24,9 @@ linux:
     - target: AppImage
       arch:
         - x64
+    - target: snap
+      arch:
+        - x64
 
 win:
   artifactName: ${name}-${version}-${os}-${arch}.${ext}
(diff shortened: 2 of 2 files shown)


meriadec:
interesting ! thanks for the PR.
Curious about the `--dangerous` flag (can be alarming). Do you know if there is a way to get rid of it?

elopio:
Yes, that's only to test a local build that comes unsigned, maybe from an unknown, unsafe place. In this case it's all good because you are building it yourself, but if I sent the snap to you that would be very dangerous.
Once you push it to the store, it will be signed frome a trusted source, so you'll just have to do `sudo snap install ledger-live-desktop`

gre:
we have no plan for now to use `snap`, we will continue to distribute the application with .AppImage so we have control on the auto update mechanism. Alternatively you can always build a custom version of the app. Thanks

PR #1537, opened 2018-09-25
Kill libcore when reset

meriadec:
Not sure it really fix anything, but users have reported having solved some problems by manually removing their `sqlite` folder, while it's normally done by clear-cache/hard-reset. So let's kill the internal (libcore) process before touching the folder.

### DIFF
--- a/src/commands/killInternalProcess.js
+++ b/src/commands/killInternalProcess.js
@@ -1,0 +1,18 @@
+// @flow
+
+import { createCommand, Command } from 'helpers/ipc'
+import { of } from 'rxjs'
+
+type Input = void
+type Result = boolean
+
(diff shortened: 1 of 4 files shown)


src/commands/killInternalProcess.js
```diff
@@ -0,0 +1,18 @@
...
+import { of } from 'rxjs'
+
+type Input = void
+type Result = boolean
+
+const cmd: Command<Input, Result> = createCommand('killInternalProcess', () => {
+  setTimeout(() => {
+    // we assume commands are run on the internal process
+    // special exit code for better identification
+    process.exit(42)
```
gre: obamanotbad

PR #1540, opened 2018-09-25
Ripple: handle case the destination does not exist

gre:
error if trying to send XRP to a non existing accounts and if not enough funds to create it.

### Type

improvement

### Context

LIVE-252

ripple / send

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 package.json
+test-e2e/**/*.json
--- a/static/i18n/en/errors.json
+++ b/static/i18n/en/errors.json
@@ -99,6 +99,9 @@
     "title": "Oops, insufficient balance",
     "description": "Make sure the account to debit has sufficient balance"
   },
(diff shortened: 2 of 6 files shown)


.prettierignore
```diff
@@ -1 +1,2 @@
 package.json
+test-e2e/**/*.json
```
meriadec: :+1:

static/i18n/en/errors.json
```diff
@@ -99,6 +99,9 @@
     "title": "Oops, insufficient balance",
     "description": "Make sure the account to debit has sufficient balance"
   },
+  "NotEnoughBalanceBecauseDestinationNotCreated": {
+    "title": "Destination does not exist. Send at least {{minimalAmount}}"
```
gre: @dasilvarosa can you think of a better wording or it is ok?

dasilvarosa: "Recipient address is inactive. Send at least {{minimalAmount}} to activate it"

Issue #1386, opened 2018-08-06
Ledger device: Internal error, please report (0x6f01)


dmflahive:
Ledger device: Internal error, please report (0x6f01)

I get this error every time I try to send currency. On both MacOS and Windows.

meriadec:
Can you specify which app is used?

dmflahive:
Both BTC and LTC give this error. 

How is this invalid? My currency is currently locked in the wallet and I can't get it out.

meriadec:
The error `0x6f01` indicates a technical problem while communicating with the device (see [btchip-doc](https://ledgerhq.github.io/btchip-doc/bitcoin-technical-beta.html#_status_words)). You can start testing with another usb port, another usb cable, another computer and see if it's working.

If still not working, can you try sending your funds with the [legacy Chrome app](https://chrome.google.com/webstore/detail/ledger-wallet-bitcoin/kkdpmhnladdopljabkgpacgpliggeeaf)?

dmflahive:
Thank you for that information. I have already tried a different cable and USB port and it still doesn't work, but I can accept that it may not be the software. I'm using a 2018 MacBook Pro with a usb-c to usb-a adapter I get the feeling there may be something lost in the conversion there. 

I see that the legacy app is working now. When I submitted the ticket the legacy app would force me to download Ledger Live without the option to use the app.

dmflahive:
So the legacy app is working fine, I just sent funds using it. Ledger Live is still giving the following error: Ledger device: Internal error, please report (0x6f01). Seems like it must be a software issue if the legacy app works but Live doesn't using the same wallet, same USB port, etc...

If you are able to reproduce the issue and would like for me to test on my end just let me know.

meriadec:
Thanks for the details. As you said, if legacy app is working but not Live, it's probably a software problem (there is also one other possibility: Chrome app is using U2F for USB connection, while Live is using HID).

I reopen this

northbound1:
Same here sending LTC 
Win 10. 

![ledger live error](https://user-images.githubusercontent.com/15316281/44268641-1cb31c00-a200-11e8-868a-c925b9ff50c9.GIF)

meriadec:
We may want to try to add delay between each APDU call (parametrable with environment variable). Seen with firmware team, it can happen when device is sollicitated too much.

samacumen:
[...]

PR #1549, opened 2018-09-29
Stress less the device

gre:
This is an attempt to address #1386 . according to @btchip the internal error experienced by some users might be caused by a too fast exchange with the device / apdu interleaved.

**As no one from our develop team was able to reproduce the problem,** we encourage our community to help us testing this, investigating on the problem and experimenting solutions. Thanks!

---

This PR changes will attempt to fix the problem by locking more time before and after each "atomic" exchange with the device, done by this withDevice function.

To be clear: in current implementation of Ledger Live, all code that exchange with the device is always run in a queue, never in parallel so the only possible thing that can happen is that two "commands" get executed one after the other without any delay.

By "commands" i mean an atomic series of APDUs. for instance signing a transaction is unified into one command (even if it's multiple APDUs). Today we are already protected from access multiple time the device and having "interleaved" APDUs **within the Ledger Live app**. NB the "within the LL app" , because this does not actually protect from another app accessing the device at the same time, we **can't** protect us from that.. So if users use both Ledger Live and another app using the device (examples: MEW / Radar Relay / any ddex / any GUI like monero,decrediton,whatever) , you're going to have issues.

### Type

bugfix attempt

### Context

#1386

all part accessing the device

### DIFF
--- a/src/config/constants.js
+++ b/src/config/constants.js
@@ -46,6 +46,8 @@ export const SYNC_BOOT_DELAY = 2 * 1000
 export const SYNC_PENDING_INTERVAL = 10 * 1000
 export const SYNC_MAX_CONCURRENT = intFromEnv('LEDGER_SYNC_MAX_CONCURRENT', 1)
 export const SYNC_TIMEOUT = intFromEnv('SYNC_TIMEOUT', 5 * 60 * 1000)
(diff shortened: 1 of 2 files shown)


gre:
if this still would not fix the problem, another attempt could be to do a delay at the **APDU level**, but we would have to address it in ledgerjs repo, hid transport.

meriadec:
not sure still relevant

Issue #1468, opened 2018-08-29
Correct mouse pointer everywhere


meriadec:
actually the `cursor: text` has "leaked" in place it shouldn't have :D
e.g: on last step receive flow, the "copy address" button has cursor text, or in onboarding, the choices cards have cursor text, etc.

juan-cortes:
Seeing this has been self assigned to @meriadec does it mean it no longer needs someone to do it? Latest version I have still has this issue. I don't mind looking for those rogue cursors throughout the app

meriadec:
Hey Juan, ahah you're right, that's one of the issue being delayed again and again because annoying but not critical :) I've not started on my side so yes it would be super appreciated !

meriadec:
To me the best way to solve is to remove the `cursor text` we have at some top levels, make every cursor inherits from `cursor default`, then enable `cursor text` only on useful parts (when selecting text makes sense)

juan-cortes:
Sounds like a plan! I'll look into it

juan-cortes:
Anything else needed here since it's merged?

meriadec:
Nope I think it's good for now! Thanks :rocket:

PR #1552, opened 2018-09-30
Issue 1468, wrong cursor on receive modal

juan-cortes:
On the receive modal, the full box was affected by that style as opposed to only the input.
I couldn't find any other places where the cursor was wrong but if any of you can point out more of them, we can tie them together in this pr.

sep-30-2018 18-26-08

### Type

Bug Fix

### Context

https://github.com/LedgerHQ/ledger-live-desktop/issues/1468

### DIFF
--- a/src/components/base/Box/Box.js
+++ b/src/components/base/Box/Box.js
@@ -50,8 +50,7 @@ export default styled.div`
     p.selectable &&
     `
     user-select: text;
-    cursor: text;
-  `};
+    `};
 
   ${p =>
     p.sticky &&


gre:
thanks!

might worth checking on the onboarding (can be triggered from Settings>Help) , i think the main purpose was to show that texts are selectable there.

juan-cortes:
> might worth checking on the onboarding (can be triggered from Settings>Help) 

on it

Arnaud97234:
Tested and it works.

Issue #1428, opened 2018-08-16
No window size & position save after app restart


andreasmaks:
#### Ledger Live Version and Operating System

- tested on Ledger Live v1.1.6
- Platform and version: Mac 10.13.6

#### Expected behavior

- resize and repositioning window
- reopen App 
- app opens in **new** size and position

#### Actual behavior

- resize and repositioning window
- reopen App 
- app opens in **default** size and position

andreasmaks:
Any updates on this?

gre:
the issue is that

https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/main/app.js#L72

and

https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/main/app.js#L80

seems to be saved like in a different db location

https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/main/app.js#L98-L99


it's an easy fix if anyone want to tackle it down.

juan-cortes:
Added a pr for this since it was already obvious how to fix it :trollface:

juan-cortes:
This is done :)

clintnetwork:
Not really done,

When you maximize app, you close and you reopen the maximized state is not saved.

![image](https://user-images.githubusercontent.com/5221349/63531490-e9892400-c508-11e9-9088-9a928b888229.png)

juan-cortes:
It's working on mac, damn..

PR #1553, opened 2018-09-30
Solves issue 1428, brings back persistent window params

juan-cortes:
As pointed by @gre,  settings were being stored/retrieved from different ns in the db, I refactored it so we update both size and position on both events since resizing a window from the left/top will also alter its position but wont trigger the save since technically it's a move event.

### Type

Bug Fix

### Context

https://github.com/LedgerHQ/ledger-live-desktop/issues/1428


meriadec:
thanks

Issue #1010, opened 2018-07-10
Linux: Internal process error (null)


suvayu:
#### Ledger Live Version and Operating System

- Ledger Live **1.0.1**
- Platform: **Linux**

#### Expected behavior

I want to add my existing account for my Ledger Nano S.  Following which, I should be able to sync and access my wallet.

#### Actual behavior

When I follow the steps to add an account, I get the following error.

    Something went wrong during synchronisation, please try again.
    Internal process error (null)

There is no backtrace in the terminal.

#### Steps to reproduce the behavior

1. Click add an account
2. Select the blockchain (Bitcoin, Litecoin, etc)
3. Unlock your Ledger device with your pin
4. Open the corresponding app on the Ledger device
5. Press continue

**Screenshots**

after step (4)
![ledger-live-2](https://user-images.githubusercontent.com/229540/42503307-e239d65e-8427-11e8-86e2-3e4ffc6a9830.png)

after step (5)
![ledger-live-3](https://user-images.githubusercontent.com/229540/42503308-e26dd3aa-8427-11e8-912a-b7595e2b3db9.png)

meriadec:
can you provide the output of
```bash
uname -a
```

suvayu:
$ uname -a
    Linux hostname 4.17.2-200.fc28.x86_64 #1 SMP Mon Jun 18 20:09:31 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

bocekm:
Duplicate to https://github.com/LedgerHQ/ledger-live-desktop/issues/950

gre:
are you able to import your accounts on the same computer but with the Ledger Bitcoin Chrome app ?

suvayu:
I have not tried importing my accounts with the chrome app (as I have never wiped my wallet), but I have been successfully using the Chrome app and Electrum ever since I initialised it after my purchase.

snaggen:
I have this exact issue on Ledger Live 1.0.1 on Fedora Linux 28. However, the chrome app works fine.

drizzt:
Hi,
after a poor-man debugging I found that the problem is fixable by using the following libraries from Ubuntu 18.04: libgssapi_krb5.so.2, libk5crypto.so.3, libkrb5.so.3 and libkrb5support.so.0
[...]

PR #1557, opened 2018-10-02
Patch the .AppImage with known working libs to fix openssl crash

meriadec:
This commit fixes the crash occuring on some configuration (reported only on Fedora 28, but since it comes from a library conflict, it can affect a larger spectrum) when syncing BTC-like account.

To give more context, libcore is using libcrypto symbols, and should use the libcore-embedded libcrypto, *BUT* electron is loading the system one before and it result in a SIGSEGV when using the Fedora 28 openssl version ([list here](https://rpmfind.net/linux/rpm2html/search.php?query=libcrypto.so.1.1()(64bit))). More details on the crash can be found on the original issue.

With @KhalilBellakrid, we first tried to address the problem by changing the runtime path of lib-ledger-core (changing `LD_LIBRARY_PATH` when launching the app was solving the issue.. but it could not be made persistent). Even with successfully changing the `RPATH` inside binaries, the incorrect `libcrypto` was still loaded, and that's because.. electron load it before everything.

**TL;DR:** The fix is hacky. It basically unpack the `.AppImage`, put required libs inside, then repack it. Not perfect. But the issue has been here for too long, and has led to lot of deception. I've tested it against auto-update and I confirm it works. (As you see in the commits, I couldn't achieve to interject _inside_ `electron-builder` workflow, so had to add dedicated script to upload gh assets.. hope we find a way to get rid of that).

*(poor-quality) video of the fix in action:*

[](https://www.youtube.com/watch?v=rl5GXXdjkUY)

### Type

Bugfix

### Context

https://github.com/LedgerHQ/ledger-live-desktop/issues/1010

Parts of the app affected

To reproduce:
- install Fedora 27
- upgrade to Fedora 28 (via `dnf upgrade`, etc.) - *note to @Arnaud97234 : we have a machine on this configuration)*
- install any release before this fix (e.g [v1.2.0](https://github.com/LedgerHQ/ledger-live-desktop/releases/tag/v1.2.0))
- sync BTC-like account
- **should fail** with `Internal process error (null)` (openssl crash, called by libcore)
- build the app with `yarn dist`
- patch the app with `bash scripts/patch-appimage.sh`
- launch the app
- **should work**


Arnaud97234:
Tests done.
BTC like coin are working properly : Sync, Operations.
No regressions on other coins (ETH, XRP ...)

meriadec:
As AppImageKit use [continuous](https://github.com/AppImage/AppImageKit/releases) release system, it would be problematic to store the shasum because it could cause false positive if they push a new release. One solution will be to store our own version on S3 :D WDYT ?

KhalilBellakrid:
Yes I thought also about that and I think the best solution would be to push those libraries from Libcore CI to s3 bucket (with their checksum) so that way we always have 'compliant' versions and we get the updates if there are some.

meriadec:
appimagekit will be pushed to S3 tomorrow by infra team.

@gre do we wait or do we merge for now?

gre:
let's wait it's ready

PR #1562, opened 2018-10-03
Merged language files into a single file for easier translation

valpinkman:


gre:
Is there a way we can remove the need to do app: ? I think we can tell the i18n lib the default

juan-cortes:
I don't have permissions to this for you guys (since I'm just the random dev from spain) but it's a matter of adding `defaultNS: 'app',` to [renderer/i18n/instanciate.js#L5](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/i18n/instanciate.js#L5), this changes the default from `'translation'` to `'app'`. Allowing us to remove all the `app:` prefixes.

valpinkman:
@juan-cortes thanks, done

gre:
@Arnaud97234 could you check all texts are fine on the current develop app? normally it's all good but we never know 👍

gre:
just checking there is nowhere text like `app.foo.bar` ^^

Arnaud97234:
@gre, I've made checks on all screen of the app. Everything seems correct.

gre:
thanks!!!

PR #1571, opened 2018-10-10
Correctly handle sequence when UTXO script length is 0

meriadec:
TL;DR, following #1386, we identified & reproduced `0x6f01` error, occurring in the specific case of having a 0 length script in inputs, which resulted in not passing sequence in the transaction. So the fix is located in [ledgerjs/hw-app-btc](https://github.com/LedgerHQ/ledgerjs/commit/2b3ed5a60b4e4b42613d7c773a67fb7d42c22e12).

Thanks @btchip, @pollastri-pierre and @KhalilBellakrid for helping on this one! :rocket:

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,11 +34,11 @@
     }
   },
   "dependencies": {
-    "@ledgerhq/hw-app-btc": "4.21.0",
-    "@ledgerhq/hw-app-eth": "^4.14.0",
-    "@ledgerhq/hw-app-xrp": "^4.13.0",
-    "@ledgerhq/hw-transport": "^4.13.0",
-    "@ledgerhq/hw-transport-node-hid": "4.22.0",
+    "@ledgerhq/hw-app-btc": "4.24.0",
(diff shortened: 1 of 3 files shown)


gre:
needs important QA testing on if everything still work fine. also would deserve to **build** the app and test a build , to double check babel change things.

meriadec:
+1 for the test plan (and it will be tricky to test: I don't reproduce anymore since my UTXOs now not anymore contains empty scripts. I guess we will need faucets random luck.

@Arnaud97234 don't hesitate if you need my help

gre:
i mean. the utxo fix is just bumping some libs. i was more talking about the rest of this PR :p

meriadec:
I splitted into another PR https://github.com/LedgerHQ/ledger-live-desktop/pull/1573 so it's more atomic

Arnaud97234:
Tested with Bitcoin Testnet and didn't get the error.
No regression found (Currencies used BTC, LTC, DGB, BCH, ETH, XRP).

Issue #1576, opened 2018-10-11
Improve Identifier display in firmware update


dasilvarosa:
During a firmware update the Identifier is shown on the device and should be verified on the Manager. 

- Ledger Nano S and Ledger Blue only show the first 4 and last 4 characters of the Identifier: `[ABCD...1234]`.

- Ledger Live shows the entire string. This should be reduced to match the characters displayed on the device.
![](https://support.ledgerwallet.com/hc/article_attachments/360008515234/identifier.png)

gre:
easy to fix. hard to test.

juan-cortes:
This is done

gre:
Thanks a lot

PR #1579, opened 2018-10-12
Addresses the inconsistency between firmware hashes displayed

juan-cortes:
Makes the hash displayed more consistent between the device and the app when upgrading the firmware. Might want to change the box's width to accommodate the smaller hash, or increase the font.

### Context

https://github.com/LedgerHQ/ledger-live-desktop/issues/1576

Can be tested by hardcoding a lower version on the[ `getCurrentFirmware` helper](https://github.com/LedgerHQ/ledger-live-desktop/blob/d90543c4a0fdeea70faf4ca378a5360a4ab90605/src/helpers/devices/getCurrentFirmware.js#L14). Although it can't really be installed since the device rejects the installation, I'm assuming due to being the same version

screenshot 2018-10-12 at 17 36 42

### DIFF
--- a/src/components/modals/UpdateFirmware/steps/01-step-install-full-firmware.js
+++ b/src/components/modals/UpdateFirmware/steps/01-step-install-full-firmware.js
@@ -134,7 +134,7 @@ class StepFullFirmwareInstall extends PureComponent<Props, State> {
     }
 
     hash = hash.toUpperCase()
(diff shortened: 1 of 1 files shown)


gre:
awesome 🚀

juan-cortes:
@dasilvarosa I'm on my phone but I think it's done and committed. I initially had the brackets on because that's how you formatted it on the [linked issue](https://github.com/LedgerHQ/ledger-live-desktop/issues/1576) though. Hope it's all good now.

Issue #1588, opened 2018-10-17
Regression - Add account / Send modals, cursor not auto-focused input on opening


Arnaud97234:
**Bug found**
ledger-live-desktop develop branch.
When add account or Send modals opens, the cursor is not focused on the input by default.

**How to reproduce**
Click add account button in the sidebar menu.
![screen shot 2018-10-17 at 11 53 01](https://user-images.githubusercontent.com/13007150/47078340-679edf00-d203-11e8-83f7-9cfce0c3c68f.png)

**Expected behaviour**
Cursor should be in input when modals opens.
![screen shot 2018-10-17 at 11 54 05](https://user-images.githubusercontent.com/13007150/47078355-6cfc2980-d203-11e8-9a8f-47aa7259be75.png)

juan-cortes:
By running `document.activeElement` on both screens we see that the `autoselect` is not propagating down into the input itself but rather staying in the modal Wrapper. This appears to come from `Moldal/index.js` line 186, where we call `domWrapper.focus()` 

This comes from a commit from @meriadec  [642e72e](https://github.com/LedgerHQ/ledger-live-desktop/commit/642e72e826cf1eb2e70784aa731699d4c01356f4#diff-265374f8f9ad6ee48a8650445943fa7cR144) but it's from back in June. I played with having that line removed and I couldn't find any odd behaviour, surely enough we get the autofocus back into the currency/recipient. 

If you say it's ok to remove it I have a PR ready to request.

meriadec:
fixed.

PR #1590, opened 2018-10-17
Disable ModalWrap focus grab if lastFocus is inside it already

juan-cortes:
Simple bug fix that should not affect any functionality apart from the one specified in the issue. The autofocus prop was being ignored or rather overridden by a focus from `ModalWrapper` that was in place to ensure a more natural accessibility flow for keyboard usage.

By checking if the focused element prior to the focus grab is already inside the Modal we guarantee that any other cases that were using this `focusWrapper` functionality still use it.

### Type

BugFix
### Context

https://github.com/LedgerHQ/ledger-live-desktop/issues/1588


juan-cortes:
@meriadec I think this solution doesn't break the functionality you had in place, let me know if you think otherwise

meriadec:
Very elegant fix :+1: 
Testing on my side.

PR #1595, opened 2018-10-22
New version of qrcode exporter

gre:


src/components/SettingsPage/sections/Tools.js
```diff
@@ -30,6 +31,7 @@ class TabProfile extends PureComponent<*, *> {
         <Box flow={2}>
           <QRCodeExporter />
         </Box>
+        <Box style={{ textAlign: 'right' }}>{liveCommonPkg.version}</Box>
```
meriadec:
```
<Box align="right">
```
:p

gre: it doesn't work

gre: rollbacking. btw not very important because new design to come later.

meriadec: my bad, it's `textAlign="right"`, the `align` prop is bound to `align-items`

meriadec: nvm

Issue #1594, opened 2018-10-18
absurd memory usage


bumbar1:
#### Ledger Live Version and Operating System

- tested on Ledger Live **1.2.0**
- Platform and version: **elementary os 0.4.1**

#### Expected behavior

app should play nicely with system, not abuse it

#### Actual behavior

app doesn't seem to stop consuming memory

#### Steps to reproduce the behavior

picture 1, after left running for a while
![screenshot from 2018-10-18 19 58 28](https://user-images.githubusercontent.com/2507237/47175624-df5c2f00-d313-11e8-8eef-6ef0996de6b0.png)

picture 2, restarted and said it was synchronized 
![screenshot from 2018-10-18 20 06 15](https://user-images.githubusercontent.com/2507237/47175626-df5c2f00-d313-11e8-8a2c-478002053e68.png)

picture 3, restarted after sync finished, kept resyncing, memory usage just goes up... (got a few http 500 errors on this stage)
![screenshot from 2018-10-18 20 19 14](https://user-images.githubusercontent.com/2507237/47175627-df5c2f00-d313-11e8-86bb-a573e52a1c6a.png)

gre:
cc @KhalilBellakrid we need to investigate this. i suspect the sync is going wild

juan-cortes:
@bumbar1 it would be interesting to know which cryptos you have installed when this happened. Syncing can trigger several processes depending on which account is being synced, perhaps there's a bug in a specific one and we can narrow it down faster.

KhalilBellakrid:
@gre this is huge, you have to be sure that you kill spawned processes where libcore got instanciated (I've seen an error in code where you were not killing it as it should, there was an await missing) 
Did we change the way we instanciate libcore in Live ? 
@bumbar1 were you having this issue before 1.2.0 ? And by the way how many accounts do you have on Live ? 
Thank you for your help!

KhalilBellakrid:
Hmm I see, then we will need to analyze/profile libcore bindings ASAP (already did that with Valgrind for libcore but nothing alarming was revealed)

gre:
actually I have an idea of what's happening. it's a classical back pressure problem:

[...]

PR #1596, opened 2018-10-22
Added option of non-bitcoin currencies for testing in devtools

juan-cortes:
While researching https://github.com/LedgerHQ/ledger-live-desktop/issues/1594 I was unable to replicate the situation on my machine, neither could I replicate it on a vm running the same OS as the user. Under the assumption that what was probably causing this had more to do with the number of accounts and transactions than anything else and knowing my own accounts did not have enough movement I searched and found the devtools that allow you to add some bitcoin-family account by xpub.

Since Ethereum wasn't one of them and trying to easily add other cryptos simply by the public address I modified the dev tools page a bit, allowing Ethereum and Ripple (initially) to have accounts manually added. **I don't know if adding the method as an optional one to the bridges is a good approach** or if we'd rather abstract it out of there, but for now this works and might be useful for testing. Also handles non-bitcoin family coins with unsupported bridges without crashing.

image

I noticed that adding a mammoth account such as one used by an exchange **will never sync**, or at least it will take forever and apparently it will not sync other accounts meanwhile either. I couldn't see a huge spike in memory usage however so that might not have anything to do with it.

**However**, if we pass the `limit` option to the `getTransactions` call at the Ripple bridge, we do get the results back, seems like it's choking on hundreds of operations. In Ethereum's bridge, the "Show more" implementations seems to work a treat and doesn't choke even on thousands of transactions.

Didn't polish this since a normal user wouldn't be seeing it but I did change the inputs to ask for xpub/public address depending on the family. So the initial sync will not give the amount + transactions until we manually trigger it.

### DIFF
--- a/src/components/DevToolsPage/AccountImporter.js
+++ b/src/components/DevToolsPage/AccountImporter.js
@@ -23,6 +23,7 @@ import { CurrencyCircleIcon } from 'components/base/CurrencyBadge'
 import { idleCallback } from 'helpers/promise'
 
 import scanFromXPUB from 'commands/libcoreScanFromXPUB'
+import { getBridgeForCurrency } from 'bridge'
 
 const mapDispatchToProps = {
   addAccount,
@@ -33,6 +34,7 @@ type Props = {
 }
 
 const INITIAL_STATE = {
+  xpubLabel: 'xpub',
   status: 'idle',
   currency: null,
   xpub: '',
@@ -43,6 +45,7 @@ const INITIAL_STATE = {
 }
 
 type State = {
+  xpubLabel: string,
   status: string,
   currency: ?Currency,
   xpub: string,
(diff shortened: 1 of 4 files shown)


src/components/DevToolsPage/AccountImporter.js
```diff
@@ -79,23 +82,39 @@ class AccountImporter extends PureComponent<Props, State> {
...
+        const derivationMode = isSegwit
+          ? isUnsplit
+            ? 'segwit_unsplit'
+            : 'segwit'
+          : isUnsplit
+            ? 'unsplit'
+            : ''
+        const rawAccount = await scanFromXPUB
+          .send({
+            seedIdentifier: 'dev_tool',
```
meriadec: we may want to add xpub here
wdyt @gre ?

gre: we could yeah, it really depends if you want the DevTools feature to be used with multiple different seeds at same time. To me the DevTools should never be used by random users but only devs & in a dev session (like you basically Reset App at the end of this session), in that case it's not necessary to use a seedIdentifier )

meriadec:
i'm afraid this needs to be refactored because of [those changes](https://github.com/LedgerHQ/ledger-live-desktop/commit/151b9241dfb4b6255fa430a4d51eaafce663ec8e#diff-e132b9b8a1d48674498900cb00074bb7)

PR #1599, opened 2018-10-25
add new derivations & isIterable system

gre:
integrate latest live-common with new derivations on ETH.

### Type

polish

### Context

https://github.com/LedgerHQ/ledger-live-common/pull/88

test plan: on https://ledger-live-tools.netlify.com/ make sure you see the live-common version under in a footer (otherwise use https://5bd1c82ce39e7c3655d8bb4c--ledger-live-tools.netlify.com/ ) and go to derivations to see the new address. send a minimal amount in new eth address to check that addresses are correctly discovered. make sure there is no discovery regression on the other paths.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,11 +37,11 @@
   "dependencies": {
     "@ledgerhq/hw-app-btc": "4.24.0",
     "@ledgerhq/hw-app-eth": "^4.24.0",
-    "@ledgerhq/hw-app-xrp": "^4.24.0",
+    "@ledgerhq/hw-app-xrp": "^4.25.0",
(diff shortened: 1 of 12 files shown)


meriadec:
looks like a simple deps conflict

Arnaud97234:
Test with path 44'/60'/14'/5'/16 and 44'/60'/0'/0' for ETH  and 44'/144'/0'/0' for XRP.
New addresses are discovered during scan and accounts are added properly in Ledger Live.

Didn't test for XRP path 44'/144'/14'/5'/16 because I don't want to block an other 20 XRP.

PR #1600, opened 2018-10-25
🌲 ZCash update

gre:
Update to support ZCash sapling update.

### Type

feature

### Context

https://z.cash/upgrade/sapling/

ZCash


src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -144,7 +144,9 @@ class StepImport extends PureComponent<StepProps> {
             })
           }
         },
-        complete: () => setScanStatus('finished'),
+        complete: () => {
+          setScanStatus('finished')
+        },
```
meriadec: wesh

PR #1606, opened 2018-10-26
Ethereum libcore integration

meriadec:
**This is still a WIP.**

The purpose of this PR is to open discussion about the optimal way to handle particular behaviour based on currency, in LibcoreBridge. Here, mostly dirty `if` on `currency.family` to make scan/tx logic stay as close as possible between BTC/ETH.

What's included?

- [x] ETH scan accounts (& ops), on all derivations
- [x] ETH transaction (sign & broadcast)
- [x] UI integration / reconnect of EthereumKind fields in LibcoreBridge

What's not included?

- [ ] flow-type (I didnt even test)
- [ ] beautiful code


gre:
we're going to close this. keep this PR as reference (we'll probably look again at it when we impl the LibcoreEthereum bridge)

gre:
@meriadec  don't remove your branch basically :p

PR #1609, opened 2018-10-28
Handle sqlite gaps while scanning accounts

meriadec:
LL-329

TL;DR, the PR should **only** be merged once [this line](https://github.com/KhalilBellakrid/lib-ledger-core/blob/3fc65779c3ac1f82215e0cec8f80f511639a460c/core/src/wallet/common/database/AccountDatabaseHelper.cpp#L77) is updated with `ORDER BY idx` in lib-ledger-core, to fix the `getNextAccountCreationInfo` (actually it returns always index `0` when there is gaps in created accounts).

to reproduce the problem:

- open LL with empty user data
- scan currency with (at least) 3 accounts
- import the 3nd one
- clear cache
- scan again
- error is throw (account 0 does not exists)

to test the resolution

- switch to this branch
- do same thing and
- realize it doesnt throw
- shutdown your computer

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@ledgerhq/hw-app-xrp": "^4.24.0",
     "@ledgerhq/hw-transport": "^4.24.0",
     "@ledgerhq/hw-transport-node-hid": "4.24.0",
-    "@ledgerhq/ledger-core": "2.0.0-rc.9",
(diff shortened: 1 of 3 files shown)


meriadec:
@KhalilBellakrid can you give an ETA on when you will be able to publish a new libcore version including the patch?

Arnaud97234:
Test with bitcoin and zcash accounts. It works properly.

PR #1615, opened 2018-10-30
Implement CurrenciesStatusBanner to provide a way to inform infra pbs

meriadec:
**HODL**: we need to deploy [ll-status-api](https://github.com/LedgerHQ/ledger-live-status-api) and set endpoint [here](https://github.com/LedgerHQ/ledger-live-desktop/blob/53d13b8e9a3a4039c131d35dfc538bee2f4f4f56/src/config/urls.js#L40-L41) before merging.
(An environment var `LL_STATUS_API` can be used for local development.)

#### How it works

- an error/downtime occur in `n` cryptos
- the `status`, `message` and `link` keys are updated manually in ll-status-api
- LL-desktop poll (each 2 minutes for now, configurable via `CHECK_CUR_STATUS_INTERVAL`) the ll-status-api endpoint, and get a list of failing cryptos
- Display a banner for each of them (absolute positioned, visible on all screens)
- Ability to dismiss banner

*I also have the subtle feeling that it would need a little design polish.*

2018-10-30_1019x477


src/config/constants.js
```diff
@@ -32,6 +32,7 @@ export const MIN_HEIGHT = intFromEnv('LEDGER_MIN_HEIGHT', 700)
 export const CHECK_APP_INTERVAL_WHEN_INVALID = 600
 export const CHECK_APP_INTERVAL_WHEN_VALID = 1200
 export const CHECK_UPDATE_DELAY = 5000
+export const CHECK_CUR_STATUS_INTERVAL = intFromEnv('CHECK_CUR_STATUS_INTERVAL', 2 * 60 * 1000)
```
gre: should increase to 30mn or 1h IMO

meriadec: updated to 1h

src/reducers/accounts.js
```diff
@@ -86,6 +87,10 @@ export const accountSelector = createSelector(
   (accounts, accountId) => accounts.find(a => a.id === accountId),
 )
 
+export const accountsCurrenciesSelector = createSelector(accountsSelector, accounts =>
```
gre: we already have currenciesSelector

meriadec: updated

meriadec:
Updated design

err

gre:
@meriadec there is one important thing in the scope of this feature.

If some currencies are down, all related accounts must stop attempting to synchronize (when we know something is down, no point trying to sync it).

- it means we likely want to keep in a redux cache the result of `currencies-status` so if you restart the app, it already knows some currency are down.
- some work to do in BridgeSync thing to also connect to this currencies-status redux state

meriadec:
- [x] added a `currenciesStatus` reducer, the fetch is done inside `fetchCurrenciesStatus` action. The reducer contains pretty much the result of the call :)
- [x] added a condition in `BridgeSyncContext` that prevent a "down" currency to be sync

I would like to point out that the banner just display the message (and link), but **doesn't put the currency name or logo anywhere**, so we are responsible to be clear as crystal when we create the message, server side (e.g: no "server is down", but "XRP server is down"). IMO it's a **really bad** decision. I raised the point to design (ignored).

PR #1628, opened 2018-11-09
Decred Integration

KhalilBellakrid:
Preparing for Decred integration

### Type

New coin integration

### Context

Related to https://github.com/LedgerHQ/ledgerjs/pull/239

All Bitcoin like currencies should be tested with all flows : import of accounts, sync and (specially) transaction signing !

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,12 +35,12 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


KhalilBellakrid:
@gre CI fails because it needs bumping version of LedgerJS and ledger-live-common

PR #1641, opened 2018-11-14
LL-448 segwit parsing

amougel:
This will allow any bitcoinlike altcoin account to spend utxo received from segwit transactions.


meriadec:
:thinking_face:

gre:
I also don't understand the fix. it's touching the code of "signAndBroadcast" how does this would help for "**receive** segwit transactions"

amougel:
I meant that right now, a user receiving a segwit transaction is not able to spend it because of the reception of this output because he is not able to parse it properly. So the fix will have an effect o the sending flow indeed.

meriadec:
@amougel if you don't mind, can you put a comment next to the `true` for clarity?

gre:
@amougel ok, i've updated the description because it was not saying that ;)

PR #1644, opened 2018-11-15
Decred integration

meriadec:
Decred integration

### Type

New coin integration

### Context

Related to https://github.com/LedgerHQ/ledgerjs/pull/239

All Bitcoin like currencies should be tested with all flows : import of accounts, sync and (specially) transaction signing !


meriadec:
```
curl http://cdn-fastly.deb.debian.org/debian/pool/main/s/systemd/libudev-dev_232-25+deb9u4_amd64.deb
404
```
thanks debian packages

meriadec:
Tested successfully:

- [x] Add accounts
- [x] Receive (display address)
- [x] Sign transaction with device, and [broadcast it](https://mainnet.decred.org/tx/81153d853aa45c1dd5bce2f786dde60820d691ca14757c90f5d758232a6c1079)

Ok for me :ok_hand:

meriadec:
Digged with @KhalilBellakrid, the problem is caused by decred nodes, that's returning `200` also for bad transactions (at broadcast), then libcore assume it's correct, then crash.

meriadec:
This is fixed now, we have correct error when incorrect fees.

2018-12-03-180149_606x554_escrotum

@gre do you think it's acceptable in term of UX for now? (pretty ugly error). We can address parsing the error later, to display only the message (I thought we already did that).

gre:
Actually I meant word-wrap:break-word

meriadec:
I'm literally testing that in the current millisecond

meriadec:
`word-break: break-word;` :doge: but for your defense, no developer in the world remember the correct key+value couple.

2018-12-03-181613_562x257_escrotum

PR #1645, opened 2018-11-15
Fix estetic glitch that display "Paused" when a sync is disrupted

gre:
(not tested)

this should fix current visual glitch that says "Paused" because BCH is disrupted.

### Type

patch


gre:
this is not tested so don't merge yet

Arnaud97234:
global Sync state is OK.
Can't add BCH account.
If BCH account is present -> Can't do transaction.
BTC transactions still works.
BCH QR Code is not displayed in mobile app (Can't import BCH by flashing)

PR #1647, opened 2018-11-16
Standardized Searchbar behavior

valpinkman:
Improvement

Import account searchbar and Manager app searchbar


meriadec:
i guess we need the same treshold in mobile

meriadec:
someone just told me it's already the case

PR #1660, opened 2018-11-22
Fix CI that was throwing because debian like removing its own libs

meriadec:


meriadec:
still no circle status, but I guarantee it works https://circleci.com/gh/meriadec/ledger-live-desktop/1579

PR #1661, opened 2018-11-22
Updated stale translations

juan-cortes:
Addresses the stale translations (they wouldn't change on language change) throughout the app. I replaced the appropriate `t` calls with `Trans` elements as suggested by @gre and it works. There doesn't seem to be any non-translateable string in the app as far as I could see.

### DIFF
--- a/src/components/modals/Send/index.js
+++ b/src/components/modals/Send/index.js
@@ -4,7 +4,7 @@ import React, { PureComponent } from 'react'
 import invariant from 'invariant'
 import { compose } from 'redux'
 import { connect } from 'react-redux'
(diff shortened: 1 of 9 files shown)


src/components/modals/Send/index.js
```diff
@@ -136,7 +136,7 @@ class SendModal extends PureComponent<Props, State<*>> {
     }
   }
 
-  STEPS = createSteps({ t: this.props.t })
+  STEPS = createSteps()
```
gre: sounds like `const STEPS` could be defined globally and no longer need for a createSteps() function

gre:
good to ship!

PR #1667, opened 2018-11-25
Moves the account exporter from the experimental to a top level tab

juan-cortes:
Tried to stay as true as possible to the design while removing the references to QrCodes
I kept the experimental tab available but without content right now.

image

image


dasilvarosa:
Will have to change wording in mobile app. I expected the feature to go into General.

juan-cortes:
Maybe there will be several ways of exporting accounts in the future? If not, then it's a bit overkill to have its own section, I agree. I have no issue with moving it there

khalilbenihoud:
Is there any way to decrease the size of the QR Code?

juan-cortes:
I followed the dimension on the zpl file which marked it at 330px, making it smaller due to the density of the data points. [This article explains the minimum _printed_ size ](https://blog.qrstuff.com/2011/01/18/what-size-should-a-qr-code-be). I tested it at 100px and it wont scan, and at 165 and if you are close enough it will without issue. 🤷‍♂️

khalilbenihoud:
@bscholving What do you think?

gre:
I would prefer not decrease the size that much. it's going to be too hard to scan on some screens (e.g. non retina screens).

bscholving:
We keep 330px width. More secure.

juan-cortes:
Following the new wording from @dasilvarosa, pushed the updated literals

image

juan-cortes:
@khalilbenihoud anything stopping this one from moving forward on your end?

bscholving:
@juan-cortes "**s**" is missing on the word _Account_ for the 1st bullet point. Moreover, the numbers in circles seems bigger than the designs in Zeplin, I used font size 12.

gre:
@Arnaud97234 i think it's the wording of the Manager access

static/i18n/en/app.json
```diff
@@ -619,7 +629,7 @@
       "step2": {
         "title": "Did you save your recovery phrase by yourself?"
       },
-      "step3": {
+      "step3":{
```
meriadec: prettier will not like that

juan-cortes: yeah I was looking at it, prettier doesnt douch this file though because its in `static`, change the script in package.json?

src/components/SettingsPage/index.js
```diff
@@ -48,6 +49,11 @@ class SettingsPage extends PureComponent<Props, State> {
         label: <Trans i18nKey="settings.tabs.currencies" />,
         value: SectionCurrencies,
       },
+      {
+        key: 'export',
+        label: <Trans i18nKey="settings.tabs.export" />,
+        value: SectionExport,
+      },
```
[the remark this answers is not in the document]
juan-cortes: done

Luuk2016:
In some old preview screenshots we could see a mobile button at the top right corner:
image
Wouldn't that be a more convenient place?
Is there a reason why it's now in the settings? Or is this gonna be changed if the mobile app releases?

juan-cortes:
I'm guessing the logic goes that after an initial export it won't be a button with much use to justify it being in the main screen. Right now the UI is pretty bare so perhaps a future iteration brings it back to the portfolio, who knows.

PR #1701, opened 2018-12-11
Adds tag to transaction details, XRP

juan-cortes:
Adds the tag to the transaction details on an XRP operation, I also allowed the keys stored in the extra field of an operation to be translatable by going into the `operationDetails.extra.key` format, to avoid having computer-named labels.

image

### DIFF
--- a/src/components/modals/OperationDetails.js
+++ b/src/components/modals/OperationDetails.js
@@ -3,7 +3,7 @@
 import React, { Fragment, Component } from 'react'
 import { connect } from 'react-redux'
 import { openURL } from 'helpers/linking'
-import { translate } from 'react-i18next'
+import { Trans, translate } from 'react-i18next'
 import styled from 'styled-components'
 import moment from 'moment'
 import { getOperationAmountNumber } from '@ledgerhq/live-common/lib/operation'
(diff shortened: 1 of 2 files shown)


src/components/modals/OperationDetails.js
```diff
@@ -226,6 +226,14 @@ const OperationDetails = connect(mapStateToProps)((props: Props) => {
               <OpDetailsTitle>{t('operationDetails.to')}</OpDetailsTitle>
               <DataList lines={recipients} t={t} />
             </Box>
+            {Object.entries(extra).map(([key, value]) => (
+              <Box key={key}>
+                <OpDetailsTitle>
+                  <Trans i18nKey={`operationDetails.extra.${key}`} defaults={key} />
```
gre: missed to commit the translation addition?

Arnaud97234:
@gre @juan-cortes 
Tag is not visible in operation details when transaction is pending (Status "Sending..." in transactions history). Once the transaction is confirmed Tag is displayed in operation details.
You are ok with that ?

src/components/modals/OperationDetails.js
juan-cortes: Nope, it defaults to "key" and it's capitalized, so it's find on desktop

gre: i'm not sure why making it translatable if it's not in translation then 🤔

gre: I guess "Tag" is technical enough to not translate it ? wdyt @dasilvarosa

juan-cortes: I foresaw the possibility of other extra fields using this, since it's a mapping, and not having a friendly key to display. This case on the desktop it capitalizes the key and it shows as "TAG" so it's ok, if the key had been something like "destinationTag" we would use the translation to rewrite it.

If you don't like it we can remove it, on mobile for example since there's no text-transform I had to use the translation for it to show as "Tag"

juan-cortes:
Updated it to account for optimistic updates too, it was showing the tag field regardless.

PR #1705, opened 2018-12-12
Secure update

### LINKED ISSUE #942: Sign all release binaries
- [x] I have checked this feature was not yet requested.

#### Part of the application

Application binary

#### Description

Previous releases such as the Ledger Chrome manager/wallet and Ledger standalone apps had all been signed using the key infrastructures of Google and the Ledger wallet itself. 

However, the binaries offered for release are not signed by PGP or another method. It's therefore impossible to protect against malicious modifications of the binaries. 

A public key for signing should be announced on the Ledger website, and then the current and all future release binaries should be signed with the corresponding private key.

meriadec:
<p align="center">
  
</p>

### Implementing secure updates

The purpose of this PR is to improve security of the update process. Two axis are involved:

- the update patch is checked internally just after being downloaded:
  - sha512sum check against sum file, stored outside of github
  - signature verification of sum file, with Ledger pubkey (embedded inside the app)
- the update doesn't install automatically anymore: a user action is required (click on "Install")

It also brings some refactoring on front side:

- `autoUpdate` check & events, and `quitAndInstall` action has been migrated to "command" pattern. Because they require `electron` dependency, there is now a way to launch a command in the main process instead of the internal process (afb8000eaee306d093b89b3a0331193472c2efeb). It should not affect existing behaviour in any way.
- `updater` reducer has been dropped in favor of `React.Context` pattern for multiple reasons — could develop later —, and multiple consumer components. As a side-effect, React has been bumped because of bug in context refreshing.

As a side effect, it colocates sha512sum of the binaries + their signatures next to the release binaries, which resolves #942.

**How the app handle a Ledger pubkey update?**

The particular scenario here, is if by any chance the public key has to change (it will be a qorum based key, so if any of the qorum people leave, a new key needs to be generated).

If we stay in the naive approach, the next updates will be refused. So the need is to be able to sign app with another key, sign the new key with the previous one, and make the updater verify everything.

```
                      +------------------------+   +--------------------+
                      | KEY SERVER             |   | GITHUB             |
                      |                        |   |                    |
                      | pubkeys/root.asc       |   | ll-1.8.0.AppImage  |
                      | pubkeys/key-2.asc      |   | ll-1.8.0.exe       |
                      | pubkeys/key-2.asc.sig  |   | ll-1.8.0.dmg       |
                      | pubkeys/key-3.asc      |   | ...                |
                      | pubkeys/key-3.asc.sig  |   +---+----------------+
                      | ll-1.8.0.sha512sum     |       |
                      | ll-1.8.0.sha512sum.sig |       |
                      | ll-1.9.0.sha512sum     |       |
                      | ll-1.9.0.sha512sum.sig |       |
                      | ...                    |       |
                      +------------+-----------+       |
                                   |                   |
                                   |     +-------------+
                                   |     |
                                   V     V
                                +--------------+
                                | LIVE DESKTOP |
                                |              |
                                | key-2.asc    |
                                +--------------+

1- check is made with embedded pubkey
2- if check fail, app try to fetch the next key, which is named
   with the fingerprint of the previous one, and signed with the previous one
3- if no key found, error
4- else, jump to step 1
```

**What's next?**

- [ ] I've setup a key-server example here: https://ledger-live.sigsev.io, but we need to **setup it inside our infra**. And **craft some scripts** to feed it with current release shasums, etc. And set the feed URL inside the Live ([here](https://github.com/meriadec/ledger-live-desktop/blob/2751d88cca433d532afb4c15dfda238115cebd37/src/config/constants.js#L101)).
- [ ] The actual [pubkey](https://github.com/meriadec/ledger-live-desktop/blob/secure-update/src/main/updater/ledger-pubkey.js) stored in source code is mine. We need to **put the qorum pubkey**.
- [x] The check is only informative for us for the moment ("white bullet"). This means if the check fail, update is still marked as available (no false positive, etc.). This behaviour can be deactivated by changing the boolean [here](https://github.com/meriadec/ledger-live-desktop/blob/2751d88cca433d532afb4c15dfda238115cebd37/src/config/constants.js#L100).
- [x] The update notification banner has had some changes, and require new wording. cf screenshot below. Could also need cosmetic polish. See screenshots at the end.

<p> </p>
<p> &nbsp;</p>
<p align="center">
  
</p>

### Test plan

All my tests were done on Linux. Even if it shouldn't affect anything, IMO it's worth to test also on other platforms. I generated three PGP keys (`TEST-1`, `TEST-2`, `TEST-3`) to simulate key invalidation. 

  - Download [test-pubkeys.zip](https://github.com/meriadec/ledger-live-desktop/files/2672670/test-pubkeys.zip) and import the three keys:

```bash
$ unzip test-pubkeys.zip -d /tmp/ll-pubkeys
$ gpg --import /tmp/ll-pubkeys/*.asc
```

  - Go to https://github.com/meriadec/ledger-live-desktop/releases
    - here you see:
      - `v2.0.0`, signed with `TEST-1`
      - `v3.0.0`, signed with `TEST-2`
      - `v4.0.0`, signed with `TEST-3`
    - each release contain a `.sha512sum` file, and its signature in `.sha512sum.sig`
- Download *every file* of *every release* (so for each release: the AppImage, the shasum, and the shasum sig)
- Check shasums:

```bash
$ sha512sum -c *.sha512sum                                                                       
ledger-live-desktop-2.0.0-linux-x86_64.AppImage: OK
ledger-live-desktop-3.0.0-linux-x86_64.AppImage: OK
ledger-live-desktop-4.0.0-linux-x86_64.AppImage: OK
```
- Check signatures:

```bash
$ gpg --verify ledger-live-desktop-2.0.0.sha512sum.sig
gpg: assuming signed data in 'ledger-live-desktop-2.0.0.sha512sum'
gpg: Signature made Wed 12 Dec 2018 05:46:46 PM CET
gpg:                using RSA key E96708A0247A58A87E3EBAE6626D563B22D51F51
gpg:                issuer "ll-test-1@ledger.com"
gpg: Good signature from "LL-TEST-1 (Ledger Live Test Key 1) <ll-test-1@ledger.com>" [ultimate]
```
(do same for other `.sig` files, and confirm that `v3.0.0` is signed with `TEST-2`, etc.)

When all is alright, we need to ensure the app is doing the same. For this, two env var to setup: `UPDATE_CHECK_FEED` and `UPDATE_CHECK_IGNORE`.

```bash
$ UPDATE_CHECK_FEED=https://ledger-live.sigsev.io \
  UPDATE_IGNORE_CHECK=0 \
  ./ledger-live-desktop-2.0.0-linux-x86_64.AppImage
```

expected result:

2018-12-12_517x102
then:
2018-12-12_527x111

Let's try with incorrect signature:

```bash
$ UPDATE_CHECK_FEED=https://ledger-live.sigsev.io/WRONG-SIG \
  UPDATE_IGNORE_CHECK=0 \
  ./ledger-live-desktop-2.0.0-linux-x86_64.AppImage
```

expected result:

2018-12-12_597x126

Additionally, there is a bunch of unit tests, that are ran in the `yarn test` command.


meriadec:
as usual, CircleCI is randomly ignoring some PR. Let's amend the last commit and first push it, because it makes total sense.

meriadec:
res

updated style of Banner.

khalilbenihoud:
Should we reword the different states @dasilvarosa ?

gre:
This is super cool. I'll test asap

dasilvarosa:
I'd like to suggest an adjustment to the states:

* Update available [What's new?] [Install]
* Download update... 49% completed
* Installing update...
* Update failed, please try again [Retry]
* Restart to finish update [Restart]

The first step would need a button [What's new?] that pulls up the changes in the latest release. The user can then decide whether to update. Clicking [Install] launches the download, after which verification and installation are automatically started. Users don't need to know about the awesome and secure verification you guys have built, and I expect this step to be very short so no need to have the state flash shortly. (sorry for the Donjon icon). Then for the final part, I think the user needs to control when the app will restart to finalize the update, otherwise, it would be quite disturbing as we don't announce it anywhere.

gre:
let's discuss internally but i don't think we can change the state semantic. this is not a "please try again" it's actually you need to download again LL because something is compromised afaik.

meriadec:
Additionally, we can't change the behavior of electron-updater: an update is **only** installed during a "quit" process, we can't "hot install" it.

PR #1707, opened 2018-12-13
Insert variable currencyName in error message

dasilvarosa:
Variable got erroneously replaced by Bitcoin.

### Type

Wording fix

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -137,7 +137,7 @@
   "currentAddress": {
     "title": "Current address",
     "for": "Address for account <1><0>{{accountName}}</0></1>",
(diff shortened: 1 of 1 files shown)


dasilvarosa:
I'm testing... HODL

dasilvarosa:
All good.

image

meriadec:
omg is it for real

dasilvarosa:
Yeah, I copy-pasted a wording from mobile project to align but without correcting it. :sorry:

meriadec:
Ah ok :sweat_smile: so it's not here from the beginning !
I think we should also use variable for this, right?

PR #1731, opened 2019-01-03
Support for XSN and activate Sapling for Komodo, fix fees (negative or high one)

KhalilBellakrid:
Integrations: Everything is in the title ! 

Bug Fix: This PR fix also issues related to fees (negative or high one)

!!! WARNING !!! 
Before merging this PR, KMD and Stakenet Nano Apps (Sapling one for KMD) should be available from manager side. 

### Type

Features + Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LLC-113
https://ledgerhq.atlassian.net/browse/LLC-114

Bug Fix: because we activated lazy mode for transaction construction, we were using an existing method that was never called before and was constructing txs from local DB, but unfortunately the queried outputs for this tx were doubled when we have txs between accounts belonging to same user.

Bitcoin Like currencies: send/receive flow,
Stakenet: everything since it's a new support !

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,13 +40,8 @@
     "@ledgerhq/hw-app-xrp": "^4.32.0",
     "@ledgerhq/hw-transport": "^4.32.0",
     "@ledgerhq/hw-transport-node-hid": "^4.32.0",
-<<<<<<< HEAD
-    "@ledgerhq/ledger-core": "2.0.0-rc.15",
-    "@ledgerhq/live-common": "4.13.0",
-=======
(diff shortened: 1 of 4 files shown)


gre:
@Arnaud97234 can you test this?
affects all Bitcoin like so we need to test them. Especially DCR, XSN, KMD, ZEN, ZEC,

gre:
@KhalulBellakrid does this PR also fixes the fee glitch? If yes please update the pr description/title so we document it. Thx

KhalilBellakrid:
@gre I updated the PR description

PR #1750, opened 2019-01-17
<!-- Description of what the PR does go here... screenshot might be good if appropriate -->

valpinkman:


gre:
@Arnaud97234 that covers one more case for the Firmware Update & also add one rare error if you install an app not being on dashboard (like mobile, except here most of the time, we redirect to home of manager anyway)

gre:
I don't think it's related to the changes done by this PR

PR #1754, opened 2019-01-18
repair mcu not genuine

valpinkman:


src/commands/firmwareRepair.js
```diff
@@ -3,12 +3,15 @@
...
+type Input = {
+  version: ?string,
+}
+
 type Result = { progress: number }
 
 const cmd: Command<Input, Result> = createCommand(
   'firmwareRepair',
-  () => repair(''), // devicePath='' HACK to not depend on a devicePath because it's dynamic
+  ({ version }) => repair(''), // devicePath='' HACK to not depend on a devicePath because it's dynamic
```
gre: need to pass second param version

PR #1755, opened 2019-01-20
 Temporary fail-safe for 0x6804 locked derivation path scan on etc

juan-cortes:
Due to the derivation lock introduced in the latest firmware update, the current implementation of the account scanner for the Ethereum bridge causes the scan to crash and the inability to add accounts. While this pr negates the portion of users that still use legacy derivations (they won't be able to find them until the changes are reverted) it allows the rest of the users to add their accounts.

Considering a share of users will have the updated firmware at this point we might want to issue a notice (think changelog, or banner 🤷‍♂️) explaining that if they can't see their account, and it was a legacy account, it's due to this. Or we could delay this until the next ~firmware~ eth app update but perhaps that affects more users.

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -317,27 +317,36 @@ const EthereumBridge: WalletBridge<Transaction> = {
               const freshAddressPath = runDerivationScheme(derivationScheme, currency, {
                 account: index,
               })
-              const res = await getAddressCommand
-                .send({ currencyId: currency.id, devicePath: deviceId, path: freshAddressPath })
-                .toPromise()
-              const r = await stepAddress(
-                index,
-                res,
-                derivationMode,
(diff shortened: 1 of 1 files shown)


gre:
Hopefully we'll have a new ETH app that fixes this

gre:
I think sometimes it's preferable to have a crash rather than silenting the problem, it's obviously a regression on the app side and we need to continue scanning the legacy paths so all users discover the paths.

It would ultimately be better if the interface would allow more choice but current LL UX choice is to hide this from user so it's the simplest possible.

gre:
If people urgently need to retrieve ETC accounts that are on the etc path (44'/61'/..) they can use this PR version to restore it.
We will have a fix of ETC app before the next ledger live release will happen and it's also the proper fix to wait (so etc accounts on eth can be imported again).
So we'll skip this failsafe for now.

PR #1756, opened 2019-01-21
adds UnknownMCU error text

valpinkman:
@dasilvarosa you might want to check out the text used when we cannot find the MCU during the update process.

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -902,6 +902,10 @@
     },
     "FeeNotLoaded": {
       "title": "Couldn’t load fee rates, please set manual fees"
+    },
+    "UnknownMCU": {
+      "title": "Oops, MCU not found",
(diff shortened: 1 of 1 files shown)


static/i18n/en/app.json
```diff
@@ -902,6 +902,10 @@
     },
     "FeeNotLoaded": {
       "title": "Couldn’t load fee rates, please set manual fees"
+    },
+    "UnknownMCU": {
+      "title": "Oops, MCU not found",
```
dasilvarosa: Unknown MCU version

static/i18n/en/app.json
```diff
@@ -902,6 +902,10 @@
...
+      "description": "Please try again or contact us"
```
dasilvarosa: Please contact Ledger Support

PR #1778, opened 2019-01-31
LL-939 Updated the export description text

juan-cortes:
image


dasilvarosa:
Could you make sure the tab and the LiveQR tool have their own keys? Now one key **Export accounts** is used.

PR #1780, opened 2019-02-01
Use the freshResetAll libcore’s function when resetting the app.

phaazon:
**Be careful: there are two important points:**
  - **The QA should test a lot because there is a lot of breaking changes and new features. Please heavily test.**
 - **Please inform clients about resetting their applications because of these breaking changes; these changes are for preparing Ethereum ERC20 integration.**

Also, bump libcore’s version to 2.5.0 (release-candidate).

> Please tell me how you want to handle release candidates. Your current code is currently using the 1.9.0 version IIRC (ping @KhalilBellakrid).

The change was integrated into the following screen: 

screenshot 2019-02-01 at 15 20 35

I removed the `rimraf` stuff as long as the test you were used to do because Live shouldn’t worry about libcore’s data — if you still think you should have a way to check for that, maybe we need to add something in libcore, but I think it’s error-prone that Live make that check.

Also, thanks to @juan-cortes for his wise directions and indications about where to change code to integrate the change about resetting Live! :) 

Please, feel free to comment and open reviews for fixing anything you might dislike.

### DIFF
--- a/src/commands/libcoreReset.js
+++ b/src/commands/libcoreReset.js
@@ -1,15 +1,15 @@
 // @flow
 
 import { createCommand, Command } from 'helpers/ipc'
-import { of } from 'rxjs'
+import { from } from 'rxjs'
 import withLibcore from 'helpers/withLibcore'
 
 type Input = void
 type Result = boolean
 
-const cmd: Command<Input, Result> = createCommand('libcoreReset', () => {
-  withLibcore(core => core.getPoolInstance().freshResetAll())
-  return of(true)
-})
+const cmd: Command<Input, Result> = createCommand(
+  'libcoreReset',
+  () => from(withLibcore(core => core.getPoolInstance().freshResetAll()))
+)
 
 export default cmd
--- a/src/commands/libcoreReset.js
+++ b/src/commands/libcoreReset.js
@@ -7,9 +7,8 @@ import withLibcore from 'helpers/withLibcore'
 type Input = void
 type Result = boolean
 
-const cmd: Command<Input, Result> = createCommand(
-  'libcoreReset',
-  () => from(withLibcore(core => core.getPoolInstance().freshResetAll()))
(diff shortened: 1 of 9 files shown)


juan-cortes:
It should also be called on "clear all", on that same screen the "Clear" cta, good job!

phaazon:
@juan-cortes AFAIK, this should already be the case since the clear calls the `softReset` function. ping @gre

gre:
thanks for the detail explanation. especially for the QA instructions 🚀

gre:
we will not ship that in this release. will be postpone to next sprint.

phaazon:
@gre

gre:
haha! yeah also we need to write the "cleanup logs files" at the same time I feel, just to be sure we do at least one migration pass to clean the legacy files ;) since the clear don't do that anymore.

phaazon:
Yeah, what a journey. :D

phaazon:
@Arnaud97234 you should be good to test now. Please ping me if something is still wrong. I feel a bit uncomfortable with that `withLibcore` not being awaited. Maybe commands already handle that but 🤷‍♂️ (ping @gre).

src/commands/libcoreReset.js
```diff
@@ -0,0 +1,15 @@
...
+
+import { createCommand, Command } from 'helpers/ipc'
+import { of } from 'rxjs'
+import withLibcore from 'helpers/withLibcore'
+
+type Input = void
+type Result = boolean
+
+const cmd: Command<Input, Result> = createCommand('libcoreReset', () => {
+  withLibcore(core => core.getPoolInstance().freshResetAll())
```
gre: return `fromPromise(withLibcore(...`

gre: safer as you commented ;)

phaazon: Cool!

phaazon:
This PR…

PR #1785, opened 2019-02-04
Handle source===destination on select recipient for xrp

juan-cortes:
Similar to https://github.com/LedgerHQ/ledger-live-mobile/pull/740  it depends on https://github.com/LedgerHQ/ledgerjs/pull/285 being merged and ready so I can use a translated error instead of a generic key.
There are some differences in the desktop implementation of the validation since it doesn't throw an error from but rather return a boolean, then a second method to specify the warning (if I got it right), at any rate, this would be the equivalent.

image

### DIFF
--- a/src/components/modals/Send/fields/RecipientField.js
+++ b/src/components/modals/Send/fields/RecipientField.js
@@ -53,12 +53,8 @@ class RecipientField<Transaction> extends Component<
     const { account, bridge, transaction } = this.props
     const syncId = ++this.syncId
     const recipient = bridge.getTransactionRecipient(account, transaction)
-    const isValid = await bridge.isRecipientValid(account.currency, recipient, account.freshAddress)
-    const warning = await bridge.getRecipientWarning(
-      account.currency,
-      recipient,
-      account.freshAddress,
-    )
+    const isValid = await bridge.isRecipientValid(account, recipient)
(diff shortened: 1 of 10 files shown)


src/components/modals/Send/fields/RecipientField.js
```diff
@@ -97,7 +101,7 @@ class RecipientField<Transaction> extends Component<
     const error =
       !value || isValid
         ? QRCodeRefusedReason
-        : new InvalidAddress(null, { currencyName: account.currency.name })
+        : warning || new InvalidAddress(null, { currencyName: account.currency.name })
```
gre: So if I understand correctly we for now consider warning as error as workaround for the fact we don't have yet the better bridge we did on mobile? That's a smart workaround, we'll refactor when sharing more code soon for bridge.


There are legit warning like in case of ethereum when the address is in lowercase and it displays as yellow if Im correct. Do you confirm this doesn't break this case?

juan-cortes: Ethereum warning for lowercase (the lacking of a checksum) shows in yellow since it still passes the `isValid`, so there's no error. In the workaround for XRP it will not be valid so it will be in red, 

image

image

As always I'll feel happier after arnaud goes over it :trollface:

Issue #1751, opened 2019-01-17
Ability to send to one of your own accounts (E.G. send from spending to saving)


Luuk2016:
- [x] I have checked this feature was not yet requested.

Sometimes it can be handy to transfer to own of your own other Ethereum addresses.
It would be nice to select one of the addresses at the send tab.

**Example:**
A person has a BTC account he uses as daily spending account.
Sometimes he will send some assets to his 2nd accounts (savings).

PR #1786, opened 2019-02-04
Send to your own accounts

meriadec:
Basically auto-fill the `transaction.recipient` with `account.freshAddress`.

account

nb: When coming back to this amount step after having selected an account in the list, (e.g if you press back from device step) the address will appear as text (the account is not selected back). IMO it would not be worth / useful to do it (can explain later).


gre:
relevant feature request addressed #1751

gre:
very cool!

static/i18n/en/app.json
```diff
@@ -316,6 +316,9 @@
         "title": "Details",
         "selectAccountDebit": "Account to debit",
         "recipientAddress": "Recipient address",
+        "recipientAccount": "Account to credit",
+        "useAccount": "Use one of your accounts",
```
dasilvarosa: Enter own account

meriadec:
brace yourself, I amend & push force *just to trigger CI build* :1st_place_medal:

meriadec:
Updated the `Select` component to handle `error` and `warning` props the same way `Input` does. A shame that error message doesn't respect vertical rythm, and so it's not aligned with natural text placed after. Could be polished later, not the purpose of this PR.

**Error style** (here for XRP):

2019-02-06_22-06-29_935x396

**Warning style**

2019-02-06_22-08-08_937x209

gre:
thoughts: I wonder if we have a way to validate on the device that the account belong to you, like during a transaction if we provide the path it belongs too it could say "it's yours" but i doubt anything is done for that as it seems a edgecase (a bit like for the change address)

meriadec:
An attacker could always change address at the very end, when transaction is built, so the only security remains verifying address on the device. Again, that's the equivalent of doing the receive flow "without device" and pasting the address, and today we have no check on that on the send first step.

gre:
one way would be to have to do two things in a row on the device, you verify address and then you remember it and then you verify the destination. worse UX.
so it will be a security discussion to have.

meriadec:
An attacker could take control of your clipboard, so pasting an address has the exact same level of security than selecting an account in a list. The PR address convenience problem, the security is handled on the next steps of the Send flow.

gre:
but in the normal address case you verify the addressed display on the device with where you send funds, and if possible in a 2FA way (like on another screen / phone).
with the account selection, you have no way to verify.

meriadec:
I don't get how it's not a "normal address", it is the equivalent of doing `Receive modal => Don't have my device => [copy] => [close] => Send modal => [paste]`.

And you literally verify the address on the "verify" step of the Send flow. If you validate a wrong address on your device, yes I guess SFYL, but I don't see how this feature lower at any point the security.

Again, attacker would just change [the address here](https://github.com/LedgerHQ/ledger-live-desktop/blob/ecef03d656d847e3d507cdc14fc2bc0a40de4f3c/src/commands/libcoreSignAndBroadcast.js#L235) so it bypass everything (except the device check).

gre:
You are right that it's not less secure than current approach **for sending from one account to another**.

I was only comparing the use-case of sending to somewhere else (like a friend, or an exchange), in that case you will be more secure because you have a way to verify properly.


`Receive modal => Don't have my device => [copy] => [close] => Send modal => [paste]` is annoying experience but at least you have the display on the device that happens. I'm not fan of adding a feature that helps the user doing that but in a less secure way. It's like we say to the user "you can directly move your funds, don't worry" but it's not a secure path.

please let's continue discussion internally ;) next live-security meetup

gre:
So I think we could have a yellow warning that always appear (if no error) and that says something like

`Make sure to have verified on device the destination account address. You will need to verify the transaction with this address.`

not sure how to make it more concise but it's fundamentally what we need to tell the user.

gre:
the security/verification/legal consequence is too tricky to have this feature at this point.

PR #1787, opened 2019-02-05
LL-832 skip 'restart in bootloader' if device is already in bootloader

valpinkman:


src/components/SettingsPage/RepairDeviceButton.js
```diff
@@ -45,7 +51,7 @@ class RepairDeviceButton extends PureComponent<Props, State> {
   repair = (version = null) => {
     if (this.state.isLoading) return
     const { push } = this.props
-    this.setState({ isLoading: true })
+    this.timeout = setTimeout(() => this.setState({ isLoading: true }), 500)
```
gre: it's a very edge case scenario but you should also clearTimeout in the complete case

gre: as well as in close() case.

valpinkman: ok i'll do that

PR #1788, opened 2019-02-05
Add NIX Support

mattt21:
NIX is added to the list of supported cryptocurrencies.

### Type
Feature

### Context
Since all the support for nix in `ledger-live-common` is complete and `lib-ledger-core` is pending(https://github.com/LedgerHQ/lib-ledger-core/pull/94). With lib-ledger-core pulled, it will be safe to add nix to the list of supported cryptocurrencies

No current behavior in the app has been modified since only the bitcoin-like behavior is complete, nix behaves exactly as bitcoin.

### DIFF
--- a/src/config/cryptocurrencies.js
+++ b/src/config/cryptocurrencies.js
@@ -30,6 +30,7 @@ const supported: CryptoCurrencyIds[] = [
   'clubcoin',
   'decred',
   'bitcoin_testnet',
+  'nix',
 ]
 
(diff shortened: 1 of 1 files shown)


gre:
we'll reopen when the PR is fully ready (infra, libcore). this require a business decision from Ledger side. Feel free to contact @FabriceDautriat .

tomasgulas:
@FabriceDautriat when can we expect NIX in Ledger Live app? It's the only coin I'm missing there ATM.

PR #1791, opened 2019-02-06
LL-931 Add a banner to promote the mobile version of the Live + Banner refactoring

juan-cortes:
Depends on https://github.com/LedgerHQ/ledger-live-desktop/pull/1705 being merged and leverages on the awesome work once again of @meriadec so hopefully this is a step forward on his work and not backward:trollface:  So I extracted the concept of the top banner from being just used for the secure updated to being something more generic that we can use for things like this story.

> Refer to the [last commit in here](https://github.com/LedgerHQ/ledger-live-desktop/commit/b7b81ad420176c984d43729ab6fc0c8cb500a447) for my changes, all the rest is not me.

image

I've integrated it with the `dismissedBanners`  functionality that we are already taking care of so we have an easy way of making the banner depend on its ID for visibility and making it dismissable. Also, since @gre wanted that we only show one banner at a time I'm only rendering the first element, maybe we want to extract that from the style into a setting somewhere, I don't know.

~Design probably has something to say about my choice of dismiss button/colors so feel free to weigh in on that @dekkaki @KhalilBellakrid @bscholving~ Updated with the design from @KhalilBellakrid  thanks it looks great!.

Hopefully, this doesn't break the secure update awesome banner feedback, I did my best to test it (hardcoding the values that would come from the context) and it seems to be just fine, but might want to check that further if you have the setup ready.

screenshot 2019-02-06 at 01 05 55


gre:
adding HODL as we need to wait #1705

juan-cortes:
Updated to follow the design

meriadec:
nice :sunglasses: as said earlier, let's approve until 1705 is merged in develop then develop merged back in this one :)

khalilbenihoud:
Nice job @juan-cortes , as discussed earlier this week, the banner should stick to the top, can you take a look?

juan-cortes:
> Nice job @juan-cortes , as discussed earlier this week, the banner should stick to the top, can you take a look?
Once the parent pr is merged and we have this one updated I'll see if I can do what we talked about

juan-cortes:
Latest commit comments the `await` of the updater (and some other stuff so ci doesn't complain) and makes the separator line that was part of the topbar an independent component that will only be visible if there are no visible banners. 

- If there is an updater banner, it will hide the separator and any promotional banner.
- If there is only a promotional banner and we dismiss it, it will show the separator bar.
- If there is nothing to show, the separator bar will still be visible for now

image

I'm hesitant to resolve the conflict for the release notes, so if @gre or @meriadec don't mind solving it I appreciate it.

juan-cortes:
@meriadec thank you

meriadec:
@juan-cortes done :heavy_check_mark: , in fact only the `centered` prop [was causing conflicts](https://github.com/LedgerHQ/ledger-live-desktop/pull/1791/commits/4abf811706969d2b2ff82b6dea715ab860eea469) ^^

khalilbenihoud:
All good from my side!

PR #1794, opened 2019-02-06
Fixes LL-807

gre:
The previous code was actually incorrect, we also proved the fix works on mobile, as we already have that code from the start: https://github.com/LedgerHQ/ledger-live-mobile/blob/ebc9f9af8a519afa56911f0c1d888b349eb545af/src/libcore/signAndBroadcast.js#L143

**Technically, it should finally fixes some ZEC errors that shows as "getVarint error".**

### Type

bugfixes

### Context

LL-807

everything still works. emphasis on the sign and broadcast of transactions for BTC & derivates.

### DIFF
--- a/src/commands/libcoreSignAndBroadcast.js
+++ b/src/commands/libcoreSignAndBroadcast.js
@@ -129,7 +129,7 @@ async function signTransaction({
       const hexPreviousTransaction = Buffer.from(rawPreviousTransaction).toString('hex')
       const previousTransaction = hwApp.splitTransaction(
         hexPreviousTransaction,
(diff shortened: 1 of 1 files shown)


meriadec:
AFAR it was actually intended to force to `true`. It has already been set to `currency.supportsSegwit` (see https://github.com/LedgerHQ/ledger-live-desktop/commit/2f4299b4cca682da945ea77cad42955367f6d93d) and was updated to `true`. I would :warning: to change without hardly test

gre:
yeah but on the other end @KhalilBellakrid  tell me this creates the actual bug of ZEC (that is not segwit).

Issue #905, opened 2018-07-07
Send: add MAX on amount to set the account balance


gre:
#### Part of the application

Send > step1

#### Description

Allow to use the Maximum available balance in the amount field (Send>step1).

gre:
we can't implement it easily in 1.1.0, will have to postpone it for later.

one of the reason is, our Ethereum api returns json and uses JSNumber for the balance field. We can't rely on JavaScript to be precise enough to have this field right.
We'll need to do #1110 and then, either we would have to implement our own JSON parser (that i would not recommend taking that path 🐰🕳), or we'll wait libcore to support Ethereum (not sure about ETA on this @pollastri-pierre  ?)

gre:
The question is do we want to have it :100: reliable or do we want an escape hatch that will leave some micro amount in accounts you left / sometimes will actually fail (before number imprecision).
I think we prefer option #1 and ship a bug-free experience.

gre:
status update: 

BigNumber is here, and libcore have a built in way to "empty everything" so we'll give a try to this.
We should be able to reliably implement MAX for all coins except Ethereum and Ethereum Classic, because these still have imprecision due to our API, but as soon as it gets backed by libcore, we should be able to do it as well.

I suggest we ship it as experimental (with a env variable) for all libcore backed coin for now and turn it on only when it's available everywhere.

eliasjnior:
Any progress? I always use the old Ledger Wallet Bitcoin to transfer full amount, because I don't know what size will be the transaction to calculate the network fee and discount from my balance. Thanks :)

gre:
Probably it expect around Q1 2019. Relatively easy to add but we're waiting on shipping Ethereum with libcore to have it reliable as said above

nilskp:
Please make sure to make the fee adjustable for this scenario also. I.e. send MAX _given_ an entered fee.

gre:
Yeah I think MAX toggles a mode that locks the amount field and fee remains "free". That's how I would imagine it but we'll discuss it internally and have good ux for this.

AndydeCleyre:
Is it now possible to do this, with some env var set, as suggested? If not, is it targeted for a particular release milestone? My case is not using eth, if that matters.

alexkyse1:
Any update regarding this feature? It should be at MVP from the start imo.

AcidRaZor:
I regret buying Nano Ledger S now

gre:
[...]

PR #1798, opened 2019-02-07
Basic send max integration

meriadec:
relatable: #905

**Disclaimer**: just connecting the UI.

Aim of this PR is to provide a working solution for the UI, not intrusive, obvious to use. No need weird/complex behavior for such a simple action.

> "- And what if the user click on max then increase the amount?
> "- Then he can't send.
> "- Ah ye

sendmax

:warning: Still needs to use libcore instead of naive search of max :)
:warning: When #1786 get merged, use the same "link style" for its action. More obviously interactable.

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -464,8 +464,8 @@ const EthereumBridge: WalletBridge<Transaction> = {
       ? Promise.resolve(t.amount.plus(t.gasPrice.times(t.gasLimit)))
       : Promise.resolve(BigNumber(0)),
 
-  getMaxAmount: (a, t) =>
-    Promise.resolve(a.balance.minus((t.gasPrice || BigNumber(0)).times(t.gasLimit))),
+  // getMaxAmount: (a, t) =>
+  //   Promise.resolve(a.balance.minus((t.gasPrice || BigNumber(0)).times(t.gasLimit))),
 
   signAndBroadcast: (a, t, deviceId) =>
     Observable.create(o => {
--- a/src/bridge/LibcoreBridge.js
+++ b/src/bridge/LibcoreBridge.js
@@ -46,6 +46,8 @@ const EditFees = ({ account, onChange, value }: EditProps<Transaction>) => (
   />
 )
 
+const editTransactionAmount = (account, t, amount) => ({ ...t, amount })
+
 /*
 import AdvancedOptionsBitcoinKind from 'components/AdvancedOptions/BitcoinKind'
 const EditAdvancedOptions = ({ onChange, value }: EditProps<Transaction>) => (
@@ -183,10 +185,7 @@ const LibcoreBridge: WalletBridge<Transaction> = {
     isRBF: false,
   }),
 
-  editTransactionAmount: (account, t, amount) => ({
-    ...t,
-    amount,
-  }),
+  editTransactionAmount,
 
   getTransactionAmount: (a, t) => t.amount,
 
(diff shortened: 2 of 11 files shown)


gre:
I think MAX needs to toggle a mode where the funds are locked. Maybe the inputs even says "max" in disabled style.
So there are some internal discussion to have in this but it's fundamental at least for mobile that it's this mode because you can change the fees afterwards

meriadec:
Was expecting a bit the answer ;) On the other hand, I think MAX should *just* be a facility to input a value (like selecting a recipient among your accounts is a facility to input recipient field). It should not follow hidden / magic rules that disable fields, and try to be smarter than user. The fact we shoke on that is the reason this feature is still not in the app since > 7 months while one of the most requested.

I can understand the mobile concern but honestly it's edge case, no? Like said in the PR desc, what if user increase the fees (are we sure it's so frequent?) ? Then he will get informed by an error message, and have to click again on send max. Not **too horrible** (it's a click, after all), and logical & understandable. Also, this is desktop PR.

Personally I always used send max in this way (e.g Kraken) and never wondered about a "send max mode".

I propose "make it work, then make it better": so we try the minimalist / not machinery / actually asked solution, then if we see *users* are willing to lock the input while clicking max, we refacto :) referendum style

gre:
> It should not follow hidden / magic rules that disable fields, and try to be smarter than user.

I disagree, I think it should. we should really show to user that you send ALL. especially because under the hood it's totally two different implementation (wipeAll). so it's better to show the reality.

It's also more convenient in term of UX: you do a MAX then you lower the fee, in your proposal you have no way to show to the user that you will actually no longer "send max" because you have remaining funds that correspond to the fee difference you changed. SFYL.

> The fact we shoke on that is the reason this feature is still not in the app since > 7 months while one of the most requested.

this is not the real reason. The real reason I blocked that feature is because Ethereum API don't allow us to reliably implement the feature (we don't have a string version of the amount so we can't map it to bignumber precision).
It was commented here: https://github.com/LedgerHQ/ledger-live-desktop/issues/905#issuecomment-404835524 so i'm the main reason this was blocked, no one else to blame ;)

I don't think at all mobile is an edge case, i'm pretty sure most of our users change the fees.

To summary: I think UI should tell the user that you send it all due to the fact that "sending MAX" is a totally different usecase than doing a normal transaction. In fact it could even be done somewhere else, not using "Send" but using "Wipe All" but obviously it's more work, so I think something between the two is good. The summary screen should tell the the user next to the amount maybe that it's `(MAX)`

gre:
> I propose "make it work, then make it better": so we try the minimalist / not machinery / actually asked solution, then if we see _users_ are willing to lock the input while clicking max, we refacto :) referendum style

I prefer not to implement hacks and having many users leaving some dust each time. A feature must be reliable, otherwise it's a legacy you add to your pile of bugs.

gre:
I understand your point of view tho, in fact for Desktop it could be ok to do this "set max", i'm just annoyed by the "and then you change the fees" behavior.
But since for Mobile we can't (without big changes) do the same approach is a dead end to me and we should have same approach for Mobile & Desktop as much as possible.

meriadec:
And I totally understand your point too :+1: But maybe we can deliver, then iterate based on feedbacks.

> the real blocker here is ETH

As currently proposed, the send max will display only if [the bridge has the method](https://github.com/LedgerHQ/ledger-live-desktop/pull/1798/files#diff-5066aa9f01d1ae48e9ced8f2d202e6dbR58), so it's incremental adoption. ETH like is only 2 coins among 26.

> I prefer not to implement hacks

come one :smile: once we use [libcore's `wipeToAddress`](https://github.com/LedgerHQ/lib-ledger-core-node-bindings/blob/d1d715a17bab0604d4e5c9b16eade56d46c5a308/src/ledgercore_doc.js#L1536) it could be kind of clean

> we should have same approach for Mobile & Desktop as much as possible

Good point, but hard to achieve since amount step is splitted in steps in mobile.

SimonHausdorf:
How is that feature still not a thing or experimental function? Please, give us this feature in you next release at least as an experimental function.

gre:
yes this will be an experimental in the next release in fact.

SimonHausdorf:
> yes this will be an experimental in the next release in fact.

it's still not there?

gre:
it's coming in #2164

PR #1800, opened 2019-02-08
Improve update wording

dasilvarosa:
This should address some UX problems during the update.

**Note at implementing dev:** 
- [ ] Put `Processing` in bold style in the _mcuSecond_ key.


gre:
@juan-cortes  if you can give a second 👁  to this, but there is no structural change (like new arguments in wording) so it should be smooth and we can merge it

meriadec:
@dasilvarosa can you amend your commit to use spaces instead of tabs? :pray:

dasilvarosa:
@meriadec Done. Thanks!

dasilvarosa:
Too bad it's already merged without putting the `mcuSecond` key in bold!

PR #1808, opened 2019-02-13
LL-1006 bump live-common to benefit from the refactoring 

juan-cortes:
Depends on [LedgerHQ/ledger-live-common#149](https://github.com/LedgerHQ/ledger-live-common/pull/149) being merged and a version bump pending too. This completes the refactoring of the partners section out into the wild of live-common.

> **I had to add that part of live-common to the untyped list on `.flowconfig`, I can't understand why it's complaining when running flow.**


gre:
@juan-cortes can you fix the conflict?

juan-cortes:
Done

PR #1809, opened 2019-02-13
Bump ledgerjs libs

valpinkman:
Update some core ledgerjs libraries

### Type

Bump


gre:
that PR ports the LedgerJS new transport impl. there is a bump of node-hid in it , but it should work the same. maybe it will help fixing some usb problem but I highly doubt it. we'll need to see what better can be done than `usb` events

valpinkman:
@Arnaud97234 these issues have been fixes with the latest update of live-common

needs testing tho

src/components/ManagerPage/FirmwareUpdate.js
```diff
@@ -81,18 +81,19 @@ class FirmwareUpdate extends PureComponent<Props, State> {
...
     const { firmware, modal, stepId, ready } = this.state
+
+    const deviceSpecs = getDeviceModel(device.modelId)
+
     return (
       <Card p={4}>
         <Box horizontal align="center" flow={2}>
           <Box color="dark">
-            {device.product === 'Blue' ? <Blue size={30} /> : <NanoS size={30} />}
+            {deviceSpecs.id === 'blue' ? <Blue size={30} /> : <NanoS size={30} />}
```
gre: note for future refactoring: we should have a component map like this: `{ blue: Blue, nanoS: NanoS }`

valpinkman: Yup I was thinking about it as well. Will come up with something like this when implementing Nano X designs

PR #1814, opened 2019-02-14
Adds the developer mode to exported data

juan-cortes:
Depends on https://github.com/LedgerHQ/ledger-live-common/pull/150 being merged and a version bump, but this allows the developer mode setting to be imported on mobile (no change needed on that end).

### DIFF
--- a/src/reducers/settings.js
+++ b/src/reducers/settings.js
@@ -244,10 +244,12 @@ export const exportSettingsSelector = createSelector(
   counterValueCurrencySelector,
   counterValueExchangeSelector,
   state => state.settings.currenciesSettings,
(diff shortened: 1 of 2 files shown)


gre:
Ultimately it does not strongly depends on the merge , it works without it no? Seems like we have a loosely type in this end.
I guess it's more on mobile side that type will be used? Can't remember 😅

gre:
No change needed on mobile? Interesting

juan-cortes:
I tried straight away by doing the common update. On mobile we import all settings, so if it's mapped with the correct key, any setting we pass will be updated hence why on desktop it's `developerMode` and on mobile `developerModeEnabled` (We could actually make them the same eventually)

gre:
makes sense

PR #1815, opened 2019-02-19
Onboarding LL with a new Nano X

valpinkman:
Add steps to set up a brand new Ledger Nano X from onboarding

### Type

Feature

### Context

LL-635 LL-636 LL-637

Onobarding

init
init selected
write pin
restore


src/helpers/devices.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import { getDeviceModel } from '@ledgerhq/devices'
+
+import type { DeviceType } from 'reducers/onboarding'
+
+export const deviceModelName = (id: DeviceType): string => {
```
gre: I would suggest to directly do the `getDeviceModel` code in the UI side, no need for this function I feel (we can directly resolve the device model object on the UI side.

src/reducers/onboarding.js
```diff
@@ -15,6 +15,8 @@ type Step = {
   },
 }
 
+export type DeviceType = 'nanoX' | 'nanoS' | 'blue' | ''
```
gre: is it possible we don't have to treat the `''` case? I imagine we could init the value at `nanoS` ? or is it going to actually select it in the UI?

src/components/Onboarding/steps/SelectPIN/index.js
```diff
@@ -4,17 +4,36 @@ import React from 'react'
...
 import OnboardingFooter from '../../OnboardingFooter'
 import SelectPINnano from './SelectPINnano'
 import SelectPINblue from './SelectPINblue'
+import SelectPINnanoX from './SelectPINnanoX'
 import SelectPINrestoreNano from './SelectPINrestoreNano'
 import SelectPINrestoreBlue from './SelectPINrestoreBlue'
-
 import type { StepProps } from '../..'
 
+const SelectPinSwitcher = ({
```
gre: I would just call this component `SelectPIN` because it's basically the generic version. the "Switcher" is implementation detail so don't need to be in the naming ;)

src/components/Onboarding/steps/SelectPIN/index.js
```diff
@@ -25,20 +44,20 @@ export default (props: StepProps) => {
           category="Onboarding"
           name="Choose PIN"
           flowType={onboarding.flowType}
-          deviceType={onboarding.isLedgerNano ? 'Nano S' : 'Blue'}
+          deviceType={cleanDeviceName(onboarding.deviceType)}
```
gre: i'm a bit confused why we don't pass-in deviceType, which is original a model.id , now you give a `Nano X` string? 🤔

src/reducers/onboarding.js
valpinkman: it will select it in the UI?

src/components/Onboarding/steps/SelectPIN/index.js
valpinkman: Ok

src/components/Onboarding/index.js
```diff
@@ -19,7 +19,7 @@ import {
   prevStep,
   jumpStep,
   updateGenuineCheck,
-  isLedgerNano,
+  deviceType,
```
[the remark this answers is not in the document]
valpinkman: ok will refacto

gre:
this is the `NanoXX` :trollface:

valpinkman:
@Arnaud97234 waiting for some new illustrations

valpinkman:
> this is the `NanoXX` :trollface:

`Nano X+`

src/components/Onboarding/steps/Analytics.js
```diff
@@ -67,13 +69,15 @@ class Analytics extends PureComponent<StepProps, State> {
...
+    const model = getDeviceModel(onboarding.deviceModelId)
+
     return (
       <FixedTopContainer>
         <TrackPage
           category="Onboarding"
           name="Analytics"
           flowType={onboarding.flowType}
-          deviceType={onboarding.isLedgerNano ? 'Nano S' : 'Blue'}
+          deviceType={model ? model.productName : ''}
```
gre: we don't need to do `model ?` because `getDeviceModel` have the guarantee to return a `DeviceModel` if i'm correct

src/components/Onboarding/steps/Analytics.js
```diff
@@ -67,13 +69,15 @@ class Analytics extends PureComponent<StepProps, State> {
     const { nextStep, t, onboarding } = this.props
     const { analyticsToggle, sentryLogsToggle } = this.state
 
+    const model = getDeviceModel(onboarding.deviceModelId)
```
gre: which creates a type problem `getDeviceModel` assume you give an id that is not `""` so either we need to remove this case, or have an invariant? at this step we should have it filled but it's hard to express it as a type I imagine?

PR #1819, opened 2019-02-21
LL-1063 Account for terminated cryptos

juan-cortes:
> Using **Bitcoin** as a terminated cryptocurrency just for testing, I didn't have any accounts set up for the two that we are marking as terminated soon. Don't worry.

image

Depends on https://github.com/LedgerHQ/ledger-live-common/pull/156 being merged and bumping the version for live-common to get the new type for CryptoCurrency. This pr allows us to get a list of terminated cryptocurrencies and adapts the UI to hide/disable functionality when needed.


src/config/cryptocurrencies.js
```diff
@@ -32,8 +32,10 @@ const supported: CryptoCurrencyIds[] = [
   'bitcoin_testnet',
 ]
 
-export const listCryptoCurrencies = memoize((withDevCrypto?: boolean) =>
-  listCC(withDevCrypto)
-    .filter(c => supported.includes(c.id))
-    .sort((a, b) => a.name.localeCompare(b.name)),
+export const listCryptoCurrencies = memoize(
```
gre: can you make sure memoize works with more than one param?

juan-cortes: uhmm damn, I'll check

gre:
@juan-cortes you can bump and publish a new live-common and update it in this PR.
@Arnaud97234 should be able to test the behavior

src/components/ManagerPage/AppsList.js
```diff
@@ -89,7 +91,10 @@ type State = {
 }
 
 const oldAppsInstallDisabled = ['ZenCash', 'Ripple']
-const canHandleInstall = c => !oldAppsInstallDisabled.includes(c.name)
+const terminatedCryptoCurrencies: Array<CryptoCurrency> = listCryptoCurrencies(true, true)
+const canHandleInstall = app =>
+  !oldAppsInstallDisabled.includes(app.name) &&
+  !terminatedCryptoCurrencies.some(coin => coin.managerAppName === app.name)
```
gre: Smart

juan-cortes:
@gre live-common bumped, this pr updated and develop merged into it.
@Arnaud97234 ready for testing, you'll need to have hcash/posw accounts in order to get the warning banners. They should link to the correct articles too.

gre:
Do we see the banner when you also try to use Send and Receive?

One think I've noticed is in the implementation we could have override the currenciesStatusSelector and not touch at all the UI code.
Basically the concat can be moved in the selector?

juan-cortes:
@gre changed the logic to be on the selector as you wanted, and disabled the continue on the send flow if the coin is terminated. @Arnaud97234 When testing you will only see the orange banner since it shows only one of them. But the validation should prevent continue on posw and hcash

image

PR #1839, opened 2019-03-05
LL-1114 Duplicate banners + extras: Repair modal, Reset modal

juan-cortes:
~Without being able to add accounts for `hcash` and `posw` to test this currently (the explorers reject the request) all I could find was this broken syntax for the banner id, perhaps it somehow made the banners act up~

@Arnaud97234  provided me with a data file containing a few of the cancelled accounts, this allowed me to pinpoint that the banners are in fact two locals (terminated accounts) and two remote hsm warnings that happened to have the same texts. By filtering them like this we should only have one, prioritising the HSM one.

Also added some polishes to:
- Repair modal, to allow the list to overflow
- Centered the reset modal from a locked screen
- Added a separator to the accounts page, but this is temporary since the separator should be in the header and not in the content so as to stay visible when scrolling. I will address that on a separate pr.

### DIFF
--- a/src/components/CurrenciesStatusBanner.js
+++ b/src/components/CurrenciesStatusBanner.js
@@ -103,7 +103,7 @@ class CurrenciesStatusBanner extends PureComponent<Props> {
     return (
       <Box flow={2} style={styles.container}>
         {filtered.map(r => (
-          <BannerItem key={`{r.id}_${r.nonce}`} t={t} item={r} onItemDismiss={this.dismiss} />
+          <BannerItem key={`${r.id}_${r.nonce}`} t={t} item={r} onItemDismiss={this.dismiss} />
         ))}
       </Box>
     )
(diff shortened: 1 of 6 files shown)


src/components/CurrenciesStatusBanner.js
```diff
@@ -103,7 +103,7 @@ class CurrenciesStatusBanner extends PureComponent<Props> {
     return (
       <Box flow={2} style={styles.container}>
         {filtered.map(r => (
-          <BannerItem key={`{r.id}_${r.nonce}`} t={t} item={r} onItemDismiss={this.dismiss} />
+          <BannerItem key={`${r.id}_${r.nonce}`} t={t} item={r} onItemDismiss={this.dismiss} />
```
gre: \o/

juan-cortes: but it wasn't it.

juan-cortes:
@gre take a look when you have a minute, it fixes a few regressions and it's a quick win (unless I'm breaking things)

PR #1843, opened 2019-03-07
Added noScroll flag to receiveModal

juan-cortes:
Super small fix but it looked terrible. If we can locate some more misbehaving modals on Desktop I can add the fixes here.
image

### DIFF
--- a/src/components/base/Select/index.js
+++ b/src/components/base/Select/index.js
@@ -87,7 +87,7 @@ class Select extends PureComponent<Props> {
       <ReactSelect
         ref={c => (this.ref = c)}
         value={value}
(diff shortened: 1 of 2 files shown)


valpinkman:
@juan-cortes we have the same issue on the Repair Modal ;)

juan-cortes:
I believe I fixed that already @valpinkman on https://github.com/LedgerHQ/ledger-live-desktop/pull/1839

juan-cortes:
I added a max height of 200 to all dropdowns it won't be perfect (less visible at one glance) but it should prevent that particular eyesore. Tell me if this is better if not, I can revert. @gre wdyt

PR #1853, opened 2019-03-12
implement extra countermeasure to make sure app renders error

gre:
there are obviously 'white screen' error happening. This should make sure we display the error in the future.


gre:
@Arnaud97234 you can basically test that the app still works fine. no need to dig too much, it basically improve the error crashes when we have them (the one that crashes the whole app).

gre:
@Arnaud97234 to test this you can for instance remove the libcore.so file that is resolved by libcore, it should break the internal process at boot

PR #1856, opened 2019-03-13
LL-1128 Made the operation details amount and countervalue selectable

juan-cortes:
image

### How to test?
Operation details should allow the individual selection of both the amount and the countervalue amount.


gre:
4 reviews 😅

PR #1860, opened 2019-03-14
add interruption to un/install apps

valpinkman:
### Type

Improvements

### Context

LL-1101

App install/uninstall

### DIFF
--- a/src/components/ManagerPage/AppsList.js
+++ b/src/components/ManagerPage/AppsList.js
@@ -167,7 +167,9 @@ class AppsList extends PureComponent<Props, State> {
   }
 
   flush = async () => {
-    const { device: { path: deviceId } } = this.props
+    const {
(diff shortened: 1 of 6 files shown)


src/components/ManagerPage/AppsList.js
```diff
@@ -205,7 +209,7 @@ class AppsList extends PureComponent<Props, State> {
           </Box>
         )}
         <Text ff="Museo Sans|Regular" fontSize={6} color="dark">
-          {t(`manager.apps.${mode}`, { app })}
+          {mode !== "home" && t(`manager.apps.${mode}`, { app })}
```
gre: would prefer we use ` ? .. : null`

valpinkman: ok

PR #1862, opened 2019-03-15
Cryptographic verification of updates

MortalKastor:
Follow-up to #1705 with the following

### Changes

- Switch from PGP to OpenSSL to better fit our needs
  - Since OpenSSl doesn't use fingerprints, a hex representation of the sha256 hash of the key is used instead
- Harden finding the update file location
  - This required upgrading `electron-updater` and `electron`
- Fix the updater on macOS to allow preventing a downloaded update from being applied
  - `electron-updater` had to be [forked](https://github.com/LedgerHQ/electron-builder) and [modified accordingly](https://github.com/LedgerHQ/electron-builder/commit/c8028acecc51a92dc935b317a38e5f8ab08f9798)
- Send update verifications errors to Sentry

### Still missing

- [x] The key server
- [x] The qorum pubkey

### Context

#1705

### Test plan

WIP


gre:
(key server will be same as where we put currencies.json)

src/commands/autoUpdate.js
```diff
@@ -16,25 +16,22 @@ type Result = {
...
-        // })
-        // await appUpdater.verify()
+        const appUpdater = await createElectronAppUpdater({ feedURL: UPDATE_CHECK_FEED, info })
+        await appUpdater.verify()
         sendStatus('check-success')
       } catch (err) {
-        // don't throw if the check fail for now. it's a white bullet.
+        logger.critical(err)
+        // don't throw if the check fail for now. it's a blank bullet.
         if (UPDATE_CHECK_IGNORE) {
```
gre: does this mean we'll need to set this as true for the first build ?

MortalKastor: `UPDATE_CHECK_IGNORE` is already set as `true` (since #1705).
We'll need to toggle it to `false` in release n+2, if the update from release n to n+1 didn't raise any issue.

efurrrrrrr7 (commented):
いȠ呧봗滴캒Ꜩ衿㹌扮棄鞹Ⲗ�ꅐ鿼ﺵ쀧ȠᤁƖ팥魼Ͻퟛ庨퇜폪阷菛扺뀪

PR #1869, opened 2019-03-19
Fixes broken Copy component + hover styles

juan-cortes:
youhoverwithstylenowyes
Rows after the second one for operation details were not getting the copy component, also added some styles for hover to highlight the row being copied.


src/components/modals/OperationDetails.js
```diff
@@ -45,8 +45,10 @@ const OpDetailsTitle = styled(Box).attrs({
   letter-spacing: 2px;
 `
 export const Address = styled(Text).attrs({})`
+  margin-left: -4px;
```
gre: /invoke @meriadec  :trollface:

PR #1870, opened 2019-03-20
Bump @ledgerhq/* libs

gre:
bump libraries. better typing.

### Type

bumps

- onboarding device selection is still ok
- access to manager still works
- disconnecting the device in a middle of a genuine check or an app install still stops it, potentially display a better error


gre:
@juan-cortes  that is the problem with the "one types for all" approach that Onboarding has taken. typically it would not be a problem if each step would implement its requirement as a type.

gre:
previous implementation was worse because we were literally using `""` to pass everywhere. this case should not happen anyway but in case it does, it worse case we don't crash now, we fallback on nanoS

PR #1871, opened 2019-03-20
Refine HTTP errors

gre:


gre:
I'm not really sure what can be test btw @Arnaud97234  . one typically thing this PR fixes is glitch like the one showing `JsDefined({` when you send eth with very low gas limit.

PR #1875, opened 2019-03-22
Experimental rates section (wip)

gre:
We're having hard time explaining to users how to properly configure rate providers.. this comes from the fact our settings are not well organized. This feature proposal addresses and also show you what rates are actually used in the application.

(***This is not designed**, it was quickly written in the train in 30mn^^*)

Capture d’écran 2019-03-23 à 09 56 29

### Type

feature

### Context

countervalues settings improvement


gre:
we will rework this soon

PR #1877, opened 2019-03-22
reworked import account error

valpinkman:
### Type

Improvement

### Context

LL-1098

Import Account Modal

image


src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -27,6 +28,40 @@ import DebugAppInfosForCurrency from 'components/DebugAppInfosForCurrency'
...
+  const errorToThrow =
+    name === 'BtcUnmatchedApp' || statusCode === 0x6982
+      ? new DeviceShouldStayInApp(null, { appName })
+      : err
+
+  return errorToThrow
+}
+
+const ImportError = ({ error, currency }: { error: Error, currency: ?CryptoCurrency }) => {
+  invariant(error, 'Trying to render inexisting error')
```
gre: `if (!error) return null;` otherwise this code can badly crashes the app

valpinkman:
@Arnaud97234 done

PR #1885, opened 2019-03-27
fix bitpanda logo and error on width in partner cards

valpinkman:
### Type

Fix

### Context

LL-1167

Partners


gre:
@valpinkman  can you make the CI happy about `[✘] rm: cannot remove 'flow-typed/npm/redux_*': No such file or directory` ? we can just remove that i think... we really should commit flow-typed in the future like we did on mobile.

PR #1886, opened 2019-03-27
fix gltichy message during firmware update

valpinkman:
### Type

fix

### Context

LL-1177

firmware update modals


gre:
@valpinkman can you merge or rebase develop?

valpinkman:
@gre doing it right now

PR #1887, opened 2019-03-27
add search by ticker to manager and add account

valpinkman:
### Type

Improvement

### Context

LL-425

- manager search bar and add account select bar
- potentially all select are impacted because bump of react-select


gre:
@valpinkman prettier

valpinkman:
@Arnaud97234 can you test again ?

gre:
@valpinkman can you squeeze the two commits?

PR #1888, opened 2019-03-28
Use libcore from live-common

gre:
This completely move out the libcore business logic and use the shared libcore logic of live-common. 

**Here is the rough roadmap:**

1. ✅ ~~migrate mobile code to a unified API and platform independent codebase.~~ https://github.com/LedgerHQ/ledger-live-mobile/pull/819
2. ✅ ~~Move the code in live-common and only keep the specific parts (it's already isolated in libcore/react-native folder here)~~
3. ✅ ~~Make a PoC from that live-common with the node libcore bindings to establish the minimal node code required and quickly check that all libcore functions are working in node.js bindings context.~~ https://github.com/LedgerHQ/ledger-live-common/pull/177
4.  ✅ ~~Integrate back in **mobile**~~.  https://github.com/LedgerHQ/ledger-live-mobile/pull/829
5. 👋 migrate desktop code carefully and backport any desktop innovation that was not yet in live-common libcore (for instance, the getVersion, and some extra stuff we need)
6. Simplify the libcore functions. For instance, `signAndBroadcast` need to be a simpler api `(Account, Transaction, devicePath)=>..` . Transaction also will need to evolve because it will be different on different families of coin,... komodo have a todo comment on that we should use getLockTime() , check with libcore team
7. implement ethereum libcore code in live-common. We need limit the number of branching and not doing ifs everywhere.
8. Integrate libcore ethereum back in desktop/mobile (need to rethink the bridge for libcore ethereum)

### Type

code sharing

### Context

https://github.com/LedgerHQ/ledger-live-common/pull/187

all libcore related parts.

### Dev notes

- We're still using the old Bridge interface, we might later migrate to the same Bridge type, but we won't easily be able to SHARE that bridge code until we have a proxy of device AND a proxy of libcore exposed from renderer (not planned yet).
- The CLI is moved out. If we want to, we should rework on this idea using directly live-common, which is basically what was done in https://github.com/LedgerHQ/ledger-live-common/tree/master/test/src/examples so it wouldn't be hard to recreate it.
- It is likely that we have many dead dependencies. For instance, I don't even think we need `hw-app-btc` anymore. We'll need to do a cleanup of all unused libs and unused files in Ledger Live. There is also probably code that are available in live-common and that we should continue migrate too, for instance we need to drop `helpers/signTransactionForCurrency`


gre:
- [ ] need to remap the error

PR #1890, opened 2019-03-28
Reset user data on reinstall

juan-cortes:
image

### Type

Improvement

### Context

n/a

For production builds only, this creates a flag that, on a detected reinstall of a different executable, will delete the userdata to work on a clean installation. Relaunching the app from the same executable should not affect the user data, however, 

### Known issues
- It will sometimes cause an internetal crash that I haven't had time to locate, maybe part of the dist script, or something.

### TODO

- [x] Create a cool looking diagram
- [x] MacOS testing (manual flag altering)
- [ ] Windows testing
- [ ] Linux testing
- [ ] Testing on a real update if it actually deletes the flag file
- [ ] Cleanup the check code, it's messy

### DIFF
--- a/src/commands/autoUpdate.js
+++ b/src/commands/autoUpdate.js
@@ -7,6 +7,7 @@ import { Observable } from 'rxjs'
 import { UPDATE_CHECK_IGNORE } from 'config/constants'
 // import createElectronAppUpdater from 'main/updater/createElectronAppUpdater'
 import type { UpdateStatus } from 'components/Updater/UpdaterContext'
+import { deleteInstallFlag } from '../helpers/reinstallCleanup'
 
 type Input = {}
 type Result = {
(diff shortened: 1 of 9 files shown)


gre:
@juan-cortes can you squeeze the commits into one with more meaningful into in the commit message? thanks

juan-cortes:
> @juan-cortes can you squeeze the commits into one with more meaningful into in the commit message? thanks

I mean, it said wip and mvp, I made it more verbose and moved the develop merge up to make it cleaner but it's the same thing really. We need to test it on windows, improve it, or kill it. Maybe there's a simpler way, maybe not, maybe maybe

gre:
ok let's keep as experimental until we work on it. we can rebase to one commit when we're ready ;)

PR #1894, opened 2019-04-01
feat: LL-1004 Use estimage-gas-limit to prefill gas

juan-cortes:
When using the v3 explorer for Ethereum get the estimated gas limit from the servers instead of using the hardcoded 21k. This will graciously fallback to the 21k for v2. Note that we will need to set the `USE_EXPERIMENTAL_EXPLORERS` flag programmatically for this to work, or flick the switch in the experimental features screen once it's available.

v2v3gastimate

### Type

Feature

### Blocked by

https://github.com/LedgerHQ/ledger-live-common/pull/191

Needs to have `live-common` bumped with the code from that pr

Send flow, for Ethereum, try using a contract vs non-contract address, see if the gas limit changes.

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -460,6 +460,7 @@ const EthereumBridge: WalletBridge<Transaction> = {
     t.amount.isGreaterThan(0) &&
     t.gasPrice &&
     t.gasPrice.isGreaterThan(0) &&
+    t.gasLimit &&
     t.gasLimit.isGreaterThan(0)
       ? Promise.resolve(t.amount.plus(t.gasPrice.times(t.gasLimit)))
       : Promise.resolve(BigNumber(0)),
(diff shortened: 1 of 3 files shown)


juan-cortes:
@Arnaud97234 good catch, by the way, for testing this until we have the v3 experimental feature in place it will estimate 21k always using v2. Fixed 🤞

gre:
@juan-cortes can you rebase the branch? thx

juan-cortes:
image

To test this (failing everything) you need to enable the `EXPERIMENTAL_EXPLORERS` flag in .env
For now, it doesn't make sense to make it visible for users because we are still hitting the CORS issue on dev builds. @gre don't know what we can do about this 🤷‍♂️

gre:
We're waiting the fix to be deployed for that but that shouldn't block the PR to be merged to me. For now we can test that everything works like before and assuming the feature will be available later.

gre:
@Arnaud97234 this can be retested I think

PR #1896, opened 2019-04-02
better min value for Y scale of charts

backpacker69:
better min value for charts

### Type

UI Polish

### Context

making charts more useful effort

all charts


gre:
thanks for the PR! this will be integrated in https://github.com/LedgerHQ/ledger-live-desktop/pull/1900 ;)

PR #1897, opened 2019-04-02
Fixes the implementation of the Ledger Core RNG’s interface.

phaazon:
This fixes LLC-196. A proper fix has also been pushed to Ledger Core to
have a better error message for when such a tricky incorrectly
implemented interface is detected. See [#194](https://github.com/LedgerHQ/lib-ledger-core/pull/194) for a PR in Ledger Core for a pre-emptive bug fix (you haven’t come across it yet but my unit tests found it). I also added a way to detect the RNG issue.


phaazon:
@valpinkman

PR #1898, opened 2019-04-02
feat: LL-1077 Add failsafe to the send flow when fees are too high

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1077

The send flow, try to send an amount with fees representing more than 10% of the amount, you should be prompted with the above modal to confirm that unusual transaction. The modal should not be dismissable by clicking outside, the user needs to either confirm and go to the next step, or cancel and remain on the first step of the flow.

### DIFF
--- a/src/components/base/Modal/ConfirmModal.js
+++ b/src/components/base/Modal/ConfirmModal.js
@@ -17,7 +17,7 @@ type Props = {
   isDanger: boolean,
   title: string,
   subTitle?: string,
-  desc: string,
+  desc?: string,
   renderIcon?: Function,
   confirmText?: string,
   cancelText?: string,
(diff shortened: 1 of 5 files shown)


src/components/base/Modal/ConfirmModal.js
```diff
@@ -52,13 +53,14 @@ class ConfirmModal extends PureComponent<Props> {
...
       children,
+      narrow,
       ...props
     } = this.props
 
     const realConfirmText = confirmText || t('common.confirm')
     const realCancelText = cancelText || t('common.cancel')
     return (
-      <Modal isOpened={isOpened} centered={centered}>
+      <Modal isOpened={isOpened} centered={centered} width={narrow && 380}>
```
gre: how did this pass flow?

narrow can be a `boolean`, so it can be `false`. width however takes a `?: number` not a `?: number | boolean` . 

I would suggest to do `narrow ? 380 : undefined` , less fancy but usually better for type safety

juan-cortes: oh my

PR #1900, opened 2019-04-02
Performance: rework graph calculation

gre:
- Performance: rework the graph data calculation with more caches (done in live-common)
- React: upgrade to latest
- Portfolio: account individual charts are in crypto values
- Account page: graphs are shown in crypto values instead of fiat value to show the actual value evolution.

(mock data, not my actual portfolio^^)

Capture d’écran 2019-04-02 à 20 02 49
Capture d’écran 2019-04-02 à 20 02 31

Note for more performance hunting:
- we could add even more cache if we could cache the portfolio and the historyWithCountervalues objects in live-common. I'll tackle that a bit later but it's not a blocker for this PR.

### Type

Performance and Feature improvements

Performance should be better. Need to test with as many account as possible and compare performance with previous version.


src/components/base/Chart/Tooltip.js
```diff
@@ -46,35 +30,7 @@ const Tooltip = ({
...
-              />
-            )}
-            <Box ff="Open Sans|Regular" color="grey" fontSize={3} mt={2}>
-              {moment(item.date)
-                .add(1, 'second')
-                .format('LL')}
-            </Box>
-          </Fragment>
-        )}
+        {renderTooltip ? renderTooltip(item) : null}
```
meriadec: I guess if `renderTooltip` is undefined we can avoid render the whole Tooltip

src/reducers/accounts.js
```diff
@@ -4,6 +4,7 @@ import { createSelector } from 'reselect'
 import { handleActions } from 'redux-actions'
 import accountModel from 'helpers/accountModel'
 import logger from 'logger'
+import { genAccount } from '@ledgerhq/live-common/lib/mock/account'
```
meriadec: Ye

src/components/AccountPage/AccountBalanceSummaryHeader.js
```diff
@@ -11,41 +9,24 @@ import type { Currency, Account } from '@ledgerhq/live-common/lib/types'
...
   isAvailable: boolean,
-  totalBalance: BigNumber,
-  sinceBalance: BigNumber,
-  refBalance: BigNumber,
-  accountId: string,
-}
-
-type Props = OwnProps & {
+  first: *,
+  last: *,
```
meriadec: Not sure I understand why it is `*` ? In this case we know what it contains

src/reducers/accounts.js
gre: oops

src/components/base/Chart/Tooltip.js
gre: good point

src/components/AccountPage/AccountBalanceSummaryHeader.js
gre: basically because i was lazy :p

gre:
thanks for the review @meriadec , i'll address the raised issues, all good point.

meriadec:
Nice perfs improvements btw, definitely smoothier. I can't imagine in prod mode! :rocket:

PR #1902, opened 2019-04-03
Correct active style for Exchange menu link

meriadec:
too triggering to let alive :doge:

before | after
---|---
2019-04-03_17-42-53_437x303 | 2019-04-03_17-42-36_444x303


khalilbenihoud:
cool!

PR #1904, opened 2019-04-04
adds experimental feature in a dedicated menu

valpinkman:
Adds an experimental feature section in Settings

### Type

Feat

### Context

LL-1150

Settings


src/components/ManagerPage/AppsList.js
```diff
@@ -145,12 +141,12 @@ class AppsList extends PureComponent<Props, State> {
...
 
   async fetchAppList() {
-    const { deviceInfo, isDevMode } = this.props
+    const { deviceInfo } = this.props
 
     try {
       const filteredAppVersionsList = await manager.getAppsList(
         deviceInfo,
-        isDevMode,
+        getEnv('MANAGER_DEV_MODE'),
```
gre: quick thought in the future: we might introduce a `useEnv("MANAGER_DEV_MODE")` to use with react hooks. without react hooks, it would be a `withEnvs` hoc i guess. It would trigger a re-render when env changes (with the observable we have from live-common).

here we really have no problem because the manager screens can't appear at same time as the settings. However on mobile we need it since all screens are always mounted.

src/components/SettingsPage/ExperimentalSwitch.js
```diff
@@ -0,0 +1,47 @@
...
+
+import React, { PureComponent } from 'react'
+
+import Track from 'analytics/Track'
+import Switch from 'components/base/Switch'
+
+type Props = {
+  name: string,
+  valueOn?: mixed,
+  valueOff?: mixed,
```
gre: remove the two `?:` and define a `defaultProps = { valueOn: true, valueOff: false }` . that would simplify the logic inside `onChange`

src/components/SettingsPage/sections/Experimental.js
```diff
@@ -0,0 +1,88 @@
...
+  description: string,
+  valueOn?: any,
+  valueOff?: any,
+}
+
+const ExperimentalFeatureRow = ({ feature }: { feature: Feature }) => {
+  const { type, ...rest } = feature
+  let children = null
+
+  if (type === 'toggle') {
```
gre: instead, define a map `controls = { toggle: ExperimentalSwitch }` . (if we have more cases in future it won't be a big series of if)

.prettierignore
```diff
@@ -1,2 +1,3 @@
 package.json
 test-e2e/**/*.json
+*.json
```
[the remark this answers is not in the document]
valpinkman: oops

src/components/SettingsPage/ExperimentalSwitch.js
```diff
@@ -0,0 +1,47 @@
...
+  checked: boolean,
+  readOnly: boolean,
+  onChange: (name: string, val: mixed) => boolean,
+}
+
+export default class ExperimentalSwitch extends PureComponent<Props, { checked: boolean }> {
+  state = {
+    checked: !!this.props.checked,
```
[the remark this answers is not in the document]
valpinkman: it was not a boolean before, will change

src/components/ManagerPage/AppsList.js
valpinkman: hooks <3

PR #1905, opened 2019-04-04
LL-1267 Extract non-translatables terms for English, styles + backport

juan-cortes:
This pr should address the inconsistencies in translating non-translatable strings such as dongle screen texts that would deem the instructions useless since they wouldn't match what the user is seeing. Also brings **bold text** to those non-translatable strings on the instructions for nanox, they weren't highlighted.

Screenshot 2019-04-04 at 18 07 15

> There is at least one case in Chinese where I was unable to backport the translation, I'm assuming crowdin will do somethinga bout it when the translation changes


gre:
some ideas we could improve:

- we need to remove from translation the account name things. the `(legacy) (unsplit)` things.
- for errors, each time we have `title: "{{message}}"` , I suggest we remove BUT in the TranslatedError we will fallback on `error.message` by default if no translation is available. tell me wdyt about this, just a thought, but i hate seeing these `{{message}}` useless things :D

juan-cortes:
I agree. It's a first step, there are several useless literals with only a placeholder for something else, like the repair device options right now. They shouldn't have to go through i18n, I'll take a look

juan-cortes:
This is now updated, I had some issues with the merge with Korean, same as I did with Chinese. If both cases I replaced the wording back to English version since I was unable to locate the non-translatable wording in the string.

gre:
@juan-cortes conflict again, sorry 😢 

I will set this PR as important because we want it merged soon so crowdin don't collide each time.

gre:
@Arnaud97234 i'm not sure this PR is the point for these discussion but more to be discussed directly on Crowdin or internally

static/i18n/en/app.json
```diff
@@ -526,9 +526,9 @@
         "Use this tool to restore your Ledger device if it has become stuck during a firmware update.",
       "desc": "Please select the state of your Ledger device.",
       "button": "Repair",
-      "generic": "Bootloader",
-      "mcuNotGenuine": "MCU not genuine",
-      "followDeviceRepair": "Follow device repair instructions"
+      "generic": "{{repairDeviceBootloader}}",
+      "mcuNotGenuine": "{{repairDeviceMcuNotGenuine}}",
+      "followDeviceRepair": "{{repairDeviceFollow}}"
```
gre: these 3 wordings should be removed from translation since there is no value to add them.

juan-cortes: I see

juan-cortes:
This should be ready to merge now, every time we get new translations I have to manually update the language files and it's time-consuming and error-prone -.-

PR #1907, opened 2019-04-05
header tag for experimental features + hooks - react-hot-loader

valpinkman:
### Type

feat

### Context

LL-1150

header in application should have a 'experimental feature' tag when using experimental flags

image


valpinkman:
Hooks

image

PR #1913, opened 2019-04-08
Workaround to fix bug when moving from old version

gre:
Fixes a bug where extra end up being null or undefined. No clear idea how this can happen but this will protect it.

### Type

edge case bug fix

### Context

LIVE-DESKTOP-3CNN

you should still see the **tag** extra field in a XRP operation detail when you use one.


gre:
after investigating with Arnaud, this bug can happen if users have XRP account, because in fact we don't clear cache when you upgrade. so potentially you have old op and synchro is not refreshing them because it's considered immutable.

PR #1916, opened 2019-04-09
fix add account not checking for dev mode anymore

valpinkman:
After the migration to experimental features, add account was not taking into account the dev mode anymore


src/hooks/useCryptocurrencies.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import { useState, useEffect } from 'react'
+import type { CryptoCurrency } from '@ledgerhq/live-common/lib/types'
+
+import { listCryptoCurrencies } from 'config/cryptocurrencies'
+import useEnv from 'hooks/useEnv'
+
+const useCryptocurrencies = (onlyTerminated: boolean = false): ?(CryptoCurrency[]) => {
+  const devMode = useEnv('MANAGER_DEV_MODE')
+  const [cryptos, setCryptos] = useState(() => listCryptoCurrencies(devMode, onlyTerminated))
```
gre: this state is not required. you can just `const cryptos = listCryptoCurrencies(devMode, onlyTerminated)`

valpinkman: yeah but if we switch devMode off don't we need to recalculate the cryptocurrencies ?

valpinkman:
> on the file name, can we rename to `useCryptoCurrencies.js` ? (good luck trying to say that to git now :P you might need to force push)

`git mv -f OldFile.js newFile.js`

src/hooks/useCryptocurrencies.js
gre: :magic: of hooks

PR #1919, opened 2019-04-10
Parallel sync for better performance & remove sync timeout

gre:
This should significantly increase the speed of Synchronisation when you have more than one account in Ledger Live because it will start doing synchronisation in parallel. It also remove the timeout that was existing on the sync and that was potentially creating issues for very big accounts.

### Type

performance

### Context

LL-1239

We need to test this intensively with many accounts. The risk here is that the libcore can crash because we ask it too much, we can also have race condition **but normally we shouldn't because we use the same codebase as mobile and mobile have this since the beginning**.


meriadec:
:tada:

PR #1926, opened 2019-04-11
LL-1275 feat: Allow user to search accounts by name, currency, ticker

juan-cortes:
On the send/receive flow, add the ability to search for an account using the currency ticker, currency name, or account name. It used to be using `account.id` which after talking with @gre justifies some weird matches for some currencies.

Apr-11-2019 16-57-22

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1275

Send/Receive full, make sure account selection is not broken after this.

### DIFF
--- a/src/components/SelectAccount/index.js
+++ b/src/components/SelectAccount/index.js
@@ -62,7 +62,8 @@ type Props = {
   t: T,
 }
 
-const getOptionValue = account => account.id
+const getOptionValue = account =>
+  `${account.currency.ticker}|${account.currency.name}|${account.name}`
 
 const RawSelectAccount = ({ accounts, onChange, value, t, ...props }: Props) => {
   const selectedOption = value ? accounts.find(o => o.id === value.id) : null


src/components/SelectAccount/index.js
```diff
@@ -62,7 +62,8 @@ type Props = {
   t: T,
 }
 
-const getOptionValue = account => account.id
+const getOptionValue = account =>
+  `${account.currency.ticker}|${account.currency.name}|${account.name}`
```
gre: does the `|` creates a problem for the search engine? i wonder if he consider a word. For instance, `Ethereum Classic`, if you search Classic (and your account name does not have classic in it) does it work?

because I wonder if `ETC|Ethereum Classic|foo` would make some problem

Intuitively i would have used a space

juan-cortes: I doubt it would cause any issues, the filtering is done using a simple `indexOf(input) > -1` after some cleanup. As long as the input string is somewhere in the concat we built it should be fine. If you feel more comfortable with spaces feel free to commit 🤷‍♂️ 

https://github.com/JedWatson/react-select/blob/master/src/filters.js#L40

PR #1931, opened 2019-04-12
fix ticker search in manager

valpinkman:
adds missing tickers when searching apps in manager

### Type

Improvement

### Context

LL-1246

### Parts of the app affected / Test plan

Manager

PR #1934, opened 2019-04-16
fix: Added tickers for apps not supported on Live

juan-cortes:
The manager search bar was not returning any results for tickers for currencies that are not supported on Live. This allows the merge with the tickers whilst still allowing for memoization.
Apr-16-2019 10-42-33

### Type
Regression

### Context
https://ledgerhq.atlassian.net/browse/LL-1246

In the manager search for an app that's not supported by Live


gre:
seems like it's same as #1931 ?

juan-cortes:
Oh damn, the issue on jira was not assigned to anyone 🤦‍♂️

PR #1936, opened 2019-04-16
Major bump of live-common

gre:
this is to test https://github.com/LedgerHQ/ledger-live-common/pull/189

It contains many preparation work for Q2, especially for the incoming libcore ethereum transition, erc20 and send max implementations.

### Type

preparation

### Context

literally Q2

- **Basically everything should work like before apart from the point below**
- There are some performance improvement
- There is a new Experimental Features: Scan for invalid paths. It scan for 2 new cases on Bitcoin derivations: segwit on legacy and legacy on segwit. This is an opt-in because we don't want to automatically scan for these for all users. That means path that are invalid are no longer scanned, however, path that are considered "legacy" are still scanned regardless. The logic behind is that if Ledger ever allowed to import such account it's legacy, otherwise it's invalid because user's fault.
- There is a slight change on the naming that you will see, you will see account names to contain `(segwit)` where it was not the case before. This is to prepare what we'll have later displayed as a "visual tag" instead of being part of the name.


gre:
> Invalid accounts should be retrieved during "Add account" process:

as explained in the description, you need to enable a experimental feature, have you? @Arnaud97234

gre:
@Arnaud97234  fixed

gre:
ready to review again

src/components/modals/Debug.js
```diff
@@ -51,6 +51,7 @@ class Debug extends Component<*, *> {
       for (let x = 0; x < 20; x++) {
         const { address, path } = await getAddress
           .send({
+            derivationMode: 'segwit',
```
juan-cortes: `account.derivationMode` ?

gre: in context of this Debug screen, it seems we only do segwit derivations. https://github.com/LedgerHQ/ledger-live-desktop/blob/7027e6316d5734166c1b8863c4a656791f631830/src/components/modals/Debug.js#L49

PR #1938, opened 2019-04-17
adds error modal if getLatestFirmwareForDevice fails

valpinkman:
If `getLatestFirmwareForDevice` fails, display a modal with Error

### Type

Improvement

### Context

LL-1245

Manager / Firmware Update

image

### DIFF
--- a/src/components/ManagerPage/FirmwareUpdate.js
+++ b/src/components/ManagerPage/FirmwareUpdate.js
@@ -1,7 +1,6 @@
 // @flow
-/* eslint-disable react/jsx-no-literals */ // FIXME
 
-import React, { PureComponent, Fragment } from 'react'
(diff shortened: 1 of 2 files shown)


gre:
what I was suggesting in LL-1245 (comment) is that we should be erroring in the error case of the Firmware modal. because we're in the context of a firmware update.

we don't need to introduce a new modal but just reuse the existing modal. is it feasible?

valpinkman:
Reusing update firmware modal

image

gre:
neeeaaaat!

gre:
@Arnaud97234 i would suggest we merge it because we'll have more to test regarding firmware update in my other live-common PR. i would like to rebase it with it. is that ok we merge it and test again later?

Issue #1189, opened 2018-07-18
Current crypto price on each account page & market price


leyei:
- [ x ] I have checked this feature was not yet requested.

Can you add current crypto price on each account page when a crypto is viewed? Example when Litecoin is selected, maybe somewhere in the space between the Litecoin Ł logo and the send button.

Would also be nice to have market prices just as it is on coinmarketcap.com. [Check how it is nicely implemented in the Delta Cryptop portfolio tracker app]

#### Part of the application

Under each account and for the market prices, under 'Menu'

davidzet2:
Agreed!

Now you can see your portfolio value (PRICE*QUANTITY) for each asset, but it would be nice to see the PRICE of each coin. Can you add this feature?

tmlee:
As a consideration, may i suggest looking at the CoinGecko API at https://www.coingecko.com/api
No keys and straightforward to consume.
Happy to answer any questions you guys might have, cheers!

gre:
i consider this addressed by the new portfolio distribution

Issue #1135, opened 2018-07-14
Show coin quotation


ibiscp:
In Portfolio tab, it's possible to see the list of accounts and the amount of each one, but it doesn't show anywhere the price of 1 unit of the coin. If we want to know, for example, the quotation of 1 BTC, we need to get the total of the account and divide by how many BTC we have.

fuorissimo:
i need the same. a live price on owned cryptoes

PR #1944, opened 2019-04-20
Current crypto price on each account page & market price

jorge-cob:
image

Guess this should get through design validation but thought it could be a head start for this issue :)

### Type

Feature

### Context

Fixes https://github.com/LedgerHQ/ledger-live-desktop/issues/1189
Fixes https://github.com/LedgerHQ/ledger-live-desktop/issues/1135

Account screen header


gre:
Nice!! I would suggest the use of `@` to express "at rate of" 

`USD 10.00 ( @ USD 300 )`

But we'll rediscuss internally if we have another design idea

jorge-cob:
@gre I have the actual design, I'm working on it

gre:
we can close thanks to rework in #2050

PR #1947, opened 2019-04-23
Max account decimals before trim and tooltip for whole value

jorge-cob:
maxLength

### Type

Bug Fix, UI Polish...

### Context

Account's balance was overflowing the wrapper where there are lot of decimal values in a small screen

before maxLength

### DIFF
--- a/src/components/base/FormattedVal/index.js
+++ b/src/components/base/FormattedVal/index.js
@@ -21,6 +21,7 @@ import FlipTicker from 'components/base/FlipTicker'
 
 import IconBottom from 'icons/Bottom'
 import IconTop from 'icons/Top'
+import Tooltip from 'components/base/Tooltip'
 
 const T = styled(Box).attrs({
   ff: 'Rubik',
@@ -49,6 +50,7 @@ type OwnProps = {
   animateTicker?: boolean,
   disableRounding?: boolean,
   isPercent?: boolean,
+  maxLength?: number,
 }
 
 const mapStateToProps = (state: State, _props: OwnProps) => ({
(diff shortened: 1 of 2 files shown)


src/components/base/FormattedVal/index.js
```diff
@@ -131,6 +140,11 @@ function FormattedVal(props: Props) {
       )}
     </T>
   )
+
+  if (showTooltip) {
+    return <Tooltip render={() => unit && formatCurrencyUnit(unit, val)}> {content} </Tooltip>
```
gre: should the `unit` check be done in the if? because as far as i read it, if there is no unit, we would have no tooltip right?

```
if (showTooltip && unit) {
    return <Tooltip render={() => formatCurrencyUnit(unit, val)}> {content} </Tooltip>
}
```

I imagine this case never present, but code is easier to understand

gre:
good

gre:
Ok so base on this feedback and fact the window can resize & need to be somewhat responsive, i'm going to suggest a different implementation:

- For the actual `...` rendering, I would prefer we use classical CSS ellipsis https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow . That's the only true way we can make it correct, because (1) the font is not monospace (2) it have different sizes on different OSs (3) you can resize the window so the width possibilities is infinite.
- For dynamic presence of the Tooltip, this is getty a bit harder to implement it right I feel. We would need a way to know if the CSS ellipsis is on or not, and you probably need to check for that at each window resize or even re-rendering.. we will have to be very careful about performance of this / tradeoff can be found.

gre:
I think it was fixed in Accounts rework

PR #1948, opened 2019-04-24
flow-types are committed now and added npm script to download them

valpinkman:
Removed flowtypes from post install scripts. Flowtypes are now committed and a new npm script `flow-typed` will take care of installing and cleaning the flow defs

### Type

Chore

### Context

LL-1314

Config


package.json
```diff
@@ -18,6 +18,9 @@
     "test": "jest src",
     "test-e2e": "jest test-e2e",
     "test-sync": "bash test-e2e/sync/launch.sh",
+    "flow-typed": "npm run install-flow-typed && npm run clean-flow-typed",
```
gre: do you mean clean THEN install?

package.json
```diff
@@ -18,6 +18,9 @@
...
+    "install-flow-typed": "flow-typed install -s --overwrite",
+    "clean-flow-typed": "rm -f flow-typed/npm/{reselect_*,redux-actions_*,react-redux_*,react-i18next_v7.x.x.js,styled-components_v3.x.x.js,winston*}",
```
gre: can you rename to something like `remove breaking-flow-typed` ? to make it more obvious it's not about removing the flow-typed before reinstalling

valpinkman: ok

PR #1961, opened 2019-04-30
adds usb only to nano x in onboarding

valpinkman:
adds a  `usb only` label to nano x onboarding

### Type

Feat

### Context

LL-1307

Onboarding

image

image


dasilvarosa:
🚀

PR #1965, opened 2019-04-30
adds experimental language flags

valpinkman:
Allows user to test for experimental languages

### Type

Feat

### Context

still nope

Experimental Settings and Language Settings

I'd say some of the work is done for this. It does work, however there is still a reducer that is using the `languages` exported from `config/languages.js` which probably shouldn't be the case. Also, we need to bump live-common as well (see here: https://github.com/LedgerHQ/ledger-live-common/pull/224)

so HODL for now :)


gre:
we'll address the two bugs in next sprint but since they affect exclusively the flags and were existing bugs it's ok to merge.

PR #1967, opened 2019-05-02
LL-1250 Accounts section

juan-cortes:
image
image
image

image

contextmenu

### Type

Feature
### Context
https://ledgerhq.atlassian.net/browse/LL-1250

### Pending tasks
- [x] Right click actions for cards/rows
- [x] Confirm the status is correct, I have my doubts, @gre 
- [x] Styling for Range selector is not polished
- [x] Empty state for accounts in sidebar
- [x] Sidebar _helper_ for accounts moved

New accounts screen


src/components/AccountsPage/AccountList/index.js
```diff
@@ -0,0 +1,72 @@
...
+      search: evt.target.value,
+    })
+
+  render() {
+    const { accounts, range, onAccountClick, onRangeChange } = this.props
+    const { mode, search } = this.state
+    const isUsingCards = this.state.mode === 'card'
+    const Body = isUsingCards ? GridBody : ListBody
+
+    const filteredAccounts = accounts.filter(account =>
```
gre: for performance. we need to virtually split into two groups. the visibleAccounts and invisibleAccounts.

then in the rendering we will display the rows and the grid with this order: `[ ...visibles, placeholder, ...invisibles ]` (pseudo code). invisibles would be mapped with an extra "invisible". just make sure everything are still siblings, DOM speaking. that way, react will know that to optimize it, it just need to move DOM elements around.

src/components/ContextMenu/ContextMenuWrapper.js
```diff
@@ -0,0 +1,123 @@
...
+import React, { PureComponent } from 'react'
+import { Trans } from 'react-i18next'
+import styled from 'styled-components'
+import Box from 'components/base/Box'
+import { colors } from 'styles/theme'
+
+// $FlowFixMe
+export const ContextMenuContext = React.createContext({})
+
+export const withContextMenuContext = (ComponentToDecorate: React$ComponentType<*>) => (
```
gre: this is very cool! side note for later, you could experiment around `useContext` hehe

yarn.lock
```diff
@@ -13585,6 +13599,19 @@ react-portal@^4.1.2:
   dependencies:
     prop-types "^15.5.8"
 
+react-redux@5:
```
gre: mmh, need to checkout the yarn.lock changes, none required.

src/components/AccountsPage/AccountList/index.js
juan-cortes: mmm I didn't notice any lag on my end, but I like this idea of only making them invisible instead of removing them, I'll play with it

Luuk2016:
Looking good @juan-cortes! 🎉

src/components/AccountsPage/AccountList/index.js
gre: if you have 20+ accounts and you fastly play with the search switching between 'a' and '' , you notice it, it's slight but still affect the performance perception when you type characters. i locally fixed it on my end and the result was much better

juan-cortes: Solved with your pr

yarn.lock
juan-cortes: Done

juan-cortes:
All that remain is the empty state for the accounts on the sidebar, which is pending design. And the sync state that you were going to take a look at @gre I'll look into the graph resize issue tomorrow

gre:
@juan-cortes  you can see my recent changes, i guess we'll commit that yarn.lock, for some reason develop's branch must have a wrong one.

src/components/AccountsPage/AccountSyncStatusIndicator.js
```diff
@@ -0,0 +1,130 @@
...
+      // a user action is kept in memory for a short time (which will correspond to a spinner time)
+      clearTimeout(timeout.current)
+      timeout.current = setTimeout(() => setUserAction(false), 1000)
+    },
+    [setSyncBehavior, accountId],
+  )
+
+  // at unmount, clear all timeouts
+  useEffect(() => {
+    clearTimeout(timeout.current)
```
gre:
```suggestion
    return () => clearTimeout(timeout.current)
```

gre:
- [ ] wordings

PR #1977, opened 2019-05-09
LL-1346 Partial fix for overflow on Account balance + tooltip

juan-cortes:
This is not a complete solution but it's what we've settled with, for now, we simply hide the overflowing digits and maintain the box for the range. Should be good enough for now, until we either remove the digit ticker that prevents a proper ellipsis or change the layout a big.

Also, the tooltip was getting broken. The current fix is to ellipsis it but we should make the tooltip a bit wider to accommodate the extra characters maybe.

### Type

UI Polish. 

Account page, and generally anywhere using a balance to check if it broke somewhere

### DIFF
--- a/src/components/base/FormattedVal/index.js
+++ b/src/components/base/FormattedVal/index.js
@@ -30,6 +30,10 @@ const T = styled(Box).attrs({
 })`
   line-height: 1.2;
   white-space: pre;
+  text-overflow: ellipsis;
+  display: block;
+  width: 100%;
+  overflow: hidden;
 `
 
 const I = ({ color, children }: { color?: string, children: any }) => (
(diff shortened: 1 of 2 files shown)


src/components/base/FormattedVal/index.js
```diff
@@ -114,7 +121,7 @@ function FormattedVal(props: Props) {
   })
 
   return (
-    <T color={color || marketColor} withIcon={withIcon} {...p}>
+    <T color={color || marketColor} withIcon={withIcon} {...p} style={ellipsisStyles}>
```
gre:
```suggestion
    <T color={color || marketColor} withIcon={withIcon} style={ellipsisStyles} {...p}>
```

in case at some place we were overriding the style we shouldn't break this possibility

also, shouldn't this be done in `T`'s style itself?

juan-cortes: Moved the styles to T

PR #1978, opened 2019-05-09
Fixes LL-1357 (race condition): glitch affecting ETH experimental

gre:
A race condition can happen in this case:
- if estimateGasLimit is slow
- in the meantime user edit any of the transaction fields

to cover this problem, we use the same protection as for all the resync() code: using a syncId

---

There is however room for improvement in the current implementation that we will have to tackle.

- ⚠️ current code will call `estimateGasLimit` for EVERY CHANGE of transaction changes, this means every key stroke, even if it's not needed (because it's done in a resync) will trigger an http request and will be prone to the present bug. Instead we need to only change if recipient changes.

proof when editing the amount:
Capture d’écran 2019-05-09 à 14 27 26

- RecipientField is in fact not the correct place to call this `estimateGasLimit`. Conceptually `gasLimit` is part of the transaction object but is also the asynchronous result of that same transaction object. For this kind of effect, either we need to abstract out this concept in the bridge OR we could move the current logic in the GasLimit field itself (which would somehow make sense, it's part of the same component afterall). I would suggest solution 2 for now because bridge will be refactored soon.
- ⚠️ If user have started modifying the gasLimit, it's getting **overridden** each time something else is modified. the correct behavior should be to no longer touch it if user have, or alternatively, only if address changes (which would be easy with the first point).

### Type

Bugfix

### Context

LL-1357

Send, step 1

### DIFF
--- a/src/bridge/EthereumJSBridge.js
+++ b/src/bridge/EthereumJSBridge.js
@@ -4,7 +4,7 @@ import { BigNumber } from 'bignumber.js'
 import logger from 'logger'
 import React from 'react'
 import FeesField from 'components/FeesField/EthereumKind'
-import AdvancedOptions from 'components/AdvancedOptions/EthereumKind'
+import EditAdvancedOptions from 'components/AdvancedOptions/EthereumKind'
 import throttle from 'lodash/throttle'
 import flatMap from 'lodash/flatMap'
 import uniqBy from 'lodash/uniqBy'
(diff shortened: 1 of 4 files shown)


gre:
The two problem mentioned in the description are fixed by the second commit.

There will be one problem to address tho:
We have no way to visually see something is still loading so you can potentially click next before the fees are loaded

I guess it's acceptable as the fees are calculated after you set address so it have enough time to load..

PR #1986, opened 2019-05-13
add a new failed indicator on operation list and details

valpinkman:
Adds a new indicator for failed ETH transactions in OperationsList and OperationDetails

### Type

Improvement

### Context

LL-1316

Account page and Operation Details

image

image


valpinkman:
@gre I've tested with and without experimental nodes, clearing the cache everytime. 
On v2 all the transaction say `confirmed`
Moving onto v3 we now have the `failed` indicator

PR #1991, opened 2019-05-15
LL-1219 Reduce flex-shrink to allow seeing more in modals

juan-cortes:
image

### Type

Reduce flex shrink to allow more visible fields. I still think we should _eventually_ split the first step of the send flow instead of gaining a little bit of height like this, but that's a different battle.

### Context

https://ledgerhq.atlassian.net/browse/LL-1219

Modals


gre:
merging this, easy to rollback when we'll test on develop if we don't want it :)

PR #2008, opened 2019-05-20
Rollback to previous electron and electron-updater

gre:
The 1.8.0 introduced a regression on Mac regarding bad implementation of @ledgerhq/electron-updater
The quick fixes for now is to rollback this and reimplement a better version later

The issue found was that: when upgrading from 1.8.0 to 1.8.1, the 'Install Now' button doesn't do anything, it should have quitAndInstall (it works on Linux and Windows). Also there is race conditions if you click multiple times on quitAndInstall so we're going to wait for a better implementation (fixes to be tested https://github.com/LedgerHQ/electron-builder/pull/1 ).

Finally, the fact that on Linux we were still using electron@2 is bad, we shouldn't have to diverge into two different versions. A solution need to be find to fix this.

### Type

revert a regression


gre:
@Arnaud97234 can you try to build this branch after modifying version to 1.8.0 (as if it was the previous 1.8.0) to see if the update properly work?

PR #2017, opened 2019-05-23
Update live-common, ledgerjs, libcore

gre:
Update recent changes of live-common and ledgerjs.

see https://github.com/LedgerHQ/ledgerjs/releases
and https://github.com/LedgerHQ/ledger-live-common/releases

### Type

bump

- test the different transports (proxy, hid,...)
- LL-1383 removed Coinberry from partner list
- clear cache might be affected because we have fixed the `freshResetAll` bindings. not sure how to test this but we might have to check that the libcore is still fully cleared.
- logs. export logs. we've improved our logs system. logs should be ~ the same. you will also have more verbosity and see the logs at the hid usb level.


gre:
also updated libcore

PR #2018, opened 2019-05-23
LL-1347 POC use LocalStorage as an alternative to file system

juan-cortes:
Due to the issues that some users have been reporting of losing data upon closing the application, or when updating it. This is a possible solution that will initially run as an experimental feature.

## What is this?
Leveraging on the work by @meriadec for the persistence layer I've modified it to allow the storage to happen either on the filesystem or on the local storage of the window. Had to tweak the window params part a bit since the window didn't exist at the time of getting the params but apart from that it's been pretty straightforward.

Apart from that, since there's no "User data folder" in this mode, I've refactored that portion to export the user data in a formatted JSON, we should probably do the same for the logs so as to make them more readable without formatting it (you know, just calling `JSON.stringify(wadus, null, 4)`)
image

## Chance usages
Since it stores the user data in two separate and isolated spaces we could potentially have two environments to test at the same time (dev side mainly) with different accounts and whatnot. I'm thinking @Arnaud97234 in particular could find this useful.

> Need to add the env variable to common and update the version here for this to be testable, hence the draft pr. @gre 

image
image

### Type
Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-1347

The whole app, it's a new storage system for all the data.

### DIFF
--- a/src/helpers/db/index.js
+++ b/src/helpers/db/index.js
@@ -12,6 +12,7 @@ import { fsReadFile, fsUnlink } from 'helpers/fs'
 import { promisify, debounce } from 'helpers/promise'
 
 import { NoDBPathGiven, DBWrongPassword } from '@ledgerhq/errors'
+import { getEnv } from '@ledgerhq/live-common/lib/env'
 
 type Transform = {
   get: any => any,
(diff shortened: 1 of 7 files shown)


src/helpers/db/index.js
```diff
@@ -28,7 +29,8 @@ let encryptionKeys = {}
 let transforms = {}
 
 const DEBOUNCE_MS = process.env.NODE_ENV === 'test' ? 1 : 500
-const save = debounce(saveToDisk, DEBOUNCE_MS)
+const save = debounce(persistData, DEBOUNCE_MS)
+const EXPERIMENTAL_LOCAL_STORAGE = getEnv('EXPERIMENTAL_LOCAL_STORAGE')
```
gre: this could would probably be better done at runtime than at boot time? what if the value change over time (check the box)

juan-cortes:
Closing it for now because I have ocd and I don't like seeing it left open

gre:
we'll open the topic again when we have more QA bandwidth again. 🙏

PR #2020, opened 2019-05-24
Gas slider fees

gre:
Requires a rebase with #2017

out

Capture d’écran 2019-05-24 à 11 44 48

What is getting addressed here:

- Better UX
- Having a plain field to edit the gas was making some users setting way too much fees because the confusion with the gasLimit field.

### Type

improvement


meriadec:
pretty cool

PR #2033, opened 2019-05-28
Bootstrap libcore ETH and ERC20

gre:
**ETH migration to libcore ethereum.**

- [x] step 1: migrate to new bridge interface.
- [x] step 2: implement Ethereum in LibcoreBridge.
- [x] step 3: rework the Add Accounts using live-common new logic that have ethereum migration logic.
- [x] handle again EIP55 degrated support (lowercase/uppercase are accepted too but with a warning) => probably move this logic at the live-common level too! `isValidRecipient` function is already ready for this: it is a `Promise<?Error>` that can either fail with an error or resolve with a warning, or null if none.
- [x] live-common v6 : https://github.com/LedgerHQ/ledger-live-common/releases/tag/v6.0.0

**ERC20 preparations**

- only portfolio have beginning of support for ERC20.
- Beware of these known issues:
  - internal txs are not yet in, which means you will see negative or incomplete ERC20 balances
  - temporary erc20 icons until we have designs

### Type

migration & feature

### Context

ERC20

need to carefully test all coins, especially all the subtle behaviors of the Send flow (especially step 1).


gre:
This PR is ready. this will require careful test & i suggest all the team to help on testing the present changes. @aramab @Arnaud97234 

- "Experimental Libcore" will allow to test the ETH migration
- "Experimental Nodes" can be used in combination to also test the v3 api (that includes gasLimit calculation and ERC20 data as well! (erc20 data are just not rendered, but if you want to take a look in the data they are ready!!! 🚀) )

This is the preparation for future ERC20 integrations.

PR #2036, opened 2019-05-29
LL-1377 Asset distribution block

### LINKED ISSUE #984: Show percentage of each crypto in respect to total portfolio
For those who use "asset allocation strategy" to allocate money between investments, it is really important to know how much percent each asset represent of your total portfolio.

So, together with the value of each account, it would be really interesting to see also how much percent that value represents.

### LINKED ISSUE #1983: Show cumulative balances for each cryptocurrency
It would be very beneficial if there was an easy way to see how much of each cryptocurrency you have assuming you have multiple accounts for each cryptocurrency. 

The Chrome app showed this in the top right corner but currently there's no feature for this on Ledger Live.

juan-cortes:
image

Asset distribution block for portfolio
### Type

Feature
### Context

- fixes #1983
- fixes #984
- Fixes https://github.com/LedgerHQ/ledger-live-desktop/issues/1135
- https://ledgerhq.atlassian.net/browse/LL-1377

Portfolio screen, asset distribution

### DIFF
--- a/src/components/AssetDistribution/index.js
+++ b/src/components/AssetDistribution/index.js
@@ -49,20 +49,28 @@ const mapStateToProps = createStructuredSelector({
     const accounts = accountsSelector(state)
     const counterValueCurrency = counterValueCurrencySelector(state)
     const toExchange = counterValueExchangeSelector(state)
-    return getAssetsDistribution(accounts, (currency: Currency, value) => {
-      // $FlowFixMe
-      const currencySettings = currencySettingsSelector(state, { currency })
-      const fromExchange = currencySettings.exchange
-      return CounterValues.calculateWithIntermediarySelector(state, {
-        from: currency,
-        fromExchange,
-        intermediary: intermediaryCurrency,
-        toExchange,
-        to: counterValueCurrency,
-        value,
-        disableRounding: true,
-      })
-    })
+    return getAssetsDistribution(
+      accounts,
+      (currency: Currency, value) => {
+        // $FlowFixMe
+        const currencySettings = currencySettingsSelector(state, { currency })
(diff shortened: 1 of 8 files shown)


src/components/AssetDistribution/index.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+  render() {
+    const { distribution } = this.props
+    const { showAll } = this.state
+    const subList = showAll ? distribution.list : distribution.list.slice(0, distribution.showFirst)
+
+    return (
+      <>
+        <Text ff="Museo Sans|Regular" fontSize={6} color="dark">
+          <Trans i18nKey="distribution.header" values={{ count: 0 }} count={subList.length} />
```
gre: `distribution.list.length`

src/components/AssetDistribution/index.js
```diff
@@ -0,0 +1,101 @@
...
+  counterValueCurrency: counterValueCurrencySelector,
+})
+
+class AssetDistribution extends PureComponent<Props, State> {
+  state = {
+    showAll: false,
+  }
+  render() {
+    const { distribution } = this.props
+    const { showAll } = this.state
```
gre: we need a secondary `showAll` variable that will fallback to "true" if showFirst is already near the list.length (so the show max don't show just 2 rows, better just show them already)

src/components/AssetDistribution/index.js
juan-cortes: oh so it's **not** the visible ones but the total?

fuorissimo:
add also public key (with copy botton) on each wallet, and is perfect

PR #2041, opened 2019-05-31
LL-1443 Redirect to accounts after account deletion

juan-cortes:
### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1413

Applications screen from context menu and application screen from settings. In the case of the context menu the scroll should remain the same and we should navigate to /accounts if we were in the account screen.


juan-cortes:
Sorry not sorry, I'll change it later

juan-cortes:
@gre there you go chief

PR #2042, opened 2019-05-31
Second iteration of Experimental USB

gre:
Experimental second implementation of USB: this time, it will correct reflect the USB state (connected or not) as well as the device model (nanos, nanox,..)

This should also improve the performance by not closing the HID device but keeping the transport opened and polling it.

We'll give lot of test to this implementation on different systems and computers.

To test:

- make sure the default implementation is untouched and works like before.
- make sure everything works with Experimental USB. Firmware Update to be tested too.

### DIFF
--- a/src/actions/devices.js
+++ b/src/actions/devices.js
@@ -2,12 +2,6 @@
 
 import type { Device } from 'types/common'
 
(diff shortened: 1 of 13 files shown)


gre:
use this as reference. this is not going to make it because polling is not that good for the device perf.

PR #2050, opened 2019-06-03
countervalue current price

jorge-cob:
Captura de pantalla 2019-06-04 a las 0 13 50
Captura de pantalla 2019-06-04 a las 0 18 30

### Type
UI Polish, Feature

### Context

closes https://github.com/LedgerHQ/ledger-live-desktop/pull/1944

Account screen header


gre:
Thanks for the contribution!

Code lgtm
We need to check what happens when countervalue are not available/ not yet loaded. We probably shouldn't render the whole thing in that case (a grey placeholder will already be shown).

Also, the rendering should probably go into a dedicated component.

juan-cortes:
https://ledgerhq.atlassian.net/browse/LL-1391

gre:
we will not ship this in the last sprint release, so feel free to take the time to finalize on these two topics:

- sharing logic in a component that returns that new price
- that component would not render anything if the rate is ZERO. which means maybe you might want to move the countervalue logic up in your new component and just use the underlying `<CurrencyUnitValue`

gre:
@jorge-cob Thanks for your contribution!

we'll take it from here. I'll work on the notes mentioned above because we have the same problem on the distribution grid anyway so i'll merge them all into a single component.

PR #2060, opened 2019-06-05
remove dot from text

valpinkman:
fix fr lang file

### Type

Fix

### Context

ll-1457

Onboarding


gre:
@valpinkman mmh the normal flow shouldn't fix the translated text in PR but directly from crowdin, i'll merge it because i  think crowdin is able to recover , but it can create conflicts issues in future.

PR #2063, opened 2019-06-05
CI: Start Ledger Live and wait for loading success

Arnaud97234:
The aim is to check if LL starts without error, in order to be sure that the app started properly I check in the logs that the libcore has been loaded.

**Known issue**
I'm not able to kill the running process so the job never ends in circle ci.

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,7 +162,7 @@
     "chance": "^1.0.13",
     "concurrently": "3.5.1",
     "dotenv": "^5.0.1",
-    "electron": "1.8.8",
+    "electron": "^2.0.17",
     "electron-builder": "~20.34",
     "electron-devtools-installer": "^2.2.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -162,7 +162,7 @@
     "chance": "^1.0.13",
     "concurrently": "3.5.1",
     "dotenv": "^5.0.1",
-    "electron": "^2.0.17",
+    "electron": "1.8.8",
```
MortalKastor: This would break Mac version (and maybe Windows version).
A temporary fix/hack would be to conditionally downgrade `electron` for Linux in a postinstall script.

gre: or even better, but only for linux

gre:
Capture d’écran 2019-06-06 à 20 18 55

false positive?

should use `set -e` ?

PR #2066, opened 2019-06-06
Tokens block for account page

juan-cortes:
image

### Type

Feature

Will need to revisit the structure as we integrate more, not totally convinced right now. Also lacks some ui polishes.

### DIFF
--- a/src/components/AccountPage/TokensList.js
+++ b/src/components/AccountPage/TokensList.js
@@ -1,0 +1,78 @@
+// @flow
+
+import React, { PureComponent } from 'react'
+import type { PortfolioRange } from '@ledgerhq/live-common/lib/types/portfolio'
+import styled from 'styled-components'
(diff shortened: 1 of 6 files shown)


src/components/AccountPage/TokensList.js
```diff
@@ -0,0 +1,78 @@
...
+            {t('tokensList.title')}
+          </Text>
+          <Button small primary onClick={() => null}>
+            <Box horizontal flow={1} alignItems="center">
+              <IconPlus size={12} />
+              <Box>{t('tokensList.cta')}</Box>
+            </Box>
+          </Button>
+        </Wrapper>
+        {account.tokenAccounts &&
```
gre: This shouldn't be required with the first return null.
If it is, you can still put it in a const before the if to prove to Flow

PR #2067, opened 2019-06-07
Rework ordering

gre:
- https://github.com/LedgerHQ/ledger-live-common/releases/tag/v6.2.0
- ordering is how working across token accounts.
- the order list looks like this:
Capture d’écran 2019-06-07 à 10 39 06


gre:
mmh why all these commit :hodl:

juan-cortes:
> mmh why all these commit :hodl:

You branched from the wrong branch?

PR #2068, opened 2019-06-07
Breadcrumbs functionality for Accounts

juan-cortes:
image

### Type
Feature

Also updates the token icon, forces refresh of the account page on navigation, decreases the animation of tokens (remove it altogether maybe?), and fixes some alignment issues on the rows.

> @gre I didn't manage to fix the issue I mentioned on Slack about the nested token row left border. It'd be great if you can take a look

### DIFF
--- a/src/components/AccountPage/index.js
+++ b/src/components/AccountPage/index.js
@@ -92,7 +92,7 @@ class AccountPage extends PureComponent<Props> {
     const color = getCurrencyColor(currency)
 
     return (
-      <Box>
+      <Box key={account.id}>
         <TrackPage
           category="Account"
           currency={currency.id}
--- a/src/components/AccountsPage/AccountRowItem/Header.js
+++ b/src/components/AccountsPage/AccountRowItem/Header.js
@@ -17,8 +17,7 @@ type Props = {
 // NB Inside Head to not break alignment with parent row;
 const NestedIndicator = styled.div`
   border-left: 1px solid ${p => p.theme.colors.lightFog};
-  min-height: 40px;
-  height: 100%;
+  height: 44px;
   margin-left: 9px;
   padding-left: 5px;
 `
@@ -44,7 +43,7 @@ class Header extends PureComponent<Props> {
         horizontal
         ff="Open Sans|SemiBold"
         flow={3}
-        flex={`${nested ? 30 : 42}%`}
+        flex={`${nested ? 42 : 30}%`}
         pr={1}
         alignItems="center"
       >
@@ -53,11 +52,12 @@ class Header extends PureComponent<Props> {
           <CryptoCurrencyIcon currency={currency} size={20} />
         </Box>
         <Box grow>
-          {nested && (
-            <Box style={{ textTransform: 'uppercase' }} fontSize={9} color="grey">
-              {title}
-            </Box>
-          )}
+          {!nested &&
+            account.type === 'Account' && (
+              <Box style={{ textTransform: 'uppercase' }} fontSize={9} color="grey">
+                {title}
+              </Box>
+            )}
           <Ellipsis fontSize={12} color="dark">
             {name}
           </Ellipsis>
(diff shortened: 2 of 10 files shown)


src/components/TopBar/Breadcrumb.js
```diff
@@ -0,0 +1,104 @@
...
+class Breadcrumb extends PureComponent<Props> {
+  processItemsForDropdown = (items: any[]) =>
+    items.map(item => ({ key: item.id, label: item.id, account: item }))
+
+  render() {
+    const { t, push, pathname, accounts } = this.props
+    let tokenItems
+    let token
+
+    if (!pathname.startsWith('/account/')) return null
```
gre: I will suggest a rework of architecture of this component. we should use React Router.

suggest to have this move to:

```
components/TopBar/Breadcumb/index.js
```

and only doing the minimal routing needed.

look what's done here:

https://github.com/LedgerHQ/ledger-live-desktop/blob/topic/erc20/src/components/layout/Default.js#L120-L129

same logic can be done with this:

you can have a top level:

```js
                  <Switch>
                    <Route path="/account/:parentId/:id" component={AccountWithTokenBroadcrumb} />
                    <Route path="/account/:id" component={AccountBroadcrumb} />
                  </Switch>
```

or share the same component is possible too

```js

                  <Switch>
                    <Route path="/account/:parentId/:id" component={AccountBroadcrumb} />
                    <Route path="/account/:id" component={AccountBroadcrumb} />
                  </Switch>
```

but at least that would avoid all the ifs.

only that `AccountBroadcrumb` would then get Connected. 

could be done in this file:

```
components/TopBar/Breadcumb/Account.js
```

and you can actually take the same connect logic of https://github.com/LedgerHQ/ledger-live-desktop/blob/topic/erc20/src/components/AccountPage/index.js#L33 (that could be interesting to split into a dedicated selector that is basically this:

```js
const accountAndParentAccountSelector = (state, {
    match: {
      params: { id, parentId },
    },
  },
) => {
  const parentAccount: ?Account = parentId && accountSelector(state, { accountId: parentId })
  let account: ?(TokenAccount | Account)
  if (parentAccount) {
    const { tokenAccounts } = parentAccount
    if (tokenAccounts) {
      account = tokenAccounts.find(t => t.id === id)
    }
  } else {
    account = accountSelector(state, { accountId: id })
  }
}
```

src/components/TopBar/Breadcrumb.js
```diff
@@ -0,0 +1,104 @@
...
+      const tokens = account.tokenAccounts
+      tokenItems = this.processItemsForDropdown(tokens)
+      token = tokens.find(t => t.id === tokenId)
+    }
+
+    return (
+      <Wrapper>
+        <Button onClick={() => this.props.push('/accounts/')}>{t('accounts.title')}</Button>
+        <Separator />
+        <Crumb push={push} items={accountsItems} account={account} />
```
gre: i would suggest we directly pass in the accounts, not sure why we need to map them with label and key from this component point of view.

src/components/TopBar/Crumb.js
```diff
@@ -0,0 +1,150 @@
...
+`
+
+const Check = styled.div`
+  color: ${p => p.theme.colors.wallet};
+  flex: 1;
+  text-align: right;
+  margin-left: 20px;
+`
+
+class Crumb extends PureComponent<Props> {
```
gre: in the current implementation, since things are `.map(` ed by the parent, there are no optim of this Pure. it will always re-render

src/components/TopBar/Breadcrumb.js
```diff
@@ -0,0 +1,104 @@
...
+    items.map(item => ({ key: item.id, label: item.id, account: item }))
+
+  render() {
+    const { t, push, pathname, accounts } = this.props
+    let tokenItems
+    let token
+
+    if (!pathname.startsWith('/account/')) return null
+
+    const [, , accountId, tokenId] = pathname.split('/')
```
gre: mmh the ordering is weird. `accountId` and then `tokenId`.
it should be `parentAccountId` and `accountId` and then you make a `parentAccount` and `account` (there are no token)

at least, this is the choice we did in other places so we need to be consistent.

src/components/TopBar/Breadcrumb.js
juan-cortes: I reused the dropdown component, it seemed to want this, I'll look again

PR #2078, opened 2019-06-09
Prevent drag action on topbar buttons/elements

juan-cortes:
Screenshot 2019-06-09 at 01 59 27

The green area (without the cutouts) is the current dragable area for the window, at least on mac. This means that we can get some weird behaviour when dragging the window when the event starts on top of  a button/dropdown from topbar.

This pr makes the area exclude the cutouts.

### Type

UX Polish


gre:
lgtm

PR #2081, opened 2019-06-10
Search for list mode on accounts page

juan-cortes:
Jun-10-2019 14-36-50

Allow tokens to be matched when in list mode in accounts page, making the parent row inactive but still showing the tokens that match.

### Type

Feature, UI Polish

### DIFF
--- a/src/components/AccountsPage/AccountList/index.js
+++ b/src/components/AccountsPage/AccountList/index.js
@@ -28,16 +28,26 @@ const BodyByMode = {
   list: ListBody,
 }
 
-const matchesSearch = (search: string, account: Account | TokenAccount): boolean =>
-  !search ||
-  (account.type === 'Account'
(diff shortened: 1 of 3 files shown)


src/components/AccountsPage/AccountList/index.js
```diff
@@ -28,16 +28,26 @@ const BodyByMode = {
...
-      }`
-    : `${account.token.ticker}|${account.token.name}`
-  )
-    .toLowerCase()
-    .includes(search.toLowerCase())
+export const matchesSearch = (
+  search?: string,
+  account: Account | TokenAccount,
+  subMatch: boolean = false,
+): boolean => {
```
gre: great! can you later add it in live-common?

juan-cortes: sure :)

PR #2082, opened 2019-06-12
LL-1446 Rework Settings for Rates

juan-cortes:
> ~I was going to make this a PR but it's going to remain as a draft. I'm not sure anymore the price should be shown using the countervalue or the intermediary value. If it's the intermediary value, then we need to rework the price component too, because it's not ready for that right now.~

**Note** I couldn't identify why some coins are not remembering the exchange selected. Please @gre take a look if you can.

image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1446

The application settings section for currencies is now called "Crypto assets" and includes a way to set an exchange for each pair we have.


gre:
yeah Price need to be rework because we don't show the final rate but the intermediary raw rate.
if it's not easy to generalize, it's better to make another component like Price and maybe we can just factorize the rendering part.

src/components/SettingsPage/sections/Currencies.js
```diff
@@ -1,68 +1,208 @@
...
+              exchangeId={counterValueExchange}
             />
-          }
-        />
-        <Body>
-          <CurrencyRows currency={currency} />
-        </Body>
-      </Section>
+            {currencies.filter(c => c !== intermediaryCurrency).map(from => {
+              const cs = currenciesSettings[from.id]
```
gre: i'll fix it but we should not directly use `currenciesSettings[]` there are selector for that, and in fact that's where the bug come from: it's not from.id but it's from.ticker (today!)

PR #2089, opened 2019-06-17
Add failsafe for usb to reset after 5s without device

gre:
In some system like Windows, USB events can be wrong and we miss the "connect" event that happens when you switch apps on the device. That makes Ledger Live being stuck in first step "connect your device". This also seems to reproduce more often with the Nano X (because disconnect/connect events are faster?!).

At the same time as we're about to deliver a better implementation under Experimental USB, this should fix the problem by a workaround that restarts the device listing after 5s of non activity & no device was detected.

### Type

Bug fix

### Context

LL-1537

- there should be absolutely no usb behavior changes for when there were no problem with USB (typically on mac)
- for those computers/hardware having problems, you should get unstuck from the stuck situation after 5s


meriadec:
dammit prettier

PR #2090, opened 2019-06-17
LL-1441 added fees counter values to operation details

IAmMorrow:
### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/projects/LL/issues/LL-1441

Operation Details view / Counter values

image

### DIFF
--- a/src/components/base/FormattedVal/index.js
+++ b/src/components/base/FormattedVal/index.js
@@ -55,6 +55,7 @@ type OwnProps = {
   animateTicker?: boolean,
   disableRounding?: boolean,
   isPercent?: boolean,
+  subMagnitude?: number,
 }
 
 const mapStateToProps = (state: State, _props: OwnProps) => ({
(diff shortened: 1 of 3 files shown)


src/components/base/FormattedVal/index.js
```diff
@@ -143,4 +146,8 @@ function FormattedVal(props: Props) {
   )
 }
 
+FormattedVal.defaultProps = {
+  subMagnitude: 0,
```
MortalKastor: Is this useful?
`FormattedVal` already defaults `subMagnitude` to `0` IIRC

IAmMorrow: If i don't set a default value at the component level, some use cases will forward `subMagnitude: undefined` to the helper func (formatCurrencyUnit) which will throw an error. The function's default values are set by an object merge which will not omit undefined values. This can be fixed by omitting the input's undefined values before merging with the defaults.

IAmMorrow: (previously, `formatCurrencyUnit` wasn't using `subMagnitude` as a parameter in this context but i had to add it for this use case)

PR #2092, opened 2019-06-17
Show half of the last row if scrollable

juan-cortes:
Until we have another solution for scrollbars, the display of dropdown selectors will attempt to show half of the last row to indicate there are more items.

image

### Type

UI Polish

 ### Context
image


gre:
@juan-cortes eslint errors

juan-cortes:
> @juan-cortes eslint errors

🤷‍♂

PR #2093, opened 2019-06-17
New experimental USB

gre:
this new Experimental USB implementation uses a new transport that keeps reusing Transport instance and hook to disconnect.


gre:
that PR will deserve more work. we'll disable experimental usb in next release for now.

gre:
@Arnaud97234 i think this is now ready to go.

PR #2096, opened 2019-06-18
LL-1527 Use maxAmount for high fee warning calculation when using send max feature.

juan-cortes:
### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-1527

Send flow, send max

### DIFF
--- a/src/components/modals/Send/steps/01-step-amount.js
+++ b/src/components/modals/Send/steps/01-step-amount.js
@@ -113,6 +113,7 @@ export class StepAmountFooter extends PureComponent<
   StepProps<*>,
   {
     totalSpent: BigNumber,
+    maxAmount: BigNumber,
     canNext: boolean,
     isSyncing: boolean,
     highFeesOpen: boolean,
@@ -122,6 +123,7 @@ export class StepAmountFooter extends PureComponent<
     isSyncing: false,
     highFeesOpen: false,
     totalSpent: BigNumber(0),
+    maxAmount: BigNumber(0),
     canNext: false,
   }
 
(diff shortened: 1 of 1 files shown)


gre:
@aramab you said yesterday we still want the 10% protection even if you use max?
this commit seems to totally disable the modal in USE MAX case, if we want the 10% warning modal even in max case, we will have to use the calculated max instead of the `t.amount` 🤔

juan-cortes:
What I implemented is disabling the modal if in usemax mode, I was pretty sure that's what we agreed on. If not, I'll change it sorry.

aramab:
@gre @juan-cortes  Yes, 10% protection even for send max

juan-cortes:
Ok, I will change it

gre:
btw is max amount calculation even working for eth? i always had the value of 0.00 in my field

src/components/modals/Send/steps/01-step-amount.js
```diff
@@ -171,28 +173,38 @@ export class StepAmountFooter extends PureComponent<
...
+          totalSpent
+            .minus(transaction.amount)
+            .times(10)
+            .gt(transaction.amount)) ||
+          (maxAmount &&
+            maxAmount.gt(0) &&
+            totalSpent
+              .minus(maxAmount)
+              .times(10)
+              .gt(maxAmount)))
```
gre: sorry to not ask that before, but could you make that whole logic into a function that we will put in live-common? we can only later refactor that part but at least so it's getting prepared there as well..

Future thought: what if it actually would be part of the bridge into a helper function like `async getTransactionStatus(account, transaction): { expensiveFees: bool, fees: BigNumber, amount: BigNumber }` I'm sure we could put all stuff we need into one big status object like this so we can share some code too 🤔  

Asking that because we will need same for mobile soon..

juan-cortes:
Sure, I can refactor that tonight / tomorrow morning

gre:
1. bech32 bug is irrelevant in this PR as it's happening basically since last bump of libcore and we're waiting for a bech32 fixes. (LLC-276)
2. i'm not sure what we should do as discussed on Slack because the dependency on fields. splitting this form into two would help solving this, in meantime we can make errors to not throw if recipient is empty, or something.

juan-cortes:
don't even know the state this is in anymore

juan-cortes:
Since the logic was moved to common, closing this one and opening another PR using that

PR #2105, opened 2019-06-18
LL-1530 Persisting (until relaunch) state for expand/collapsed rows

juan-cortes:
> Had to remove the animation on open/close for now.

### Type

UX Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1530

Accounts section, rows of ethereum accounts with tokens. See that when navigating away and coming back, the expand/collapse state is now persisted until we relaunch the app.

### DIFF
--- a/src/components/AccountsPage/AccountRowItem/index.js
+++ b/src/components/AccountsPage/AccountRowItem/index.js
@@ -104,15 +104,27 @@ const mapDispatchToProps = {
   openModal,
 }
 
+const expandedStates: { [key: string]: boolean } = {}
+
 class AccountRowItem extends PureComponent<Props, State> {
-  state = {
-    expanded: false,
+  constructor(props) {
+    super(props)
+    const { account, parentAccount } = this.props
+    const accountId = parentAccount ? parentAccount.id : account.id
+
+    this.state = {
+      expanded: expandedStates[accountId],
+    }
   }
 
-  componentWillReceiveProps(nextProps: Props) {
-    this.setState({
-      expanded: !!nextProps.search,
-    })
+  static getDerivedStateFromProps(nextProps: Props) {
+    const { account } = nextProps
+    if (account.tokenAccounts) {
+      return {
+        expanded: expandedStates[account.id] || !!nextProps.search,
+      }
+    }
+    return null
   }
 
   onClick = () => {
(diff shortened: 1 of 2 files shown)


src/components/AccountsPage/AccountRowItem/index.js
```diff
@@ -104,14 +104,25 @@ const mapDispatchToProps = {
...
 
+const expandedStates: { [key: string]: boolean } = {}
+
 class AccountRowItem extends PureComponent<Props, State> {
-  state = {
-    expanded: false,
+  constructor(props) {
+    super(props)
+    const { account, parentAccount } = this.props
+    const accountId = parentAccount ? parentAccount.id : account.id
```
gre: why is there a need for this, aren't expand state only used for the parent case anyway, so basically just account.id case is needed?

src/components/AccountsPage/AccountRowItem/index.js
```diff
@@ -104,14 +104,25 @@ const mapDispatchToProps = {
...
+    }
   }
 
   componentWillReceiveProps(nextProps: Props) {
+    const { account, parentAccount } = nextProps
+    const accountId = parentAccount ? parentAccount.id : account.id
+
     this.setState({
-      expanded: !!nextProps.search,
+      expanded: expandedStates[accountId] || !!nextProps.search,
```
gre: so same here, shouldn't we just don't do anything in case account is not an expandable?

src/components/ParentCryptoCurrencyIcon.js
```diff
@@ -1,21 +1,21 @@
 // @flow
 import React, { PureComponent } from 'react'
 import styled from 'styled-components'
-import get from "lodash/get"
+import get from 'lodash/get'
 import type { Currency } from '@ledgerhq/live-common/lib/types'
-import CryptoCurrencyIcon from "./CryptoCurrencyIcon"
+import CryptoCurrencyIcon from './CryptoCurrencyIcon'
```
gre: this is in fact correct changes.

@IAmMorrow please make sure to configure your editor to prettier at save.

gre:
@juan-cortes could you rebase with develop when addressing my feedbacks? thx

src/components/AccountsPage/AccountRowItem/index.js
juan-cortes: uhmm I'll check, you are probably right, I assumed without a state change it wouldn't trigger a render @gre

juan-cortes:
@gre

PR #2108, opened 2019-06-19
LL-1504 fixed delta change

IAmMorrow:
Hiding delta value when > 9999% and when account has been empty for the selected period.
Calculating delta value with the first non-zero value of the period.

image

### Type

Bug Fix

### Context

LL-1504

Portfolio and Account page

### DIFF
--- a/src/components/ParentCryptoCurrencyIcon.js
+++ b/src/components/ParentCryptoCurrencyIcon.js
@@ -7,7 +7,6 @@ import CryptoCurrencyIcon from './CryptoCurrencyIcon'
 
 type Props = {
   currency: Currency,
-  parentCurrency?: Currency,
   borderColor?: string,
 }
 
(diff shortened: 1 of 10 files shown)


src/components/ParentCryptoCurrencyIcon.js
```diff
@@ -21,15 +20,18 @@ const ParentCryptoCurrencyIconWrapper = styled.div`
...
-    const double = !!parentCurrency
+    const { currency, borderColor } = this.props
 
-    return (
-      <ParentCryptoCurrencyIconWrapper borderColor={borderColor}>
-        {parentCurrency && <CryptoCurrencyIcon currency={parentCurrency} size={double ? 16 : 20} />}
-        <CryptoCurrencyIcon currency={currency} size={double ? 16 : 20} />
-      </ParentCryptoCurrencyIconWrapper>
-    )
+    if (currency.type === 'TokenCurrency' && currency.parentCurrency) {
```
gre: the second condition is not necessary. only the `currency.type` will help flow.

gre:
i'm going to send a new PR that have moved a lot of % calculation logic in live-common and will have less code in desktop & in order to prepare for mobile too

PR #2109, opened 2019-06-19
display infos when currency app is not supported by ledger live

valpinkman:
show a Learn More link when installing a cryptocyurrency app that is not supported by ledger live

Also reworked the whole modal when installing / uninstalling apps

### Type

UI Improvement

### Context

LL-1482

Install supported app

image

Install unsupported app

image

Uninstall app

image

@khalilbenihoud @bscholving @dekkaki are the design ok witchu ?

### DIFF
--- a/src/components/ManagerPage/AppsList.js
+++ b/src/components/ManagerPage/AppsList.js
@@ -5,6 +5,8 @@ import React, { PureComponent } from 'react'
 import styled from 'styled-components'
 import { translate } from 'react-i18next'
 import { compose } from 'redux'
+import get from 'lodash/get'
+import { getDeviceModel } from '@ledgerhq/devices'
 
 import type { Device, T } from 'types/common'
 import type { ApplicationVersion, DeviceInfo } from '@ledgerhq/live-common/lib/types/manager'
@@ -12,7 +14,9 @@ import type { CryptoCurrency } from '@ledgerhq/live-common/lib/types/currencies'
 import manager from '@ledgerhq/live-common/lib/manager'
 import { getEnv } from '@ledgerhq/live-common/lib/env'
 import { getFullListSortedCryptoCurrencies } from 'helpers/countervalues'
-import { listCryptoCurrencies } from 'config/cryptocurrencies'
+import { openURL } from 'helpers/linking'
+import { listCryptoCurrencies, supported } from 'config/cryptocurrencies'
+import { urls } from 'config/urls'
 import installApp from 'commands/installApp'
 import uninstallApp from 'commands/uninstallApp'
 import flushDevice from 'commands/flushDevice'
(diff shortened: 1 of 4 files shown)


src/components/ManagerPage/AppsList.js
```diff
@@ -227,15 +268,20 @@ class AppsList extends PureComponent<Props, State> {
           </Box>
         )}
         <Text ff="Museo Sans|Regular" fontSize={6} color="dark">
-          {mode !== 'home' ? t(`manager.apps.${mode}`, { app }) : null}
+          {mode !== 'home' ? t(`manager.apps.${mode}`, { app: get(app, 'name', '') }) : null}
```
gre: I prefer we don't use `get(`. it is not typesafe. just use `app.name||""` (why is this needed?)

if app can be falsy, it's probably better to add the condition in the if so we never would render the wording with a falsy case 🤔

src/components/ManagerPage/AppsList.js
```diff
@@ -227,15 +268,20 @@ class AppsList extends PureComponent<Props, State> {
...
         </Box>
       </Box>
     ) : status === 'error' ? (
       <Box>
-        <TrackPage category="Manager" name="Error Modal" error={error && error.name} app={app} />
+        <TrackPage
+          category="Manager"
+          name="Error Modal"
+          error={error && error.name}
+          app={get(app, 'name', '')}
```
gre: same

src/components/ManagerPage/AppsList.js
```diff
@@ -264,9 +310,12 @@ class AppsList extends PureComponent<Props, State> {
...
     ) : status === 'success' ? (
       <Box grow align="center" justify="center" mt={5}>
-        <Box color="positiveGreen">
-          <CheckCircle size={44} />
-        </Box>
+        <IconWrapper>
+          <AppCheck>
+            <CheckFull size={24} />
+          </AppCheck>
+          <AppIcon src={this.getIconUrl()} />
```
gre: live-common already exposes a method for that, please use it:
https://github.com/LedgerHQ/ledger-live-common/blob/master/src/manager.js#L32

src/components/ManagerPage/AppsList.js
```diff
@@ -275,17 +324,45 @@ class AppsList extends PureComponent<Props, State> {
...
+            color="smoke"
+            mt={2}
+            fontSize={4}
+            ff="Open Sans|Regular"
+            textAlign="center"
+            style={{ maxWidth: 350 }}
+          >
+            {t(`manager.apps.${isSupported ? 'supportedCurrency' : 'unsupportedCurrency'}`, {
+              app: app.name,
+              device: get(getDeviceModel(device.modelId), 'productName', ''),
```
gre: same

src/components/ManagerPage/AppsList.js
```diff
@@ -211,13 +238,27 @@ class AppsList extends PureComponent<Props, State> {
...
+    const icon = app ? ICONS_FALLBACK[app.icon] || app.icon : ''
+    return `https://api.ledgerwallet.com/update/assets/icons/${icon}`
+  }
+
   renderBody = () => {
-    const { t } = this.props
+    const { t, device } = this.props
     const { app, status, error, mode, progress } = this.state
 
+    const isSupported = supported.includes(get(app, 'name', '').toLowerCase())
```
gre: same as comment below, we use this logic too often

src/components/ManagerPage/AppsList.js
```diff
@@ -275,17 +324,45 @@ class AppsList extends PureComponent<Props, State> {
...
   }
 
   renderFooter = () => {
     const { t } = this.props
-    return (
+    const { app, mode } = this.state
+
+    const isInstalling = mode === 'installing'
+    const isCurrency = app && app.currency
+    const isSupported = app && supported.includes(app.name.toLowerCase())
```
gre: this does not seem correct to do `app.name.toLowerCase()` here. app name does not match app names but instead should match `app.currency.id`

so basically we should do this

```
isSupported = app && app.currency && isCurrencySupported(app.currency)
```

and probably introduce this `isCurrencySupported` instead of exporting `supported`

gre:
@dasilvarosa it's just a bug in the implementation

valpinkman:
> 1. Please change wording "Monero successfully installed" to: "Learn how to use Monero"

@dasilvarosa wording update
image

@gre @Arnaud97234 Fix implem to check on currency.id
image

image

PR #2111, opened 2019-06-19
Avoid showing placeholder for ETH account with tokens

juan-cortes:
image

For accounts that don't have ETH balance but do have token accounts with balance we are now able to access them without showing the placeholder.

### Type

Bug fix


juan-cortes:
Closed in favor of https://github.com/LedgerHQ/ledger-live-common/pull/284

PR #2113, opened 2019-06-20
LL-1502 fixed distribution

IAmMorrow:
- Assets with zero sum value aren't included in the distribution view anymore.
- non-zero Assets with no countervalue available will remain in the distribution screen with placeholders replacing the missing data.

image

### Type

Bug fix

### Context

LL-1502

Portfolio distribution card


gre:
@Arnaud97234 i've merged develop in so we should have all in this PR.

Issue #2088, opened 2019-06-15
Cannot send XST (Stealth) from Ledger.  Six months now.


myselfmyself:
Guess I'll continue posting this issue until it's resolved.

Still not resolved.  First presented the problem at the end of December 2018.  

No resolution from Ledger Support.  No resolution from Tweeting.

XST developers have already replied to me and said they repaired the issue on their end and they're waiting on Ledger to implement.

gre:
Please stick to the issue #1933 
Fix is getting worked on.

PR #2115, opened 2019-06-20
Bump live-common with fixes

gre:
details is in https://github.com/LedgerHQ/ledger-live-common/releases/tag/v6.10.0

- XST to test with lates PROVIDER=4 app for #2088 
- distribution will show the coins that also don't have countervalue but not those who sum to zero
- isAccountEmpty is taking tokenAccounts into account so it will not show the placeholder if tokens are present & also in the scan accounts logic
- in the logs, there will still be libcore logs but under a different "type" convention. you will also see logs about "getAddress"
- LL-1524 add NDAX to partners list


gre:
1. is fixed on develop
2. will be provided by @juan-cortes  in another PR, we'll merge on develop now

PR #2129, opened 2019-06-25
LL-1573 Connected the rate graphs with the user time range selection

juan-cortes:
image

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-1573

Rate graphs,  the data should correspond to the selected time range in the portfolio

### DIFF
--- a/src/components/SettingsPage/sections/CryptoAssets/RateRow/PriceGraph.js
+++ b/src/components/SettingsPage/sections/CryptoAssets/RateRow/PriceGraph.js
@@ -8,15 +8,15 @@ import CounterValues from 'helpers/countervalues'
 import { PlaceholderLine } from 'components/Placeholder'
 import { colors } from 'styles/theme'
 import { rgba } from 'styles/helpers'
+import { getDates } from '@ledgerhq/live-common/lib/portfolio'
 
-const DAY = 24 * 60 * 60 * 1000
 const mapStateToProps = (state, props: *) => {
+  const dates = getDates(props.timeRange)
   const data = []
-  let t = Date.now() - props.days * DAY
   const value = BigNumber(10 ** props.from.units[0].magnitude)
   let nbCounterValueOff = 0
-  for (let i = 0; i < props.days; i++) {
-    const date = new Date(t)
+  for (let i = 0; i < dates.length; i++) {
+    const date = dates[i]
     const cv = CounterValues.calculateSelector(state, {
       ...props,
       date,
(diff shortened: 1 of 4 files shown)


src/components/SettingsPage/sections/CryptoAssets/RateRow/PriceGraph.js
```diff
@@ -8,15 +8,17 @@ import CounterValues from 'helpers/countervalues'
 import { PlaceholderLine } from 'components/Placeholder'
 import { colors } from 'styles/theme'
 import { rgba } from 'styles/helpers'
+import { getDates } from '@ledgerhq/live-common/lib/portfolio'
+import { timeRangeDaysByKey } from '../../../../../reducers/settings'
 
-const DAY = 24 * 60 * 60 * 1000
 const mapStateToProps = (state, props: *) => {
+  const dates = getDates(props.timeRange)
+  const days = timeRangeDaysByKey[props.timeRange]
```
gre:
```suggestion
  const days = dates.length
```

isn't it the same? (in fact it shouldn't be named days, because might not be days in the future)
you can probably drop this param and just uses `dates.length` when it was used (only once for isAvailable?)

src/components/SettingsPage/sections/CryptoAssets/Rates.js
```diff
@@ -82,25 +87,31 @@ class Rates extends PureComponent<Props> {
...
             <Box ff="Open Sans|SemiBold" color="dark" fontSize={4}>
-              {'Price'}
+              <Trans i18nKey="settings.rates.rpice" />
             </Box>
             <Box ff="Open Sans|SemiBold" color="dark" fontSize={4}>
-              {'Last 30 days'}
+              <Trans i18nKey={`settings.rates.last`} values={{ range }} />
             </Box>
             <Box ff="Open Sans|SemiBold" color="dark" fontSize={4}>
-              {'Exchange'}
```
gre: good catch on these untranslated

src/components/SettingsPage/sections/CryptoAssets/Rates.js
```diff
@@ -82,25 +87,31 @@ class Rates extends PureComponent<Props> {
...
                 </Tooltip>
               </TooltipButtonWrapper>
             </Box>
             <Box ff="Open Sans|SemiBold" color="dark" fontSize={4}>
-              {'Price'}
+              <Trans i18nKey="settings.rates.rpice" />
             </Box>
             <Box ff="Open Sans|SemiBold" color="dark" fontSize={4}>
-              {'Last 30 days'}
+              <Trans i18nKey={`settings.rates.last`} values={{ range }} />
```
gre: mmh, i think we already have wording for the ranges. Can we use it?
also, the problem is you are not translating `range` here

juan-cortes: I'm not translating the range, because in that case `range` is the number of days. Probably not the best var name. The literal is `Last {{range}} days`

src/components/SettingsPage/sections/CryptoAssets/RateRow/PriceGraph.js
juan-cortes: Indeed

Issue #1849, opened 2019-03-10
Bug: Error calculating fees


filcardoso:
Steps to reproduce:

Create account in Ledger Live and sent 0.00010000 from bitcoin wallet. 
After a few moments, and after fees, the amount of 0.00009604 arrived to my Ledger Live wallet.

Then I tried to send back what I received, and got the following:

- after selecting 0.00003900 to send, gives a total to debit of 0.00004126 (fee: 226 sats)

- after selecting 0.00003940 to send, gives a total to debit of 0.00004166 (fee: 226 sats)

- after selecting 0.00003950 to send, gives a total to debit of 0.00009604 (fee: 5654 sats) --> My Full Balance!

The fee is 25x higher just for sending 10 extra sats.
Also, the fees are more than the actual value i'm trying to send.

gre:
Thanks for the issue, we'll try to reproduce on our end and investigate what is wrong.

Can you confirm that you are on Ledger Live 1.5.0 ?

thanks

filcardoso:
just updated, and the problem persists.

Thank you.

gre:
Can you attempt a clear cache after this update? Thanks

filcardoso:
I cleared the cache.(went to settings -> hit Clear -> wait a few seconds -> input my password to login -> and tried again)

The same happened.

I have another account with 0.01298558.

If I insert the amount 0.0129008, the total to debit becomes 0.01290454 (fee 374 sats)

If I put 1 more sat (0.01290081), the total to debit becomes my full balance 0.01298558 (fee 8477 sats)

filcardoso:
Just tried it on the newly released Android app with support to Nano S, and the calculation is exactly the same as I described in the previous posts.

gre:
@filcardoso can you check if the bug is still present on LL 1.7.0? (probably do a clear cache to be sure)

Thanks

filcardoso:
Short answer: Bug is still present.

Long answer:
I cleared the cache (settings -> help -> Clear Cache)

Strange thing happened after. I input the destination address, and the following error appeared in the amount field:  "Account 1, for wallet '<random numbers>_bitcoin_cash': doesn't exist"

I changed account and the other gave me this error, also in the amount field:
"There is no UTXO on this account."

I figured its something to do with clearing the cache.

I hit Clear Cache again, and it stopped happening.

But the original bug calculating the fee is still present.

filcardoso:
Hi @gre !
[...]

PR #2130, opened 2019-06-25
live-common@7 ledger-core@3

gre:
this PR brings lot of fixes from libcore.

- Native Segwit is fixed and is in general availability :)
- Ethereum have validation on "not enough balance"
- ERC20 have validation on "not enough balance"
- ERC20 balances are correct!
- Fixes https://github.com/LedgerHQ/ledger-live-desktop/issues/1849
- self transactions of ERC20 to make two operations
- ERC20 have validation on "not enough gas"
- block gas limit 0
- Correctly showing alert modal on fees > 10% when using send max (https://ledgerhq.atlassian.net/browse/LL-1527)
- factorize in live-common the market cap sorting logic. plz check work related to listing currencies and apps & make sure we also still display the "you successfully install apps of coin X"


gre:
@Arnaud97234 can you confirm there are no regression & that the PR fixes the things mentioned? thanks

KhalilBellakrid:
For ERC20 self transactions, indeed we provide just one operation (balance will be of though since we use `balanceOf` now) but I will provide a fix for this issue !

gre:
ETC fixed for now by not using it with libcore.

gre:
I don't understand.

gre:
@Arnaud97234 it should be fixed now

gre:
oh wait i need to bump

gre:
@Arnaud97234 if this bug is present on develop, i would assume we can still merge this PR. is it a regression?

Arnaud97234:
@gre The ERC20 transactions are working properly on develop branch in Linux OS so yes it's a regression.

KhalilBellakrid:
This should be fixed by : https://github.com/LedgerHQ/lib-ledger-core/commit/4e94cc1fffac52f4465e6b8ef0ad73869d5414a1

PR #2132, opened 2019-06-25
LL-1546 Made the Parent account name ellipse when needed

juan-cortes:
Jun-25-2019 17-24-41

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1546

Account breadcrumbs that used to overflow, now should behave nicer. Note that the original name of the account is _"There once was a very long account name that overf"_ because I couldn't keep on typing after that many characters.

### DIFF
--- a/src/components/base/DropDown/index.js
+++ b/src/components/base/DropDown/index.js
@@ -40,6 +40,10 @@ export const DropDownItem = styled(Box).attrs({
   white-space: nowrap;
 `
 
+export const Wrapper = styled(Box)`
+  flex-shrink: 1;
+  ${p => p.shrink && `flex-shrink:${p.shrink};`}
+`
 function itemToString(item) {
   return item ? item.label : ''
 }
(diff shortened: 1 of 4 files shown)


src/components/base/DropDown/index.js
```diff
@@ -156,12 +161,12 @@ class DropDown extends PureComponent<Props> {
           selectedItem,
           ...downshiftProps
         }) => (
-          <Box {...getRootProps({ refKey: 'innerRef' })} horizontal relative>
+          <Wrapper shrink={shrink} {...getRootProps({ refKey: 'innerRef' })} horizontal relative>
```
gre: just make sure this changes does not impact any of the other usage of DropDown (plz check for all places of dropdown)

juan-cortes: I've checked, it's only used on the range and sort selector as far as I can tell, they work.

PR #2135, opened 2019-06-26
Fixes ellipsis on account selector

juan-cortes:
image

Since the current custom implementation of the `MenuList` for performance, uses `FixedSizeList` we can't got to two lines on selector rows (it reuses them). There is a `DynamicSizeList` that I later found out about so perhaps that's a good future task. For now, this fixes the ellipsis which wasn't working on the account selector.

### Type

UI Polish

### Context

There was a `//  FIXME` from @meriadec  so I fixed it.


juan-cortes:
@bscholving  approves

PR #2136, opened 2019-06-26
LL-1594 Hide clock icon on failed transactions

juan-cortes:
To make sure we are on the same page here @gre this will not show the confirmation clock on a failed transaction regardless of the confirmation status. So, if the circle is red, there's no clock.

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1594

Operation list, failed transactions should show no confirmation status indicator


gre:
totally. and it works conceptually too: a pending case is one that is not successful or failed. like a Promise ;)

PR #2137, opened 2019-06-26
On Live update, we still keep experimental but only those who exists

gre:
good point from @Arnaud97234 , we shouldn't remove all experimental on an update unless we clearly inform in a UI modal in the future, for now we keep the experimentals, we only remove if those disappear from the available experimentals.

This PR also removed Experimental USB that is no longer needed.


gre:
@Arnaud97234  i've fixed the problem

PR #2140, opened 2019-06-27
Fixes gas limit issues

gre:
- gas limit field is clearly visible so you can see when it's loading
- Continue is blocked until gas limit is loaded
- you can still edit the field manually
- if calc fails (e.g. no network) it fallbacks to 21000
- ~~a gas limit of 0 is invalid and blocks Continue~~ => WILL BE FIXED LATER

We will need to backport this in live-common for mobile implementation.

gl

### Type

bug fix

### Context

LL-1599

### Technical notes

gasLimit becomes an optional field. that helps tracking if it was settled or not. either network settles by calculation or the user does it. it get set back to undefined if it needs recalculation. a cache has also been introduced for performance (now it's a backend perf problematic).

### DIFF
--- a/src/bridge/LibcoreBridge.js
+++ b/src/bridge/LibcoreBridge.js
@@ -78,7 +78,7 @@ const createTransaction = () => ({
   useAllAmount: false,
   feePerByte: undefined,
   gasPrice: undefined,
-  gasLimit: BigNumber(0x5208),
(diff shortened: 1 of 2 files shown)


gre:
seen with Arnaud, we'll fix gasLimit=0 in another PR, i'll update descrpition

PR #2141, opened 2019-06-27
order crypto by market cap in add account select

valpinkman:
Make sure the cryptoassets are ordered by market cap in manager and add account's select

### Type

Bug Fix

### Context

LL-1584

image

image

### DIFF
--- a/src/hooks/useCryptoCurrencies.js
+++ b/src/hooks/useCryptoCurrencies.js
@@ -1,6 +1,8 @@
 // @flow
 import type { CryptoCurrency } from '@ledgerhq/live-common/lib/types'
+import { useState, useEffect } from 'react'
 
+import { getFullListSortedCryptoCurrencies } from 'helpers/countervalues'
 import { listCryptoCurrencies } from 'config/cryptocurrencies'
 import useEnv from 'hooks/useEnv'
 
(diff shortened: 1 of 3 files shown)


src/components/GenuineCheck.js
```diff
@@ -109,7 +110,7 @@ class GenuineCheck extends PureComponent<Props, State> {
     // Preload things in parallel
     Promise.all([
       // Step dashboard, we preload the applist before entering manager while we're still doing the genuine check
-      manager.getAppsList(deviceInfo),
+      manager.getAppsList(deviceInfo, getEnv('MANAGER_DEV_MODE')),
```
gre: was this accidentally added? i don't think there is such param and if there is, it's not scope of what this PR fixes?

src/hooks/useCryptoCurrencies.js
```diff
@@ -11,8 +12,15 @@ const useCryptocurrencies = ({
...
   onlySupported?: boolean,
 }): CryptoCurrency[] => {
+  const [cryptos, setCryptos] = useState([])
   const devMode = useEnv('MANAGER_DEV_MODE')
-  const cryptos = listCryptoCurrencies(devMode, onlyTerminated, onlySupported)
+
+  useEffect(() => {
+    getFullListSortedCryptoCurrencies(devMode, onlyTerminated, onlySupported).then(c =>
+      setCryptos(c),
+    )
```
gre: you can directly `.then(setCryptos)`

src/hooks/useCryptoCurrencies.js
```diff
@@ -11,8 +12,15 @@ const useCryptocurrencies = ({
   onlyTerminated?: boolean,
   onlySupported?: boolean,
 }): CryptoCurrency[] => {
+  const [cryptos, setCryptos] = useState([])
```
gre: initial value should be the result of `listCryptoCurrencies` because we don't want to wait network OR in case something is bad with API

PR #2142, opened 2019-06-27
LL-1118 Use user password to encrypt libcore db

MortalKastor:
Up until now, only the local Ledger Live db was encrypted using the optional password. Now that libcore implemented optional encryption with a password, this PR makes Ledger Live use it.

Now libcore needs to be init with the user password from the renderer without relying on env variables (for security/privacy reasons), ~so we end up doing the following IPC dance when Internal process starts 💃~

No more complicated dance, nor Redux store involved: `renderer` thread sends to `main` thread the password when the user enters it, the former keeps it in memory and sends it to `internals` after spawning it.

### Type

Feature
Security/privacy improvement

### Context

https://ledgerhq.atlassian.net/browse/LL-1118

### Parts of the app affected

- Libcore loading: since the password lives on the `renderer` side, `internal` has to send a message to the renderer for it to init libcore with a the password via a command sent back.
- Setting/changing a password: it now additionally affects libcore.
- Unlocking the app: it now also send the password to libcore for decryption.

#### Test plan

- Setting a password in the app should encrypt libcore db
- Disabling password should decrypt libcore db
- Changing the password should also be reflected in libcore db
- Unlocking the app should also unlock libcore
- Updating from an older version of Ledger Live with a password should encrypt libcore db with the password on first unlock

To check the current status of the libcore db encryption, just open the sqlite file with [DB Browser for SQLite](https://sqlitebrowser.org/), and use the defaults encryption settings (SQLCipher 4 defaults)

#### Caveat

Downgrading Ledger Live after migrating to this version while using a password ~will end up in a crash on start~ actually it just breaks accounts sync (that can be fixed by clearing the cache).


src/commands/libcoreChangePassword.js
```diff
@@ -0,0 +1,22 @@
...
+import { withLibcore } from '@ledgerhq/live-common/lib/libcore/access'
+
+type Input = {
+  oldPassword: string,
+  newPassword: string,
+}
+type Result = void
+
+const cmd: Command<Input, Result> = createCommand(
+  'libcoreChangePassword',
```
gre: do you blacklist this command to be logged in order to filter out the logs to contain the input object?

src/commands/libcoreInit.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import { createCommand, Command } from 'helpers/ipc'
+import { Observable } from 'rxjs'
+
+type Input = { password?: string }
+type Result = void
+
+const cmd: Command<Input, Result> = createCommand('libcoreInit', ({ password }) =>
```
gre: same question

src/components/IsUnlocked.js
```diff
@@ -102,23 +109,67 @@ class IsUnlocked extends Component<Props, State> {
...
+      if (unlockedWithoutPW) {
+        // Libcore is indeed unencrypted, so let's encrypt with db password
+        await changeLibcorePassword('', password)
+        // And reload with this new password
+        await setLibcorePassword(password)
+      } else {
+        // Something is wrong, let's set the password and reset libcore, so that
+        // it creates a brand new db encrypted with the current password
+        await setLibcorePassword(password)
+        resetLibcore()
```
gre: this is async, we don't wait it?

src/components/Onboarding/steps/SetPassword.js
```diff
@@ -45,6 +48,7 @@ const INITIAL_STATE = {
 
 type Props = StepProps & {
   saveSettings: any => void,
+  setLibcorePassword: string => void,
```
gre: this is a `string => Promise<bool>` no?

src/components/Onboarding/steps/SetPassword.js
```diff
@@ -61,6 +65,8 @@ class SetPassword extends PureComponent<Props, State> {
     const { nextStep, saveSettings } = this.props
 
     await db.setEncryptionKey('app', 'accounts', newPassword)
+    await changeLibcorePassword('', newPassword)
+    setLibcorePassword(newPassword)
```
gre: which we don't await?

src/helpers/implement-libcore.js
```diff
@@ -1,7 +1,4 @@
 // @flow
-import implementLibcore from '@ledgerhq/live-common/lib/libcore/platforms/nodejs'
-
-implementLibcore({
-  lib: require('@ledgerhq/ledger-core'),
-  dbPath: process.env.LEDGER_LIVE_SQLITE_PATH,
+process.send({
```
gre: semantic of this have changed and makes it a bit weird to understand. can you comment to explain how this works? might work a ascii schema to explain how things works between process

src/components/IsUnlocked.js
MortalKastor: No logic in this part rely on the completion of the function, so I didn't think it was necessary to await for it, but that doesn't seem to be a good practice

src/components/Onboarding/steps/SetPassword.js
MortalKastor: Same as before, I didn't want to keep the user hanging while waiting for libcore to reload to go to the next step :/

src/helpers/implement-libcore.js
MortalKastor: Flow chart added as a comment to this file and in the PR description

src/helpers/reset.js
```diff
@@ -1,21 +1,35 @@
...
 import db from 'helpers/db'
-import { delay } from 'helpers/promise'
+import { delay, promisify } from 'helpers/promise'
 import killInternalProcess from 'commands/killInternalProcess'
 import libcoreReset from 'commands/libcoreReset'
 
-async function resetLibcore() {
+async function deleteSqlite() {
+  const sqlitePath = path.resolve(resolveUserDataDirectory(), 'sqlite')
+  const rm = promisify(rimraf)
```
meriadec: really not important but could be created once in `helpers/fs.js`

gre:
@MortalKastor let's peer program a bit next week to see what we can improve. I think the current flow can be simplified to be more unidirectional (renderer -> main -> internal) a bit like we did with envs, but in a rpc way. Ultimately, I wonder if we couldn't use envs but just no longer make them "process envs" to avoid the visibility problem.

gre:
@MortalKastor :clap: , remain some conflict. When you're done I suggest @juan-cortes could review the code logic and flow.

MortalKastor:
PR description updated + rebase and conflicts fix

gre:
cc @KhalilBellakrid

gre:
this will be revived in another PR

PR #2151, opened 2019-07-01
LL-1560 Stop overriding font styling for inputs

juan-cortes:
There was an global override for inputs receiving the `ff` attribute, meaning all the amount inputs that were supposed to get the Rubik's font were getting Open Sans. This was frustrating design.

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1560

Inputs, numeral inputs should have Rubiks.

### DIFF
--- a/src/components/base/Input/index.js
+++ b/src/components/base/Input/index.js
@@ -64,6 +64,7 @@ const Base = styled.input.attrs({
   fontSize: 4,
 })`
   font-family: 'Open Sans';
+  font-weight: 600;
   ${fontFamily};
   ${fontSize};
   border: 0;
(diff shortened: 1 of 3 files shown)


gre:
did you checked all inputs in our app set a `ff` ?

juan-cortes:
> did you checked all inputs in our app set a `ff` ?

Not really, but if the intention was that all inputs use open sans _unless_ they receive the ff, I can change it to that.

juan-cortes:
Made Open sans the default unless overriden by the instance, so we should be good for testing @gre @Arnaud97234

Arnaud97234:
**Bug found**
Send -> XRP, Tag input is a numeral one so it should be Rubik as well ?

juan-cortes:
Fair enough, I'll update the pr when I'm back home

PR #2157, opened 2019-07-02
Enabling privacy mode where no account balances are shown.

jorge-cob:
Enable/Disable a discreet mode where account balances are hidden/shown

discreetmode1

### Type

Feature

Using context a component can access and toggle the discreet mode.

Send and receive modals are also affected by Discreet Mode:

discreetmode3

At this moment graph is unaffected by Discreet Mode:

discreetmode2

Blurring might be a good alternative to using "*"

### DIFF
--- a/src/components/base/FormattedVal/index.js
+++ b/src/components/base/FormattedVal/index.js
@@ -12,7 +12,8 @@ import type { State } from 'reducers'
 import { formatCurrencyUnit } from '@ledgerhq/live-common/lib/currencies'
 
 import { DISABLE_TICKER_ANIMATION } from 'config/constants'
-import { marketIndicatorSelector, localeSelector } from 'reducers/settings'
+import { marketIndicatorSelector, localeSelector, discreetModeSelector } from 'reducers/settings'
+import { DiscreetModeContext } from 'components/Discreet/DiscreetModeWrapper'
 
 import { getMarketColor } from 'styles/helpers'
 
@@ -61,12 +62,14 @@ type OwnProps = {
 const mapStateToProps = (state: State, _props: OwnProps) => ({
   marketIndicator: marketIndicatorSelector(state),
   locale: localeSelector(state),
+  discreetMode: discreetModeSelector(state),
 })
 
 type Props = OwnProps & {
   marketIndicator: string,
   locale: string,
   ellipsis?: boolean,
+  discreetMode?: boolean,
 }
 
 function FormattedVal(props: Props) {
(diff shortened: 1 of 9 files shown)


gre:
neat! IMO percentages should not be affected by hiding the amounts :)

src/components/base/FormattedVal/index.js
```diff
@@ -115,6 +118,9 @@ function FormattedVal(props: Props) {
     })
   }
 
+  // NB could be extracted to an env var
+  const discreetText = text.replace(/[\d.,]/g, '*')
```
MortalKastor: Replacing each char with a star doesn't really do a good job of hiding if you hold a lot of money: `EUR *******` is still telling that the user is loaded 💰

I would suggest something like this
```suggestion
  const discreetText = text.replace(/[\d.,]+/g, '***')
```

(Also, TIL you don't need to escape dots between brackets inside regexes 😮)

meriadec: yeah always surprised by this :D you don't even need to escape `/` inside brackets!

gre: also, replacing with `.` and `,` is a bit dangerous because it assumes formatter will uses these character which is not guaranteed and shouldn't be assumed (we're pretty careful about that in live-common codebase if i recall correctly). In some languages we might have other things, and for instance we have spaces for thousands. In present case it would work, but with Thibaut solution we would still have the problem of seeing big amounts.

I would suggest this needs to be solved at the *formatter* level, aka in live-common, to provide a way to set a `discreet` as an option there.


So here is some refactoring direction/spec I would suggest:

- live-common: introduces such discreet option to solve it once (think about mobile, we'll need same). https://github.com/LedgerHQ/ledger-live-common/blob/master/src/currencies/formatCurrencyUnit.js see the concept of "fragment" which is part of the formatted value, i think it's as easy as replacing the "value" case by a placeholder there. We could replace it by a `discreetPlaceholder` (which could be a second option, that defaults to `***`)
- FormattedVal could make uses of React hooks, so you don't need to wrap a Consumer and you can directly access the discreet mode value in the JS code too. `useContext(DiscreetModeContext)` https://es.reactjs.org/docs/hooks-reference.html#usecontext

gre:
=> on the graph we should hide the axis
=> `****` for all. to be done in live-common level like commented

juan-cortes:
Closing the pr since we merged  https://github.com/LedgerHQ/lld-v2/pull/109 which is basically this.
Thank you @jorge-cob

PR #2174, opened 2019-07-09
fixes wording on error when not enough space

valpinkman:
During firmware update, if there is not enough space available on the device, fixes the wording from "uninstall some apps" to "uninstall all apps"

### Type

Improvement

### Context

LL-1647

Firmware update

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -971,7 +971,7 @@
     },
     "ManagerNotEnoughSpace": {
       "title": "Sorry, not enough storage left",
-      "description": "Uninstall some apps to make space. This will not affect your crypto assets."
+      "description": "Uninstall all apps to make space. This will not affect your crypto assets."
     },
     "ManagerUninstallBTCDep": {
       "title": "Sorry, this app is required",


gre:
mmh, so i think @juan-cortes was on it (but if not i guess he can forward you the task :p)
and in fact this is not the way we will fix it:

ManagerNotEnoughSpace => only install apps since latest live-common version
ManagerFirmwareNotEnoughSpace => the mcu and firmware install error

so @dasilvarosa have to opportunity to give us two different wordings now. one for "not enough space during an installation of app" and one for firmware.

dasilvarosa:
If it's not too long this would be great: 

ManagerNotEnoughSpace => only install apps since latest live-common version
title: Sorry, not enough storage left
desc: Uninstall some apps to make space. The private keys providing access to your crypto assets will stay safe on your device.

ManagerFirmwareNotEnoughSpace => the mcu and firmware install error
title: Uninstall all apps from your device 
desc: The private keys providing access to your crypto assets will stay safe on your device and should be backed up on your recovery sheet.

gre:
@valpinkman can you bump live-common and integrate? thanks

we'll then let Arnaud can test this (firmware out of space case)

gre:
thanks david, will be fixed as part of https://github.com/LedgerHQ/ledger-live-desktop/pull/2176

PR #2176, opened 2019-07-10
Important bugfixes (USB, erc20 balance, QTUM)

gre:
- Fixes erc20 bad balances. https://github.com/LedgerHQ/ledger-live-common/pull/299
- Fixes some USB issues typically if interrupting add accounts before the end. You should also no longer see a lot of "Cannot open device"
- Fixes QTUM to support segwit (requires last version of QTUM app)
- Fixes wording of firmware update when not enough space. LL-1647
- Better logs for the retry() calls

## Regarding QTUM

This enables QTUM segwit support.
It essentially requires a new version of QTUM nano S app but we still have a minimal breaking changes plan to still make legacy QTUM accounts works with the legacy version of the QTUM app during the transition phase.

This will fix issue with QTUM when you have receive funds via a segwit transaction.

**Minimal breaking changes:**

Is expected to work in old version of QTUM:

- add accounts
  - With the limitation that you can no longer see a "New Account" (because you can't create legacy accounts when segwit is supported in Ledger Live)
  - With the limitation of that you will see some 0x6faa if you later try to do a receive/send, so you need to restart the QTUM app after having added accounts (because we feature detect if segwit works)
- receive
- send

Is not expected to work in old version of QTUM:

- segwit accounts
- ?legacy accounts that have received from segwit


gre:
ok i badly named the error in `@ledgerhq/errors` 🤦‍♂ , bumping...

PR #2182, opened 2019-07-12
LL-1556 Account migration flow

juan-cortes:
Adds the migration flow when necessary. It will show a banner on the portfolio screen if there are eligible accounts to be migrated. On this modal flow we will iterate all different cryptocurrencies and after each loop we will suggest the user changing devices until the migration is completed.

image
image
image
image
image

### Unnecessary ascii diagram
```
+-----------+   +-------------+   +-------------+   +-------------+
|           |   |             |   |             |   |             |
| Overview  ---->   Connect   ---->   Migrate   ----->  Summary   |
|           |   |             |   |             |   |             |
+-----^-----+   +------^------+   +-------------+   +------|------+
      |                |                                   v
      |                |                              Looped all
      |                +-------------------------no---currencies?
    yes                                                    |
      |                                                    |
      +------More migratable------------------------------yes
             accounts?
                |
               no
                |
                v
           +---------+
           |  DONE   |
           +---------+
```

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1556

To test this you will need ideally two devices with both ETH and ETC
that you imported with the experimental libcore OFF. Then you can
enable experimental libcore flag and the banner should appear on top

### Note of warning
I've also tweaked the way modals are handled in the whole app, since we were hitting the render of all modals all the time. I've thoroughly tested it to the best of my ability but pay close atention to other modal flows now breaking now.


src/components/base/Modal/ModalFooter.js
```diff
@@ -1,18 +1,19 @@
...
-}
+  &:empty {
+    display: none;
+  }
+`
 
-const ModalFooter = ({ children }: { children: any }) => (
-  <div style={MODAL_FOOTER_STYLE}>{children}</div>
-)
+const ModalFooter = ({ children }: { children: any }) => <Wrapper>{children}</Wrapper>
```
meriadec: doesn't that mean that `ModalFooter` === `Wrapper`? :doge:

juan-cortes: indeed, i didn't even notice. I'll refactor with the (i'm sure) many things coming from @gre's review :trollface:

juan-cortes:
So I don't forget, things we talked about in slack @gre @aramab 

- [ ] Make the progress step dismissable too?
- [ ] Change the wording / Remove the "no device" button on the final step?

src/components/modals/MigrateAccounts/steps/01-step-overview.js
```diff
@@ -0,0 +1,209 @@
...
+              </Text>
+            </NextDeviceWarning>
+          ) : null}
+          {migratableCurrencyIds.map(currencyId => (
+            <Currency key={currencyId}>
+              <Text color="dark" ff="Open Sans|SemiBold" fontSize={4}>
+                <Trans
+                  i18nKey="migrateAccounts.overview.currency"
+                  values={{
+                    currency: migratableAccounts[currencyId][0].currency.name,
```
gre: in this map, could you make a local variable `migratableAccounts[currencyId]` to help visibility?
the `[0]` feels a bit type unsafe but i guess ok because we guarantee there is at least one item?

src/components/modals/MigrateAccounts/index.js
```diff
@@ -0,0 +1,201 @@
...
+      logger.critical(err)
+    }
+    this.setState({ scanStatus, err })
+  }
+
+  handleNextCurrency = () => {
+    const { migratableAccounts } = this.props
+    const { migratedAccounts } = this.state
+    const currencyId = Object.keys(migratableAccounts).find(
+      d => !Object.keys(migratedAccounts).includes(d),
```
gre: i'm a bit lost in how this select the "next currency". My current intuition is it will never reach a "third" currency if some of the account was not migrated? maybe i'm missing something.
at no point you have an iteration on the currency so i'm a bit 🤔  how it works

should it work by looking at the indexOf in the migratable per currency and just take the next index?

I think there is something simpler to use in term of codebase.
I wonder if this can already work:

```js
const ids = Object.keys(migrableAccounts)
const nextCurrencyId = ids[ ids.indexOf(currentCurrencyId)+1 ]
const currency = nextCurrencyId ? getCryptoCurrencyById(nextCurrencyId) : null
```

interesting fact: `currency===null` would mean you are done on the "currency iteration", then you can check if migratableAccounts is empty to know if others remain.

src/components/modals/MigrateAccounts/index.js
```diff
@@ -0,0 +1,201 @@
...
+        )}
+      />
+    )
+  }
+}
+
+const mapStateToProps = createStructuredSelector({
+  device: getCurrentDevice,
+  accounts: accountsSelector,
+  totalMigratableAccounts: state => accountsSelector(state).filter(canBeMigrated).length,
```
gre: I think this is not necessary because it's already part of `migratableAccounts` . i would prefer we don't do twice the same thing . either we could just inject `accountsSelector(state).filter(canBeMigrated)` as one prop, or we could just infer that total accounts by doing a sum of all groups. if it's just about checking if there are some remaining, would it be the same as checking `Object.keys(migratableAccounts)` length?

src/components/modals/MigrateAccounts/steps/03-step-currency.js
```diff
@@ -0,0 +1,193 @@
...
+    this.unsub()
+    const {
+      currency,
+      device,
+      setScanStatus,
+      addMigratedAccount,
+      replaceAccounts,
+      transitionTo,
+    } = this.props
+    try {
```
gre: I know this was taken from the scan accounts but it's getting a bit odd legacy stuff.
you can remove the try catch because unecessary and you can just `if(!currency||!device) return`

src/components/modals/MigrateAccounts/steps/03-step-currency.js
```diff
@@ -0,0 +1,193 @@
...
+      invariant(currency, 'No currency to scan')
+      invariant(device, 'No device')
+
+      const bridge = getCurrencyBridge(currency)
+
+      // TODO: use the real device
+      const devicePath = device.path
+      this.scanSubscription = bridge.scanAccountsOnDevice(currency, devicePath).subscribe({
```
gre: just pass in `device.path` because that's a weird // TODO comment above

src/components/modals/MigrateAccounts/steps/03-step-currency.js
```diff
@@ -0,0 +1,193 @@
...
+        next: account => {
+          const { migratableAccounts } = this.props
+          if (
+            migratableAccounts[currency.id] &&
+            migratableAccounts[currency.id].some(
+              a => a.freshAddressPath === account.freshAddressPath,
```
gre: mmh I would not trust that to be stable and source of trust for migration. Is there a way we can either use a method in live-common that delegate this work? I think it was findAccountMigration job somehow.

NB in future, we might have migration that doesn't have guarantee of fresh address equality, imagine on bitcoin as this is not immutable.

we need to me to see why we need `addMigratedAccount `

there is a dangerous assumption that this will cover the exact diff that the `addAccounts` logic is doing and I would suggest we do some kind of "diff" using it.

src/components/modals/MigrateAccounts/steps/03-step-currency.js
```diff
@@ -0,0 +1,193 @@
...
+  migratableAccounts,
+  migratedAccounts,
+  nextCurrency,
+  totalMigratableAccounts,
+  onCloseModal,
+}: StepProps) => {
+  if (scanStatus !== 'finished') return null
+  let next = 'overview'
+
+  if (Object.keys(migratableAccounts).length > Object.keys(migratedAccounts).length) {
```
gre: see my other comment on how we can solve this.

src/components/modals/MigrateAccounts/index.js
juan-cortes: According to design we need the "There are x accounts left to migrate" after a device loop. I'd need to iterate all keys in `migratableAccounts`, and sum them. I had it like that, considered this approach cleaner since we already have the `accountsSelector` there

src/components/modals/MigrateAccounts/steps/01-step-overview.js
juan-cortes: `migratableAccounts` is an object that is made by grouping the migratable accounts, so if there are no accounts there is no group at that point.

src/components/modals/MigrateAccounts/index.js
```diff
@@ -0,0 +1,201 @@
...
+  nextCurrency: () => void,
+  flushMigratedAccounts: () => void,
+  setAppOpened: (isAppOpened: boolean) => void,
+  scanStatus: ScanStatus | string,
+  isAppOpened: boolean,
+  onCloseModal: () => void,
+}
+
+type State = {
+  stepId: StepId | string,
```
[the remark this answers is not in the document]
juan-cortes: flow 🤷‍♂

src/components/modals/MigrateAccounts/index.js
```diff
@@ -0,0 +1,201 @@
...
+  isAppOpened: boolean,
+  currency: ?CryptoCurrency,
+  scanStatus: ScanStatus | string,
```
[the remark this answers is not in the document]
juan-cortes: flow again 🤷‍♂

src/components/modals/MigrateAccounts/index.js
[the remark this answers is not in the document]
gre: you defined `type StepId = 'overview'` so it's obviously not complete. that's why flow complains, if you add the 2 missing cases, you don't need `| string`

src/components/modals/MigrateAccounts/index.js
```diff
@@ -0,0 +1,188 @@
...
+  handleSetAppOpened = (isAppOpened: boolean) => this.setState({ isAppOpened })
+  handleSetScanStatus = (scanStatus: string, err: ?Error = null) => {
+    if (err) {
+      logger.critical(err)
+    }
+    this.setState({ scanStatus, err })
+  }
+  handleNextCurrency = (forceNull: boolean = false) => {
+    const { migratableAccounts } = this.props
+    this.hideLoopNotice = false
```
gre: if this is required, it should be a state.

gre:
after testing this feature, here are my (design/product) feedbacks:

- Some wording are to be checked. e.g. "Continue to next currency"
- The update header currently appear on Portfolio but NOT on Accounts. I expected to at least be in Accounts, Portfolio i'm not even sure!
- Will we have a Warning icon instead of the sync green check icon?

src/components/modals/MigrateAccounts/index.js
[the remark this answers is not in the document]
gre: @juan-cortes still can be fixed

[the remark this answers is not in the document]
juan-cortes: 🤦‍♂

gre:
@juan-cortes some conflicts with develop.

juan-cortes:
> @juan-cortes some conflicts with develop.

done , i think

package.json
```diff
@@ -47,7 +47,7 @@
     "@ledgerhq/hw-transport-http": "^4.66.0",
     "@ledgerhq/hw-transport-node-hid": "^4.66.0",
     "@ledgerhq/hw-transport-node-hid-singleton": "^4.66.0",
-    "@ledgerhq/ledger-core": "^3.0.0",
+    "@ledgerhq/ledger-core": "^3.0.0-beta.6",
```
gre: this should not happen. it should be 3.0.0

juan-cortes: 🤦‍♂

juan-cortes: you got it

gre:
@juan-cortes i would suggest you do the live-common bump as part of this PR as well as the lil' bridge fixes i discussed with you on slack. thx

juan-cortes:
I swear this pr will never be merged

gre:
@juan-cortes can you resolve the PR conflicts? We will be able to merge and later fix ETC metamask case. If it doesn't work we would disable ETC in the next release to migrate until we fix it.

juan-cortes:
ᕕ( ಠ‿ಠ)ᕗ

PR #2189, opened 2019-07-15
Fix not-centered boxes in account page

pbondoer:
The modules shown in the account pages were not centered if the percentage module was hidden.

Before | After
---------|------------
 | 

### Type

UI polish

Account page


gre:
5x developers

PR #2192, opened 2019-07-16
added eslint rules, removed warnings

IAmMorrow:
added react-hooks eslint rule. fixed newly generated warnings (the codebase was mostly compliant already)

### Type

Code Quality Improvement

### Context

LL-1639

eslint configuration, the whole app coding style

### DIFF
--- a/src/components/modals/AddAccounts/steps/03-step-import.js
+++ b/src/components/modals/AddAccounts/steps/03-step-import.js
@@ -89,7 +89,7 @@ const SectionAccounts = ({ defaultSelected, ...rest }: *) => {
     if (defaultSelected && rest.onSelectAll) {
       rest.onSelectAll(rest.accounts)
     }
-  }, [])
+  })
   return <AccountsList {...rest} />
 }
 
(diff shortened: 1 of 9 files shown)


src/components/modals/AddAccounts/steps/03-step-import.js
```diff
@@ -89,7 +89,7 @@ const SectionAccounts = ({ defaultSelected, ...rest }: *) => {
     if (defaultSelected && rest.onSelectAll) {
       rest.onSelectAll(rest.accounts)
     }
-  }, [])
+  }, [defaultSelected, rest])
```
gre: this one will be problematic I think. rest always changes, right? this is meant to be exec once on mount.

IAmMorrow: @gre hmmm but if we actually use `onSelectAll` and `accounts` in that component should we not declare them as proper props ? I think `rest` should be passed "as is" and not be interacted with

IAmMorrow: This is OK anyway, as non providing an input arg solve the problem. The eslint hook rule doesn't like empty arrays

gre:
@Arnaud97234 you can check all changed component still work fine. @IAmMorrow can help if need more context.

PR #2194, opened 2019-07-16
Bump live-common with fixes

gre:
- Fixes "NetworkDown" to properly uses the wording (and similarly, any error that throws inside sync was accidentally obfuscated)
- ZEC explorer op view to use https://chain.so
- BTC testnet explorer op view to use blockcypher
- failsafe to ensure there is never dup accounts after an add account (mainly for mobile)

### Type

Bug Fix

### Context

bug found in 1.12.0 testing

sync and add accounts


gre:
@Arnaud97234 should be good now

PR #2195, opened 2019-07-17
if no existing accounts are found, pre-select the last new account in…

IAmMorrow:
… the list

### Type

UI Polish

### Context

LL-1673

Add account flow, account import step


gre:
great! (@IAmMorrow can same be done for mobile?)

PR #2198, opened 2019-07-18
LL-1440 Hide empty token accounts

MortalKastor:
Add a toggle in the _Accounts_ settings to hide token accounts with 0 balance.

This also introduce a new mechanism for propagating environment variables to all the threads without the need to kill anyone 🕊

This has started with the idea of removing the relevant accounts at the _sync_ level so that hidden accounts would disappear _everywhere_, but that brought issues, as reported by @Arnaud97234 

It now selectively hides the accounts only in some part of the UI:

- Accounts screen
  - List mode
  - Card mode
  - Search results
- Token list in parent account screen
- Breadcrumb
- Send modal

Other parts shouldn't be affected by this toggle, like (non exhaustive list):
- Operations
- Operation details
- Receive flow
- …

### Type

Feature

### Context
LedgerHQ/ledger-live-common#294
https://ledgerhq.atlassian.net/browse/LL-1440

### Parts of app affected

- Everywhere token accounts are shown
- Environment variables


MortalKastor:
HODL for now: waiting for final UI design.

gre:
mmh that's surprising in the first place that the category is there without any item in it.

But Arnaud have a fair point. Bad guidance from my part on the implementation on this, it sounds like it might have to be solved in the UI side then 🤔 

we need to brainstorm @MortalKastor

MortalKastor:
Waiting for live-common release with LedgerHQ/ledger-live-common#311 and LedgerHQ/ledger-live-common#312

gre:
@MortalKastor 

- do you know if wording was validated?
- `develop` is now having the function you need from live-common

MortalKastor:
> * do you know if wording was validated?
Not that I know of
> * `develop` is now having the function you need from live-common
🤝
> * rebase & squash commits if possible
✅

gre:
tested and working. we'll have Arnaud testing on develop.

PR #2201, opened 2019-07-19
LL-1533: See related transactions in ETH fee operations

pbondoer:
This PR adds related `subOperations` to the operation detail view. It helps users understand why some transactions they may not have made are there (e.g. in the case of ETH, when sending tokens, an ETH fee will apply which shows up as a separate transaction).

### :ok_hand: Type

Feature / UX

### :mag: Context

LL-1533

Operation details

### DIFF
--- a/src/components/modals/OperationDetails.js
+++ b/src/components/modals/OperationDetails.js
@@ -24,15 +24,26 @@ import Bar from 'components/base/Bar'
 import FormattedVal from 'components/base/FormattedVal'
 import Modal, { ModalBody } from 'components/base/Modal'
 import Text from 'components/base/Text'
+import LabelInfoTooltip from 'components/base/LabelInfoTooltip'
+import OperationC from 'components/OperationsList/Operation'
 
 import CopyWithFeedback from 'components/base/CopyWithFeedback'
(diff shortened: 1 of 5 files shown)


gre:
nice

src/components/modals/OperationDetails.js
```diff
@@ -203,6 +232,51 @@ const OperationDetails = connect(mapStateToProps)((props: Props) => {
...
+                        onOperationClick={() =>
+                          openModal(MODAL_OPERATION_DETAILS, {
+                            operationId: op.id,
+                            accountId: op.accountId,
+                            parentId: account.id,
+                          })
+                        }
+                        t={t}
+                      />
+                      {i < subOperations.length - 1 && <B />}
```
gre: What is `<B>` :thinking:

pbondoer: A [styled `Bar` component](https://github.com/pbondoer/ledger-live-desktop/blob/223fc2823e8a646b985f79f9ffa0146f0e1a7b64/src/components/modals/OperationDetails.js#L102) apparently. Not sure why but since the rest of the file used it I rolled with it 🅱️

PR #2210, opened 2019-07-23
Update libcore to 3.1.0-beta.1

juan-cortes:
let's merge this soon so we can later have a live-common version that consume internal transactions.

NB: this PR should be considered to not change anything, there are new feature in libcore that are not used yet. Just test everything still works. especially with ETH / BTC


juan-cortes:
@gre this one is killing ci, but no reason that I can see

PR #2211, opened 2019-07-23
Support internal transactions + libcore/livecommon bump

gre:
see https://github.com/LedgerHQ/ledger-live-common/pull/308 for full description.

Capture d’écran 2019-07-23 à 22 17 08

### Type

Feature

- Ethereum with experimental nodes and experimental core
- **you must do one clear cache to have the internal operations appearing**


gre:
@Arnaud97234 I'm afraid WETH is a special case and we'll have to live with it. I don't think you see it on etherscan neither right? https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2?a=0x0e3f0bb9516f01f2c34c25e0957518b8ac9414c5 

I think it's the problem of WETH that don't always emit a transfer event. @KhalilBellakrid do you concur?

static/i18n/en/app.json
```diff
@@ -322,8 +322,10 @@
     "viewOperation": "View in explorer",
     "showMore": "Show {{recipients}} more",
     "showLess": "Show less",
-    "tokenOperations": "Token transactions",
-    "tokenTooltip": "This operation is related to the following token transactions",
+    "tokenOperations": "Token operations",
+    "tokenTooltip": "This operation is related to the following token operations",
+    "internalOperations": "Internal operations",
+    "internalOpTooltip": "This operation have internal operations",
```
pbondoer: `This operation has internal operations` would be better ;)

src/components/modals/OperationDetails.js
```diff
@@ -272,11 +271,48 @@ const OperationDetails = connect(
...
+                      key={`${account.id}_${operation.id}`}
+                      onOperationClick={() =>
+                        openModal(MODAL_OPERATION_DETAILS, {
+                          operationId: op.id,
+                          accountId: op.accountId,
+                        })
+                      }
+                      t={t}
+                    />
+                    {i < internalOperations.length - 1 && <B />}
```
pbondoer: :b:

gre:
we agreed this PR could be merged. This WETH is accepted as a limitation for now. WETH contract is having Withdraw and Deposit events concept that is beyond the ERC20 spec and that lib-ledger-core/backend does not handle yet as an "operation" and we'll leave it to them per this LLC-306.

I'm so glad we can finally ship this feature 🚀

PR #2227, opened 2019-07-30
Wording + ci

juan-cortes:


src/components/modals/OperationDetails.js
```diff
@@ -132,7 +133,7 @@ const mapStateToProps = (state, { operationId, accountId, parentId }) => {
     ? confirmationsNbForCurrencySelector(state, { currency: mainCurrency })
     : 0
   let operation = null
-  if (account) {
+  if (account && account.type === 'Account') {
```
gre: i'll fix `findOperationInAccount` to work for a token account but it should probably not filter ==='Account' here

src/components/modals/AddAccounts/steps/01-step-choose-currency.js
```diff
@@ -29,7 +29,15 @@ const TokenTips = React.memo(({ currency }: *) => (
...
     <InfoCircle size={16} color={colors.wallet} />
     <Text style={{ flex: 1, marginLeft: 20 }} ff="Open Sans|Regular" fontSize={4}>
-      <Trans i18nKey="addAccounts.tokensTip" values={{ currency: currency.parentCurrency.name }} />
+      <Trans
+        i18nKey="addAccounts.tokensTip"
+        values={{
+          token: currency.name,
+          ticker: currency.ticker,
+          tokenType: currency.tokenType.toUpperCase(),
+          currency: currency.parentCurrency.name,
```
gre: 👌

PR #2233, opened 2019-07-31
Fixes optimistic erc20

gre:
see https://github.com/LedgerHQ/ledger-live-common/pull/313


gre:
Arnaud & I spotted another case were some op might disappear but it's more edge case -ish. we'll dig.

PR #2234, opened 2019-07-31
LL-1708 Allow SubOperations/InternalOperations to navigate back

juan-cortes:
Allows users viewing the operation details modal for sub operation and internal operations to navigate back to the parent operation.

Also added a little cleanup using the helper functions from common, we should try to do this when we touch components that are still doing the conditionals instead of it (`getAccountCurrency`, `getAccountUnit`, `getMainAccount` ...)

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1708

Operation details modal for operations with suboperations or internal transactions

### DIFF
--- a/src/components/modals/OperationDetails.js
+++ b/src/components/modals/OperationDetails.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { Fragment, Component } from 'react'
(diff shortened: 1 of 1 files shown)


juan-cortes:
@Arnaud97234 try latest version, it should be fixed. Nuances of the different operation types 😅

PR #2238, opened 2019-08-01
Collapse on token sidebar list

juan-cortes:
collapse

Hacky way (had to position the bad absolutely since otherways it wouldn't align with the parent one) to provide the user a subtle way to collapse the token list without scrolling to the bottom of the screen. While this is not an issue for users with a low number of tokens, it is an issue for us devs, and I'm sure for some power-users too. 🚗 

### Type

UX improvement

### Context

🤷‍♂ 

Account list, account with token accounts, this enables collapsing the list by clicking on the side bar.


gre:
nice

PR #2244, opened 2019-08-02
Bump live-common 7.10

gre:
add failsafe for parsing error happening in LNS 1.2.0.

### Type

bugfix

- it is not guarantee for a LNS 1.2 to be fully functional but at least it fixes one problem.
- everything else should still work like before around add account/receive/send flows.


gre:
@Arnaud97234 i'm not sure why this is blocking this PR?

gre:
replaced by #2269

PR #2246, opened 2019-08-02
LL-1725 Fix ellipsis on address in operations list

MortalKastor:
Embedding `OperationsList` in modals brought out that addresses where not correctly truncated, and are overflowing over their siblings.

This should fix that.

Before:
Screen Shot 2019-08-02 at 18 49 14

After:
Screen Shot 2019-08-02 at 18 50 13

### Type

Bugfix/UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1725

### Parts of the app affected

Operation lists, especially in modals, like for ERC-20 related operations from parent ETH account.

### DIFF
--- a/src/components/OperationsList/AddressCell.js
+++ b/src/components/OperationsList/AddressCell.js
@@ -10,29 +10,41 @@ const Address = ({ value }: { value: string }) => {
     return <Box />
   }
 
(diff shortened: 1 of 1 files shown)


gre:
great

PR #2250, opened 2019-08-07
LL-1713: Fix amount not updated when user enters recipient address after checking send max

pbondoer:
This PR fixes a bug where checking send max before entering the address would result in transaction amount 0. This was due to a race condition: the transaction gets updated twice: the first time (without a `gasLimit`) would throw on `checkValidTransaction` but it would get caught *after* the second (with `gasLimit`) one was validated :thinking: 

The fix was simply to check the sync ID before updating state on error :tada: 

### :ok_hand: Type

Bug fix

### :mag: Context

LL-1713

Send max ERC20

### DIFF
--- a/src/components/modals/Send/fields/AmountField.js
+++ b/src/components/modals/Send/fields/AmountField.js
@@ -39,10 +39,10 @@ class AmountField extends Component<
   componentDidMount() {
     this.resync()
   }
-  componentDidUpdate(nextProps: *) {
+  componentDidUpdate(prevProps: *) {
     if (
-      nextProps.account !== this.props.account ||
-      nextProps.transaction !== this.props.transaction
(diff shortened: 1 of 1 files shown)


src/components/modals/Send/fields/AmountField.js
```diff
@@ -39,10 +39,10 @@ class AmountField extends Component<
   componentDidMount() {
     this.resync()
   }
-  componentDidUpdate(nextProps: *) {
+  componentDidUpdate(prevProps: *) {
```
juan-cortes: indeed, good catch

pbondoer: It actually messed with me during debugging, I was seeing the previous props and thinking "wait, that's not right..." :thinking: :smile:

PR #2252, opened 2019-08-07
LL-1775 Remove Edit from context menu on tokens

MortalKastor:
On _Accounts_ page in grid mode, context (right click) menu was showing an _Edit Account_ entry, resulting in a crash when selected.

This PR remove the entry for token accounts.

<details>
<summary>Here lies an out of date screenshot ⚰️</summary>

Screen Shot 2019-08-07 at 17 48 55

</details>

EDIT:
- Redundant star removed
- Pixel polish

Screen Shot 2019-08-08 at 14 57 46

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-1775

### Parts of the app affected 

Context menu on accounts cards

### Test plan

- Go to _Accounts_ page
- Set to grid mode
- Right click on a token account: _Edit account_ should not be present
- Right click on a regular account: _Edit account_ should be there

### DIFF
--- a/src/components/ContextMenu/AccountContextMenu.js
+++ b/src/components/ContextMenu/AccountContextMenu.js
@@ -27,7 +27,7 @@ const mapDispatchToProps = {
 class AccountContextMenu extends PureComponent<Props> {
   getContextMenuItems = () => {
     const { openModal, toggleStarAction, account } = this.props
-    return [
+    const items = [
       {
         label: 'accounts.contextMenu.send',
         Icon: IconSend,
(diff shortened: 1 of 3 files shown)


src/components/ContextMenu/AccountContextMenu.js
```diff
@@ -2,32 +2,28 @@
...
-  toggleStarAction: Function,
   openModal: Function,
 }
 
 const mapDispatchToProps = {
   openModal,
-  toggleStarAction,
 }
 
 class AccountContextMenu extends PureComponent<Props> {
```
valpinkman: this component could be refactored as a functional component

juan-cortes:
@MortalKastor this is what I mentioned in the daily (I mumbled I should say), for token accounts we need to pass the maybe parentAccount to `AccountContextMenu` so that the `openModal` calls work. When I refactored this I completely forgot about this fact.

PR #2258, opened 2019-08-09
Adds double icon to stars, reworked parentcryptocurrencyicon

juan-cortes:
Reworked the way `ParentCryptoCurrencyIcon` works to no longer maintain an arbitrary size but rather a smaller and a bigger one. There used to be an erroneus use of this with a size of 40 in the add account flow but that made no sense. 

In order to fix the white border glitch on the double icon component when the background is not white we are now clipping the svg icon of the parent account. This could potentially be done better but for now it works.

Basically passing a clip path polygon to the svg element. 

image

image

Allows us to have hover/active states changing the background of the parent element without showing off a non-matching border color.

image

### Type

UI Polish


gre:
great

PR #2260, opened 2019-08-09
LL-1759: Pixel push - margins edition™

pbondoer:
Fix some margins, with @dekkaki :tada: 

Some changes were a bit weird to make (everything is so interlocked it was hard to change one thing without everything else changing with it) so we made some quick hacks, but maybe we should rethink this?

### :ok_hand: Type
UI polish

### :mag: Context
LL-1759

* Portfolio
  * Check banner case (i.e. migration)
* Accounts
  * Check with banner as well
* Buy crypto

### DIFF
--- a/src/components/AccountsPage/AccountsHeader.js
+++ b/src/components/AccountsPage/AccountsHeader.js
@@ -27,7 +27,7 @@ class AccountsHeader extends PureComponent<Props> {
   render() {
     const { t } = this.props
     return (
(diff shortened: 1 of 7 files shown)


gre:
status of this PR?

pbondoer:
@gre Waiting for validation from @Arnaud97234 for the banner but should be good now, we decided to keep the changes, just add back the little margin on the banner which we removed by mistake :ok_hand:

PR #2262, opened 2019-08-12
LL-1747 Persist starred accounts when migrating them

juan-cortes:
This should allow to persist the starred account when we do the migration. Not the prettiest code but until we have a nicer solution this should be enough.

### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-1747

Make sure that migration with.without starred accounts is not broken now. According to my testing it's all good :trollface: 

---------

**Note** the extra non related changes are triggered by prettier. These have passed out ci so something is wrong on that end. Until we get that fixed.... things like this will happen.

### DIFF
--- a/src/actions/settings.js
+++ b/src/actions/settings.js
@@ -78,3 +78,8 @@ export const dragDropStarAction = (payload: {
   type: 'SETTINGS_DRAG_DROP_STAR',
   payload,
 })
+
+export const replaceStarAccountId = (payload: { oldId: string, newId: string }) => ({
+  type: 'SETTINGS_REPLACE_STAR_ID',
+  payload,
+})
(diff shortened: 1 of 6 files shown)


gre:
BTW i think we are still prone to the problem if you "add accounts" which can also happen to do a migration. there are and there will be in future a few other cases where ids of account can happen to change and i'm not sure how we aim to address these. like a sync() is itself allowed to change ids, we've used that in the past in order to "migrate" logic like we have a libcore version in the id `libcore:1:...` when things fundamental change (example in the past: we've fixed a bug in missing info in operations/wrong amounts so we changed id to force a full clear)

**the question that is raised under the hood is shouldn't the star info be part of Account and TokenAccount after all..**

juan-cortes:
I would like to have an "extras" object to allow flags in the account model but I fear that can get dirty pretty quickly.

gre:
it doesn't need to be an extra blackbox in that it's not specific to an account/currency. can be explicit like account.name is

PR #2264, opened 2019-08-14
LL-1781 Keep keyboard focus visible on lists

juan-cortes:
This should allow moving in a dropdown selector using the arrows from the keyboard.

### Type

UX Polish, feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1781

Any dropdown basically, add account, select account etc.

### DIFF
--- a/src/components/base/Select/index.js
+++ b/src/components/base/Select/index.js
@@ -49,14 +49,47 @@ const Row = styled.div`
 `
 const rowHeight = 40 // Fixme We should pass this as a prop for dynamic rows?
 class MenuList extends PureComponent<*> {
+  state = {
+    children: null,
+  }
+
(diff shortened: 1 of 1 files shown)


src/components/base/Select/index.js
```diff
@@ -49,14 +49,43 @@ const Row = styled.div`
...
+      return {
+        children,
+        currentIndex,
+      }
+    }
+    return null
+  }
+
+  componentDidMount() {
+    this.componentDidUpdate()
```
valpinkman: hmm, not a big fan here.

Can you create a function that does this, and use it in `componentDidMount` and `componentDidUpdate` instead ?

juan-cortes: Done, I copies that and didn't even check it 😅

PR #2265, opened 2019-08-14
jump to account from operation details

valpinkman:
Possibility to jump to account page directly from operation details

### Type
feature

### Context

LL-1721

Operation details
image


valpinkman:
@Arnaud97234 fixed

gre:
@valpinkman 
- some conflicts
- for ERC20 case, let's quickly discuss together what we do. We could do something like `Ethereum 1 / MANA` where each of `Ethereum 1` and `MANA` are individually touchable? 🤔  cc @aramab thought?

PR #2271, opened 2019-08-19
Op details

pbondoer:
Small fixes for operation details

### :ok_hand: Type
Code quality / UI polish

### :mag: Context

LL-1749
LL-1770

Operation details
Operation details when having internal/sub operations

### DIFF
--- a/src/components/modals/OperationDetails.js
+++ b/src/components/modals/OperationDetails.js
@@ -276,7 +276,7 @@ const OperationDetails = connect(
                   if (!opAccount) return null
 
                   return (
-                    <NoMarginWrapper key={`${account.id}_${operation.id}`}>
+                    <NoMarginWrapper key={`${op.id}`}>
                       <OperationComponent
                         compact
                         text={opAccount.token.name}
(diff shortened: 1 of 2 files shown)


src/components/modals/OperationDetails.js
```diff
@@ -357,6 +355,11 @@ const OperationDetails = connect(
                           value={fee}
                           alwaysShowSign={false}
                           subMagnitude={1}
+                          prefix={
+                            <Box mr={1} color="grey" style={{ lineHeight: 1.2 }}>
+                              {'≈'}
+                            </Box>
+                          }
```
pbondoer: Couldn't find a better way to do this without rewriting a lot of code. If there's a better way please let me know :eyes:

gre: lgtm. i think we have the same concept in mobile and i was suprised it was not yet the case on desktop! https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/components/CurrencyUnitValue.js#L20-L21

PR #2273, opened 2019-08-20
LL-1600: New periodicity tabs display

pbondoer:
Some changes to the desktop app:

* Moving the periodicity tabs to allow seeing full values when they would previously overflow
* Compact display of percentage / changes
* Removal of the greeting
* Some minor adjustments

### :ok_hand: Type
UI polish

### :mag: Context
LL-1600

Portfolio
Account page

### :bug: Remaining issues

* Issue with percentage (to investigate)

* Margin on Portfolio page for the card might not be enough (cc @dekkaki)


src/components/base/FormattedVal/index.js
```diff
@@ -56,6 +56,8 @@ type OwnProps = {
   disableRounding?: boolean,
   isPercent?: boolean,
   subMagnitude?: number,
+  prefix?: string,
+  suffix?: string,
```
gre: does that mean we can drop the one that was on Countervalue component and delegate to FormattedVal?

pbondoer: No as this can only accept a string due to the fact it gets appended. The other one accepts React nodes :)

gre:
@pbondoer oops, some conflict. could you also refresh the screenshot if needed? thx

juan-cortes:
This will be merged into LL-1825

PR #2278, opened 2019-08-21
LL-1819 Sort accounts by currency for migrations

juan-cortes:
Ideally this should show no different behaviour from the current migration flow, but it should prevent the edgecase where we would be eternally looping a currency instead of iterating through them. For now, this is not an issue since we remove ETC but we should revisit it if we ever allow ETC migrations.

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-1819

Full migration flow

### DIFF
--- a/src/components/modals/MigrateAccounts/index.js
+++ b/src/components/modals/MigrateAccounts/index.js
@@ -8,11 +8,9 @@ import { connect } from 'react-redux'
 import { translate } from 'react-i18next'
 import { compose } from 'redux'
 import { createStructuredSelector } from 'reselect'
-import { accountsSelector } from 'reducers/accounts'
-import groupBy from 'lodash/groupBy'
+import { migratableAccountsSelector, accountsSelector } from 'reducers/accounts'
 import logger from 'logger'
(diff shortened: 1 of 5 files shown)


Arnaud97234:
In order to be able to validate that, we need at least 2 currencies to migrate. So far only Ethereum is available for migration so we'll wait for XRP or ETH integration in libcore to test. @gre @juan-cortes what do you think ?

gre:
- first we should rebase this because code diverged. @juan-cortes 
- secondly, I would say we could just merge if this code works fine today and check again later when we enable ETC or XRP & fix at the time we do that if any glitches. @Arnaud97234

gre:
@juan-cortes can you fix the conflicts?

juan-cortes:
> @juan-cortes can you fix the conflicts?

Done, rebased with develop, the only conflict was about the error feedback for no accounts migrated done on LL-1809 that took longer to merge I guess

gre:
@juan-cortes check CI.

juan-cortes:
> @juan-cortes check CI.

Current version should be ok, can't test the next currency thing either so we'll have to put a pin on that when we have another migration happening at the same time.

PR #2279, opened 2019-08-21
LL-1564: Fix graph date display

### LINKED ISSUE #2266: Date axis is not synced with graph tooltip

- Ledger Live 1.13.2
- Platform and version: Mac 10.14.6

#### Expected behavior

As the image shows, the chart axis shows one date, but the tooltip shows one day before.

Screen Shot 2019-08-17 at 03 06 08

pbondoer:
This PR is an attempt to fix the display of graph dates. The issue is that most dates we have end on 23:59:59 and for some inexplicable reason d3 seems to round up when sending data to `tickFormat`...

### :ok_hand: Type
Bug fix

### :mag: Context
Fixes #2266
LL-1564

All graphs with a visible tick scale:
* Portfolio
* Account page

### DIFF
--- a/src/components/base/Chart/helpers.js
+++ b/src/components/base/Chart/helpers.js
@@ -5,7 +5,10 @@ export function enrichData(data) {
     ...d,
     ref: d,
     index: i,
-    parsedDate: d.date,
+    // normalize all dates to start of day for graph display to prevent d3 from
+    // rounding up
+    // see https://github.com/LedgerHQ/ledger-live-desktop/issues/2266
+    parsedDate: new Date(d.date).setHours(0, 0, 0, 0),
   }))
 }
 


gre:
the problem is we're back at the "hey but why does the two last points have same date" which was reported as another bug. anyway, i guess we can trade that bug from the other^^ maybe last point should says "now" cc @aramab

pbondoer:
@gre In my testing, the last two points always had different date, perhaps this is an edge case?

gre:
Oh ok I see. Because it was fixed on the graph. Yeah so it works but as you said on slack, definitely not going to when we implement hourly rate.
To me the graph axis is not fundamentally wrong but it's basically putting a point at the beginning of the day and not in the middle of the day. Not sure how we could fix it, maybe the position should be set at the middle of the day/hour 🤷‍♂️ anyway something is still possible to generalize when we impl hourly

pbondoer:
@gre I think the fix is reasonable here, and if the granularity is anything less than `daily` we can just return the plain date instead of flooring it.

But you are definitely right, d3's scale is "technically correct" as the date is almost on the start of the next day, but for users it doesn't make much sense. You can see this with `.nice()` which d3 uses to produce "round" scale values and leaves a big blob of empty space at the start 😄

PR #2282, opened 2019-08-21
Fix the repair process for FW 1.3.1

valpinkman:
When user's device is in FW 1.3.1 and sees `MCU Not Genuine`, the repair process should not continue and we should explain to the user that he needs to return to dashboard to complete the firmware update process.

Also fix the font size for errors in repair process

### Type

Fix/Improvement

### Context

LL-1613 (but not exclusively)

Repair Modal

image

// How To Test

@Arnaud97234 you will need a Nano S in 1.3.1 that has a `MCU Not Genuine` state. I have one lying around if you need to.

PS: The *issue*  with this fix is that if some user start their 1.4.1 in bootloader mode by themselves, (but don't have the MCU not genuine message), from what I understand, they will trigger the same error as displayed above (although they SHOULD NOT enter bootloader mode to try and repair their device if everything is fine with it)

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -1184,6 +1184,10 @@
     "UnknownMCU": {
       "title": "Unknown MCU version",
       "description": "Please contact Ledger Support"
+    },
+    "MCUNotGenuineToDashboard": {
+      "title": "Access dashboard to update",
+      "description": "Please restart the device without pressing any button. Then press both buttons on the device three times to access the dashboard. Finally open the Manager to update the firmware."
     }
   }
 }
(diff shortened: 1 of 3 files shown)


static/i18n/en/app.json
```diff
@@ -1163,6 +1163,10 @@
     "UnknownMCU": {
       "title": "Unknown MCU version",
       "description": "Please contact Ledger Support"
+    },
+    "MCUNotGenuineToDashboard": {
+      "title": "No MCU update required",
```
dasilvarosa: Access dashboard to update

valpinkman:
> e clarify the title

Yeah... what should I use @dasilvarosa ?

static/i18n/en/app.json
valpinkman: thanks

dasilvarosa:
Please restart the device without pressing any button. Then press both buttons three times ...

valpinkman:
@dasilvarosa thanks

PR #2289, opened 2019-08-23
Bridge Refactoring

gre:
Big refactoring on the bridge that allow more code sharing and a better bridge interface.

### Type

Refactoring

### Context

LL-1576

Everything.
Only test the existing coins to not create any regression. especially on the Send flow.


static/i18n/en/app.json
```diff
@@ -1220,6 +1220,9 @@
     "FeeRequired": {
       "title": "Fees are required"
     },
+    "GasLessThanEstimate": {
+      "title": "Provided gas limit is less than the estimated"
```
gre: @dasilvarosa can you think of a better wording?
it's a warning that displays under the gas limit field when user lower the gas limit under the safe estimated one. (setting that low value is likely to fail the transaction)

if we can not say "gas limit" in the wording it's better so we can later reuwe this concept. if you think we need to mention it, still fine.

dasilvarosa: This may be too low, please increase

gre: thanks

src/components/AccountPage/TokensList.js
```diff
@@ -2,7 +2,7 @@
```
juan-cortes: Maybe rename this kind of files to `SubAccountList` then

gre: true. we'll do that when we rework the UI/UX with design team. to also generalize that UI on their side.

PR #2290, opened 2019-08-23
LL-1820: Show token account count in Add accounts flow

pbondoer:
This PR adds account count in Add Accouts flow. It also does some minor UI polish to align stuff.

### :ok_hand: Type
UX

### :mag: Context
LL-1820

### :clipboard:  Parts of the app affected / Test plan
Add accounts with ERC20

### DIFF
--- a/src/components/CryptoCurrencyIcon.js
+++ b/src/components/CryptoCurrencyIcon.js
@@ -14,11 +14,11 @@ type Props = {
 }
 
 // NB this is to avoid seeing the parent icon through
-const TokenIconWrapper = styled.div`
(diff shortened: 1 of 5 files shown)


meriadec:
Noice! does it expands with value? like, if I have 1000 tokens? :doge:
also, there is probably a way to get rid of this blank outline, either using mask or at least filling it with bg color

pbondoer:
@meriadec Blank outline is intentional (doesnt look centered without it), and no it doesnt expand but it does shrink the font size. After +999 no guarantees tho :D

PR #2302, opened 2019-08-30
LL-1825 See a portfolio of a specific asset

juan-cortes:
### Screenshots
image
image
image

### Type

Feature.

### Context

https://ledgerhq.atlassian.net/browse/LL-1825

https://ledgerhq.atlassian.net/browse/LL-1600

The whole new page and affected blocks from the portfolio. Special atention to the graph.
There's a known issue that would be addressed on a separate pr for the green and red arrows for delta.

---

#### NOTE THAT THIS INCLUDES THE WORK FROM @pbondoer ON LL-1600
This means that on this pr we no longer have the greeting on the portfolio, and we have the new periodicity tabs and design integrated too.


gre:
definitely feels like it need a wording review. the "you can now click on.." text feels weird.

juan-cortes:
Last commit addresses LL-1858 and LL-1859 by filtering the tokens if the flag is set and by adding the context menu to the token lists everywhere. This last one has been present ever since we added the initial context menu so good catch @Arnaud97234 let me know if you find anything else.

Pending @dasilvarosa check

juan-cortes:
I think the two bugs are more related to the balance summary block and I would rather address them on a separate pr, but I'll take a look in the morning anyway 👍

gre:
if possible please fix in this PR because we merged both work into one and i would prefer we don't merge inconsistency work in develop. 🙏

juan-cortes:
The mismatch between percentage and value is fixed on the live-common pr, the other one about the missing percentage is not related to this pr at all.

gre:
`develop` can be rebased as it contains latest live-common

juan-cortes:
@gre hover states added, ready for this "testing build"

juan-cortes:
>     * account name to not uppercase on erc20 account row
> 
> 
> Capture d’écran 2019-09-10 à 11 19 30
> 
>     * bug when you try to do "Send" with the "..." action on a erc20 account row

Both addressed on latest commit

PR #2308, opened 2019-09-03
adds fallback screen when looking for tokens in manager

valpinkman:
When searching for a Token app in manager, fallback to a screen that explains how to use ERC20 Tokens with Ledger Devices

### Type

Feat

### Context

LL-1834

Manager, looking for the name of an ERC20 token

image


dasilvarosa:
Inappropriate wording, let's discuss internally.

PR #2312, opened 2019-09-04
export to mobile after account migration

IAmMorrow:
image

### Type

Feature

### Context

LL-1829

Migrate -> Completed step -> Export to mobile


src/components/modals/MigrateAccounts/steps/03-step-currency.js
```diff
@@ -121,7 +122,11 @@ class StepCurrency extends PureComponent<Props> {
             }
           })
 
-          replaceAccounts(migrateAccounts({ scannedAccounts, existingAccounts: accounts }))
+          const migratedAccounts = migrateAccounts({ scannedAccounts, existingAccounts: accounts })
+          replaceAccounts(migratedAccounts)
+          migratedAccounts.forEach((account: Account) => {
+            addMigratedAccount(currency, account)
+          })
```
gre: I was thinking we would just expose **ALL** accounts to export, not necessarily just the one that was upgraded. Imagine you stopped the migration and did half of accounts and came back to it later.

ultimately I think it's better UX if we just provide all desktop accounts as potentially user would want to also import totally unrelated accounts he earlier added.
That also simplify the logic here and would not require the code changes.

cc @aramab thoughts on this?

IAmMorrow: It is clearly mentioned *Only export the migrated accounts* in the jira ticket.

Exporting only the migrated accounts was a pain in the ass to make, would have been way easier to simply export all accounts.

gre: ok so be it i guess. we'll chat again internally.

gre:
@IAmMorrow can you fix conflict and we should be good

PR #2313, opened 2019-09-04
LL-1740 Hide starred empty token accounts

MortalKastor:
When the user enables _Hide empty token accounts_, also hide them from the _Starred Accounts_ section in the sidebar.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1740

### Parts of the app affected

Starred accounts

### Test plan

- Toggling _Hide empty account_ should immediately reflect on the Starred accounts list
- Re-ordering accounts (via drag and drop) while empty accounts are hidden shouldn't mess up the list, even after toggling the hiding back off
- Special case brought to my attention by @juan-cortes: verify that nothing breaks when the list is shuffled around after deleting (not just unstarring) one of the starred accounts

### DIFF
--- a/src/components/Stars/index.js
+++ b/src/components/Stars/index.js
@@ -10,7 +10,7 @@ import { connect } from 'react-redux'
 import { createStructuredSelector } from 'reselect'
 import { push } from 'react-router-redux'
 import { dragDropStarAction } from '../../actions/settings'
-import { starredAccountsSelector } from '../../reducers/accounts'
+import { starredAccountsEnforceHideEmptyTokenSelector } from '../../reducers/accounts'
 import { i } from '../../helpers/staticPath'
 import Item from './Item'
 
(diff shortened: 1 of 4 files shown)


src/components/Stars/index.js
```diff
@@ -47,7 +47,10 @@ class Stars extends PureComponent<{
   onDragEnd = ({ source, destination }) => {
     const { dragDropStarAction, starredAccounts } = this.props
     if (destination) {
-      dragDropStarAction({ from: source.index, to: destination.index, starredAccounts })
```
juan-cortes: I was providing the list of accounts here because in the chance that we deleted an account, then the indexes no longer match, I'm not sure how this would behave

MortalKastor: As an attempt to prevent that issue, and to reliably move the correct account in the complete list when the drag and drop is done while empty accounts are hidden, I'm no longer sending indexes but account ids for `from` and `to`. I should probably have changed the object keys to something like `fromId` and `toId` to make it obvious though...

juan-cortes: 👏

Issue #2145, opened 2019-06-27
A theory regarding the missing accounts bug


juan-cortes:
Keep in mind this is an incomplete theory but... I got curious and started looking into the "missing accounts" issue that some of our users have been experimenting. We know that it seems to mostly happen to users who have set up a password and during an update they lose their accounts.

This points me to potentially involving the migration system where we initially had a separate file for `accounts.json` and later it was merged into a single `app.json`. 

## The migration system vs the testing system
The migration system will check a nonce variable and run all the migrations that are greater than that nonce. If you get the mock data from `userdata_v1.0.5_mock-02-encrypted-accounts`, wipe your own user data, replace it with this, and run the app, it will trigger the migration mechanism and merge all the data into `app.json`. It will all work fantastically, until it doesn't.  I was able to lose my data twice by repeating this manually for around 50 tries but we are not savages so in order to test this programmatically I ran `jest src/migrations` and it complies, it works..

## But wait!
We are debouncing the saves to the file system!, I understand this is because we have a copy in memory of the settings of the app, and we don't want a race condition or access lock affecting the `app.json` file when two or more simultaneous saves happen but when a migration is happening **we can't afford to have an active debounce in place**, or we will lose data.

<img width="822" alt="image" src="https://user-images.githubusercontent.com/4631227/60286941-f4607780-9910-11e9-9cfc-79e9cbc04fec.png">

The tests are set to only have `1ms` of debounce but the real life use to `500ms`, this means that a migration would debounce calls, when executed in the real world, if the calls are placed within half a second of eachother. If we change the `DEBOUNCE_MS` for tests to 500ms then **nothing but the window params is migrated.**

``` 
Array [
    +   "accounts.json",
        "app.json",
    -   "windowParams.json",
    +   "countervalues.json",
    +   "migrations.json",
    +   "settings.json",
    +   "user.json",
]
```

## Why does it only happen to some people
My theory is that you need a particular set of conditions for this to apply. You need enough account data for the save action to take **longer** than 500ms, so that the next call would cancel that save, so users with small accounts, small amounts of data, or super fast computers wouldn't be affected. It also means that it happens for users updating from a very old version for this to be true, so if the user has lost accounts and it's relatively new, migrations wouldn't necessarily be the cause but it could potentially be related if we are perhaps clearing the memory and saving.

[...]

PR #2329, opened 2019-09-10
LL-1844 Remove the data migration logic

juan-cortes:
Following the theory https://github.com/LedgerHQ/ledger-live-desktop/issues/2145 that a rogue migration can be causing the user data wipe, removing the data migration process will test the theory. 

### Type

Possible Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-1844

Migration is transparent for the user, make sure the app still launches.


juan-cortes:
@gre

PR #2336, opened 2019-09-11
LL-1862 Add a separator in the disclaimer modal

juan-cortes:
modal

This simply adds a border and a little bit of padding to the header so that it doesn't cut the content when scrolling. I did not do anything to the percentage loader referred to in the issue because I believe it's better to show 78% and then finish or whatever than get stuck in 100%. The screenshot is not really representative since it wont look exactly like that but the black border shown is the new addition (gray in the real world)

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-1862

Firmware update modal, disclaimer step.

### DIFF
--- a/src/components/base/Modal/ModalBody.js
+++ b/src/components/base/Modal/ModalBody.js
@@ -77,10 +77,10 @@ class ModalBody extends PureComponent<Props, State> {
           noScroll={noScroll}
         >
           {render && render(renderProps)}
+          <div style={GRADIENT_WRAPPER_STYLE}>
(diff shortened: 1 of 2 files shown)


juan-cortes:
I have no < 1.5.5 device to test can someone else take a look @valpinkman @MortalKastor @IAmMorrow please

valpinkman:
@juan-cortes will do right now

valpinkman:
@Arnaud97234 fixed

PR #2340, opened 2019-09-13
fix device not recognized after repair on manager genuine check

valpinkman:
Fixing a strange case where the manager would not find the device after a repair on the Manager Genuine Check. Found kind of a solution but not completely sure this is  over yet. More investigation / discussion is needed is think to resolve this issue

### Type

Bug Fix

### Context

LL-1856

Manager Genuine Check

### Examples

Video examples are available on the Jira issues @Arnaud97234 @gre

### DIFF
--- a/src/components/SettingsPage/RepairDeviceButton.js
+++ b/src/components/SettingsPage/RepairDeviceButton.js
@@ -49,8 +49,12 @@ class RepairDeviceButton extends PureComponent<Props, State> {
   timeout: *
 
   close = () => {
+    const { onRepair } = this.props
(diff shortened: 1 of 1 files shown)


Arnaud97234:
> Screenshot 2019-09-18 at 15 14 13
> 
> Could we change the style of the displayed text in order to make it easier readable for the user.
> For instance:
> 
> * Please restart the device without pressing any button.
> * Then press both buttons on the device three times to access the dashboard.
> * Finally open the Manager to update the firmware.

We'll deal solving that in a future PR

PR #2358, opened 2019-10-02
feat: enable deep link to ledger live

dviramontes:
### Type

feature ?

### Context

Add the ability to open ledger live from browser
ref: https://stackoverflow.com/questions/53530726/how-to-launch-my-electron-app-from-a-website

1. yarn 
2. yarn dist
3. [Use this pen to test opening ledger live from the web](https://codepen.io/dviramontes/pen/NWKQVEM)


gre:
do you have a usecase? I think it would be great if we could extend it a bit so it allows to prefill a "Send" flow so this could literally be used to put a "send donation" link on websites.

imagine stuff like

```
ledgerlive://send?currency=bitcoin&amount=0.01
```


food for thoughts :)

dviramontes:
Although i like the idea, my use case is much more simple. Just want to be able to open the app if it's installed.

gre:
we'll revive this idea soon. closing in context of a big rework introduced in https://github.com/LedgerHQ/ledger-live-desktop/pull/2596

PR #2361, opened 2019-10-07
More UI Polish

IAmMorrow:
Some more various UI pixel polish.

### Type

UI Polish

### Context

N/A

there and there ... :)


gre:
we'll take @Arnaud97234 feedback into account and we should create a LL task in our backlog to work on it.

PR #2362, opened 2019-10-07
disabling tippy when drag and drop

IAmMorrow:
### Type

Ui polish

### Context

LL-1903

Main side bar => Star accounts


gre:
@IAmMorrow conflicts (+ comments)

PR #2372, opened 2019-10-09
added Image base component

IAmMorrow:
Added a base Image component for dealing with images

### Type

Code Quality Improvement

### Context

N/A

N/A


src/components/base/Image.js
```diff
@@ -0,0 +1,44 @@
...
+
+export default ({ resource, alt, themeTyped = false, className, ...rest }: Props) => {
+  const type = useTheme('colors.palette.type')
+
+  return (
+    <Img
+      {...rest}
+      alt={alt}
+      className={className}
+      src={i(themeTyped ? `${type}-${resource}` : resource)}
```
MortalKastor: This `${type}-${resource}` won't work for images located in subfolders (which is the case for some of our images).

Also what do you think about just getting rid of the `themeTyped` prop and let this component check if `type-resource` file exists, and if not use the plain `resource`, so that themed assets get implicitly loaded if present?

PR #2395, opened 2019-10-18
partner logo now dynamic contrast enabled; bumped lib-common

IAmMorrow:
partner logo are now dynamically loaded according to the actual contrast

### Type

Ui Polish

### Context

N / A

Buy crypto page

### DIFF
--- a/src/components/PartnersPage/index.js
+++ b/src/components/PartnersPage/index.js
@@ -1,37 +1,37 @@
 // @flow
 
-import React, { PureComponent } from 'react'
+import React from 'react'
 import { T, translate } from 'react-i18next'
 
-import partners from '@ledgerhq/live-common/lib/partners/react'
+import getPartners from '@ledgerhq/live-common/lib/partners/react'
 import TrackPage from 'analytics/TrackPage'
 import Box from 'components/base/Box'
+import useTheme from 'hooks/useTheme'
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -1906,6 +1944,15 @@
     "@ledgerhq/errors" "^4.72.2"
     events "^3.0.0"
 
+"@ledgerhq/hw-transport@^4.72.1":
```
gre: this is unclear why this happens. can you ensure there is no dup of that lib?

src/components/PartnersPage/index.js
```diff
@@ -1,37 +1,37 @@
...
-          {t('partners.desc')}
-        </Box>
-        <Box flow={3}>
-          {partners.map(card => (
-            <PartnerCard key={card.id} t={t} card={card} />
-          ))}
-        </Box>
+const PartnersPage = ({ t }: Props) => {
+  const isDark = useTheme('colors.palette.type') === 'dark'
+  const partners = getPartners(isDark)
```
gre: I was not aware this was a breaking change in live-common. we need to be careful about breaking changes in future.

gre:
I'll merge this because needs to gets in with latest live-common version. we'll test on develop.

PR #2396, opened 2019-10-18
added an helper to manage automatically good contrast on currency colors

IAmMorrow:
This helper / proxy implement the lib-common `getCurrencyColor` function and provide a well contrasted color if the contrast threshold is not enough.

### Type

Ui Polish

### Context

N / A

All currency colors places

### DIFF
--- a/src/helpers/getCurrencyColor.js
+++ b/src/helpers/getCurrencyColor.js
@@ -1,0 +1,11 @@
+// @flow
+
+import { getCurrencyColor as commonGetCurrencyColor } from '@ledgerhq/live-common/lib/currencies'
+import type { Currency } from '@ledgerhq/live-common/lib/types/currencies'
+import ensureContrast from 'helpers/ensureContrast'
+
+export const getCurrencyColor = (currency: Currency, bg: ?string) => {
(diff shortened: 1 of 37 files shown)


IAmMorrow:
@Arnaud97234 This should do the trick, feel free to test it with many different currencies to see if it works every time !

src/helpers/getCurrencyColor.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+import { getCurrencyColor as commonGetCurrencyColor } from '@ledgerhq/live-common/lib/currencies'
+import type { Currency } from '@ledgerhq/live-common/lib/types/currencies'
+import ensureContrast from 'helpers/ensureContrast'
+
+export const getCurrencyColor = (currency: Currency, bg: ?string) => {
+  const currencyColor = commonGetCurrencyColor(currency)
+
+  return bg ? ensureContrast(currencyColor, bg) : currencyColor
```
gre: could be great to have this in live-common

gre: + memoize

PR #2405, opened 2019-10-21
Custom fee regression bitcoin family + translations + other fixes

juan-cortes:
### Fee translation + customs
Oct-21-2019 12-21-14
Bring back the custom label for fees on bitcoin family and make those labels translatable. 

> Flag is not there in the commit, trust me, it's just to show it's getting it from the translation file 🤷‍♂  Also flashes are only in develop due to the bug on the fonts we have.

### Sendmax  toggle glitch
subtle
There was an anomaly on the rendering of the send max toggle button caused by it being wrapped in a fragment. No idea what the underlying error is but removing the wrap fixed it.

### Type

Regression + bug fix

### Context

LLD 2019 10 21: Current Regression in develop doc

Send flow for bitcoin like currencies

### DIFF
--- a/src/components/base/Input/index.js
+++ b/src/components/base/Input/index.js
@@ -51,12 +51,26 @@ const LoadingDisplay = styled(Box)`
   left: 0px;
   top: 0px;
   bottom: 0px;
+  width: 100%;
   background: ${p => p.theme.colors.palette.background.paper};
   pointer-events: none;
   flex-direction: row;
   align-items: center;
   padding: 0 15px;
   border-radius: 4px;
+`
+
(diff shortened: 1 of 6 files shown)


MortalKastor:
`HIGHLIGHT_I18N=1` can also be used to (ungraciously) highlight translated strings, FYI:
image

juan-cortes:
> `HIGHLIGHT_I18N=1` can also be used to (ungraciously) highlight translated strings, FYI:
> image

Nice! This is one of those things I knew and keep forgetting. We need to explain all those flags somewhere in the readme.ms

gre:
good for now. we'll work on this bug in the future

PR #2411, opened 2019-10-21
Keep local state of user selected item for bitcoin-like fees

juan-cortes:
Sometimes the value for low/standard/high fees might overlap so it would appear that the user selection is not respected. This local state will show the selected item as the one the user choses even if there is a previous item in the options that matches the value.

### Type

UX

### Context

https://ledgerhq.atlassian.net/browse/LL-1970

Send flow bitcoin-like

### DIFF
--- a/src/families/FeeField/BitcoinKind.js
+++ b/src/families/FeeField/BitcoinKind.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { useCallback, useMemo } from 'react'
+import React, { useCallback, useMemo, useState } from 'react'
 import { BigNumber } from 'bignumber.js'
 import styled from 'styled-components'
 import { Trans, translate } from 'react-i18next'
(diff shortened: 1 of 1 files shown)


gre:
That's weird

juan-cortes:
That's a bad cache build not related to the pr for sure. `yarn install --check-files` should fix it

src/families/FeeField/BitcoinKind.js
```diff
@@ -48,6 +48,7 @@ const FeesField = ({ transaction, account, onChange, status }: Props) => {
 
   const bridge = getAccountBridge(account)
   const { feePerByte, networkInfo } = transaction
+  let input: ?HTMLInputElement
```
gre: You should prefer using a ref.

useRef and pass that ref to ref prop.
Check react doc

juan-cortes: Gotcha

juan-cortes:
@Arnaud97234 makes no sense to me that it would behave differently for 1000 than 1, but I'll check 🤷‍♂

PR #2419, opened 2019-10-23
Firmware identifier fix

gre:
A flash of seeing Identifier before the OSU is even uploaded to the device was caused by a recent regression.


gre:
@Arnaud97234  i tried it from 1.4.2. maybe you can test from a 1.5.5 and i think it should be good

PR #2420, opened 2019-10-24
integrate tezos ui

valpinkman:
Finish all remaining little UI fixes for the release of Tezos

### Type

UI Polish

### Context

LL-1733

### Parts of the app affected / Test plan

OperationList/OperationDetails
AccountPage
AccountsPage
...

PR #2432, opened 2019-10-30
Use the deleteLibcore function when resetting libcore.

phaazon:
This call will make future withLibcore calls to actually recreate the
Core structure.


gre:
shipped in #2420

PR #2433, opened 2019-10-30
LL-1650 Send flow rework 

juan-cortes:
Big rework of the send flow in order to make it friendlier, it involves changes in the whole flow steps, wordings, styles, it removes the double modal for high fees, etc. It also takes advantage of the introduced debouncing of the bridge status so any loading steps will be shown on the continue button.

image
Shows recipient-related fields, in this case, the tag for XRP

image

image
Loading state for the continue button

image
image
New high fees warning, replacing the modal

image


package.json
```diff
@@ -122,6 +122,7 @@
     "styled-components": "^4.3.2",
     "styled-system": "^2.2.9",
     "uncontrollable": "^6.0.0",
+    "use-debounce": "^3.1.0",
```
gre: not used

gre:
Check CI

juan-cortes:
> I love it :)

It is quite a lot nicer to go through :D

gre:
@Arnaud97234 

> To be consistent with mobile we could hide that info from Summary

which info?

> Fees amount should be displayed using gwei to ETH conversion instead of Gwei and using the corresponding unit (ETH instead of ENG for instance)

I think in fact we need to diverge away from this idea.
- We should use the same unit for all the summary part (the user account prefered unit) so user can easily "do the math".
- the device step MUST be in the currency unit because that's what displayed on the device.

gre:
sorry, i misunderstood that.

juan-cortes:
@Arnaud97234 current version should show the correct units for the erc20 tokens

image
image

PR #2439, opened 2019-11-04
fix incorrect import from live-common

valpinkman:
### Type

Fix


gre:
as it's clearly not affect runtime code (only flow typing), i'm merging

meriadec:
awesome it was working before

gre:
yeah for types it makes sense, craziest is on mobile we were having a bug because /src/ where used for some utility.

we're going to add some custom linter script to detect this in future

PR #2443, opened 2019-11-05
live-common & libcore bump

gre:
### QA notes:

- this is a bump of live-common that removed deprecated functions of the bridge (that we should have fully moved on). You might want to test a bit the send flow.
- This includes latest fixes from Libcore regarding XTZ send.
- to be tested if the Experimental XTZ works.

### Dev notes:

@valpinkman this is the work you need for what you are working on.

```js
import {loadAccountDelegation, listBakers} from "@ledgerhq/live-common/lib/families/tezos/bakers";
import whitelist from "@ledgerhq/live-common/lib/families/tezos/bakers.whitelist-default";

const delegation = await loadAccountDelegation(account)
...

const bakers = await listBakers(whitelist);
...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "@ledgerhq/hw-transport-node-hid": "4",
     "@ledgerhq/hw-transport-node-hid-singleton": "4",
     "@ledgerhq/ledger-core": "^3.4.0",
-    "@ledgerhq/live-common": "^8.6.0",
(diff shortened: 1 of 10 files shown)


gre:
@Arnaud97234 :

- Bug 1 and 2 should be fixed.
- Bug 3 will be fixed in another PR by juan and is not to consider in this PR.

PR #2445, opened 2019-11-06
Experiment: improve LL user flows designs

IAmMorrow:
The following is an experimentation to figure out a simpler / cleaner way to design ledger live user flows. It use the `xstate` library.

I reimplemented the actual receive flow using those principle and the result looks great to me !

The actual receive logic can be _visualised_ here: https://xstate.js.org/viz/?gist=5b57391de24311d21dd065efac95a79e

(This tools actually allow you to interact with the machine using a chart, this is **very useful**)

xstate is a very recent library, and already has 8k stars on github.

All the flow's logic is implemented in a machine located at `Receive2/receiveFlow.js`. This machine is very easy to test and could potentially be shared in live-common as the mobile use the same logic, only the UI change. Such machines are easily testable using the `@xstate/test` package. This is a big plus as it allow us to test user flows without having to deal with the UI.

I left the original receive flow for comparaison (`Receive/index.js`)

The UI itself is implemented and bound to the flow through a hook (useMachine) provided by `@xstate/react`

xstate's docs for the curious: https://xstate.js.org/docs/


gre:
to be rediscussed indeed, but personally I think we should also kill any sort of "connected react-router" and I wonder if we couldn't embrace more react-router features (using more `<Switch>`,  `<Route>` and stuff).
making modals using that as well as steps. now i don't know if it can work with the data size.

I can see the power in the lib you used, but i can also see the technical debt it could bring and some extra complexity in this big description object.
There is a big rework on Send flow at the moment so don't go farther than this PoC i would say but we should definitely gather the different ideas.

I don't see that current approach is such a big problem, ok we have big props object for all steps, which somehow is a bit like redux that pass-in the whole state from above. reducing that state could interesting tho. I think Send did that recently with the transaction/transactionStatus unification. 
To me the biggest problem to solve and improve is to have good typing, and not a big blob of (maybes) `{ ?Account, ?Transaction }`, where obviously at step 3 you know that Account has settled.

IAmMorrow:
I believe routers features are efficient whenever there is URL and / or history stuff involved, else it does not add a lot versus a map key/components. It's just JSX versus JS and In my experience, using it everywhere is an anti-pattern but let's discuss it !.

I know that relying too much on a library can be dangerous, but this seems pretty sane to me, the whole implementation could easily be replaced by (more complex) reducers, everything is pure, there are no side effects / mutations at all and it is fully separated from the UI so easy to replace at any moment.

I know that this configuration object can feel impressive at the first glance, but in the end it's really easy to read, every state has pre-defined ways to go to a new state, that's it. It can even allow composition as at the end of the day, it's just a plain simple serializable js object ! It's actually a lot like reducers.

To me, the current approach is really really really complicated but let's discuss it :)

We will have more and more of those flows in the future (staking, lending, activating, swapping, ...) and I do not see how we can keep maintaining those without a little bit of automatization as currently, complexity = nbr of steps ^ 2 because everything is in the root component ...

gre:
ok. why not. what's sure it's pretty clean when compared to what we come from.
Just don't do it now on Send flow, because there are new system involved in here: https://github.com/LedgerHQ/ledger-live-desktop/blob/adb394096a1f85b0fefe3321ac8bf7261401d7d1/src/components/modals/Send/Body.js#L59-L113 (excludeFromBreadcrumb thing)

to me, most of these flow are pretty linear and don't necessarily need a big system in place, but i agree that it can be handy when we need such edgecases & also send/receive are i think suitable to have enough complexity to have these.

I think where it would REALLY shine tho, is on the onboarding. The codebase is worse than the receive/send flow and really would deserve such system, so many edgecases and different flows. It could be great **when we rework it** to reuse that.

As for Receive flow, i'm not opposed to eventually move to this.

IAmMorrow:
I agree

I won't merge it as it is merely a POC, it is here for the team to review.

IAmMorrow:
I've thought of even more use cases, it could also shine a lot regarding device software updates flows (with all the boot loader recovery stuff). Also, we could version machine and load them in function of device version, this way we can have adaptative flows without needing too many if / else in UI components.

gre:
will be revived later :) closing for https://github.com/LedgerHQ/ledger-live-desktop/pull/2596 to gets in

PR #2448, opened 2019-11-07
LL-1939 Styling of select component

juan-cortes:
Style overhaul for the select component option's states. Since the introduction of the custom `menuOption` component for performance on lists including erc20 tokens, we've been experiencing a glitchy render where all rows were not quite the right height. This PR addresses tgat and the linked issue LL-1939 where the selected option did not highlight on hover/keyboard-over.

### Type

UI Polish

### Context
https://ledgerhq.atlassian.net/browse/LL-1939

All select components (Account selection, currency selection, language, etc)

### DIFF
--- a/src/styles/palette.js
+++ b/src/styles/palette.js
@@ -13,6 +13,8 @@ module.exports = {
       shade60: 'rgba(20,37,51, 0.6)',
       shade40: 'rgba(20,37,51, 0.4)',
       shade20: 'rgba(20,37,51, 0.2)',
+      shade10: 'rgba(20,37,51, 0.1)',
+      shade04: 'rgba(20,37,51, 0.04)',
     },
     divider: 'rgba(20,37,51, 0.1)',
     background: {
@@ -41,6 +43,8 @@ module.exports = {
       shade60: 'rgba(255, 255, 255, 0.6)',
       shade40: 'rgba(255, 255, 255, 0.4)',
       shade20: 'rgba(255, 255, 255, 0.2)',
+      shade10: 'rgba(255, 255, 255, 0.1)',
+      shade04: 'rgba(255, 255, 255, 0.04)',
     },
     divider: 'rgba(255, 255, 255, 0.1)',
     background: {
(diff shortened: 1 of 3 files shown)


src/styles/palette.js
```diff
@@ -69,6 +73,8 @@ module.exports = {
       shade60: 'rgba(255, 255, 255, 0.6)',
       shade40: 'rgba(255, 255, 255, 0.4)',
       shade20: 'rgba(255, 255, 255, 0.2)',
+      shade10: 'rgba(255, 255, 255, 0.1)',
+      shade04: 'rgba(255, 255, 255, 0.04)',
```
IAmMorrow: rgba(255, 255, 255, 0.04) is really close from the default background, we should not add it as it is really specific and the theme will start to go out of hands again. let's use background.default !

if you have exceptional color needs that won't be used anywhere else, consider using the color library to darken / lighten it

juan-cortes: image

juan-cortes: 🤷‍♂

IAmMorrow: Marouane is trying to cheat, but we need some consistency :) background.default is super close to .04 really

juan-cortes: I don't have an issue changing from `.04` to` background.default` I try to follow what the designer gives me and in this case it was an `.04` :trollface: 

Would it be possible to choose some names that are more consistent? I find it confusing that they are called "_text_" or "_divider_" and used for backgrounds if you know what I mean.

juan-cortes: @dekkaki

IAmMorrow: Yes this is right, "text" and "divider" definitly came out of hand. I suppose "text" should be renamed and divider replaced by shade10. We'll improve the theme a bit soon :)

juan-cortes:
The current version dropped the use of .04 in favor of background.default as per @IAmMorrow 's request to keep the palette simple. I'd say it's ok to test.

juan-cortes:
@gre why is this still here?

PR #2465, opened 2019-11-14
Polishes and bug fixes for send flow

juan-cortes:
Polishes to the send flow

image

- Fixed the back action on the Device step of the send flow to actually go back to Summary instead of Recipient steps.
- Polish of fonts (size/weight/color) for labels
- Changed cursor for inactive input app wide
- Increased opacity of placeholders
- Accessibility of header actions, they are now accessible through tabs
- New styles for the gwei slider value

### Type

UI Polish + Bug fix

Send flow, but also affects all inputs and modals, maybe there's a broken modal header somewhere 🤞

### DIFF
--- a/src/components/FeeSliderField.js
+++ b/src/components/FeeSliderField.js
@@ -35,6 +35,16 @@ const Error = styled.div`
   }
 `
 
+const Holder = styled.div`
+  font-family: Inter;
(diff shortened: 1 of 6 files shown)


meriadec:
**nice** focus style!

PR #2472, opened 2019-11-18
Fix action-less header regression on modals + stylings

juan-cortes:
This addresses a regression where the header styles were broken if there were no actions (ie only title) on a modal. Also there were some broken styles (hover/active on header actions) and we've reworked which steps are dismissible on the send flow.

### Type

UI Polish

Modals in general


gre:
i've tested it. works fine.

PR #2473, opened 2019-11-18
Cleanup svg id defs and add a collision check script

juan-cortes:
This pr provides a new script to find svg collisions in the ids defined inside `<defs>`. This usually happens when we use a tool to cleanup a raw svg and it gives ids such as `a` or `shape` to the definitions. Then sometimes we get a collision and it ends up triggering something like this which, say what you want but doesn't make the nanoX look good.

image

### Type

Code Quality Improvement

### Context

https://ledgerhq.atlassian.net/browse/LL-2043

Overall svgs in case I broke one.

### DIFF
--- a/scripts/svg-id-collisions.sh
+++ b/scripts/svg-id-collisions.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+allMatches=()
+failed=0
+
+find src/icons/ -type f -name '*.js' | while read f; do
+  if grep -q "<defs>" "${f}"; then
+    matches=($(grep -Eo 'id="[a-zA-Z0-9_\-]+"' ${f} | cut -c4- | sed 's/"//g'))
+    for match in "${matches[@]}"
+    do
+      if [[ " ${allMatches[@]} " =~ " ${match} " ]]; then
+        failed=1
+        echo "SVG def with id '${match}' in ${f} is not unique."
+      else
+        allMatches+=(${match})
+      fi
+    done
+  fi
+  if [ $failed -eq 1 ]; then
+    echo "┌────────────────────────────────────────────────────────────────────┐"
+    echo "│      The above warnings represent cases where an id collision      │"
+    echo "│     has been detected. This can potentially break the rendering    │"
+    echo "│           of the SVG, and we don't want that now, do we?           │"
+    echo "│                                                                    │"
+    echo "│               Add a prefix to the ids inside <defs>                │"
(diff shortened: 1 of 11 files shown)


scripts/svg-id-collisions.sh
```diff
@@ -0,0 +1,30 @@
...
+
+allMatches=()
+failed=0
+
+find src/icons/ -type f -name '*.js' | while read f; do
+  if grep -q "<defs>" "${f}"; then
+    matches=($(grep -Eo 'id="[a-zA-Z0-9_\-]+"' ${f} | cut -c4- | sed 's/"//g'))
+    for match in "${matches[@]}"
+    do
+      if [[ " ${allMatches[@]} " =~ " ${match} " ]]; then
```
gre: :bashmaster:

src/icons/device/NanoXBanner.js
```diff
@@ -6,7 +6,7 @@ export default ({ size = 30, ...p }: { size: number }) => (
   <svg viewBox="0 0 6 16" height={size} width={size} {...p}>
     <defs>
       <path
-        id="a"
+        id="NanoXBanner-a"
```
gre: to me, for all ids you have added we need to actually also template it with a random id (per component mount).

one trick would be to use a ref.

```
const id = useRef(_globalId++)

id={`${id}_NanoXBanner-a`}
```

scripts/svg-id-collisions.sh
```diff
@@ -0,0 +1,30 @@
...
+      else
+        allMatches+=(${match})
+      fi
+    done
+  fi
+  if [ $failed -eq 1 ]; then
+    echo "┌────────────────────────────────────────────────────────────────────┐"
+    echo "│      The above warnings represent cases where an id collision      │"
+    echo "│     has been detected. This can potentially break the rendering    │"
+    echo "│           of the SVG, and we don't want that now, do we?           │"
```
valpinkman: XD

scripts/svg-id-collisions.sh
juan-cortes: :googleandtrialanderrormaster: i would say

src/icons/device/NanoXBanner.js
juan-cortes: feel free to improve upon this! :D

scripts/svg-id-collisions.sh
juan-cortes: Condescending mofo that script

PR #2475, opened 2019-11-18
added snow event

IAmMorrow:
Added snowflakes behind modals at the end of the year
use SNOW_EVENT=1 to test it

### Type

Feature ?

### Context

N / A

Modal backdrops

### DIFF
--- a/src/components/extra/Snow.js
+++ b/src/components/extra/Snow.js
@@ -1,0 +1,119 @@
+// @flow
+
+import React, { useLayoutEffect, useRef, useState } from 'react'
+import times from 'lodash/times'
+import each from 'lodash/each'
+import throttle from 'lodash/throttle'
+
+export const isSnowTime = () => {
+  const now = new Date()
+
+  return process.env.SNOW_EVENT || (now.getMonth() === 11 && now.getDate() > 21)
+}
+
+function randomBetween(min, max, round) {
+  const num = Math.random() * (max - min + 1) + min
+
+  if (round) {
+    return Math.floor(num)
+  }
+  return num
+}
+
+function Flake(x, y) {
+  const maxWeight = 5
+  const maxSpeed = 3
+
+  this.x = x
+  this.y = y
+  this.r = randomBetween(0, 1)
+  this.a = randomBetween(0, Math.PI)
+  this.aStep = 0.01
+
+  this.weight = randomBetween(2, maxWeight)
+  this.alpha = this.weight / maxWeight / 2
+  this.speed = (this.weight / maxWeight) * maxSpeed
+
+  this.update = () => {
+    this.x += Math.cos(this.a) * this.r
+    this.a += this.aStep
+    this.y += this.speed
+  }
+}
+
+type Props = {
+  numFlakes: number,
+}
+
+const Snow = ({ numFlakes }: Props) => {
+  const canvasRef = useRef()
+  const containerRef = useRef()
+
+  const [dimension, setDimension] = useState({ width: 0, height: 0 })
+
+  useLayoutEffect(() => {
+    const updateDimension = () => {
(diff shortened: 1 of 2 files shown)


src/components/extra/Snow.js
```diff
@@ -0,0 +1,123 @@
...
+    )
+
+    canvas.width = width
+    canvas.height = height
+
+    const updateFlakes = () => {
+      ctx.save()
+      ctx.setTransform(1, 0, 0, 1, 0, 0)
+      ctx.clearRect(0, 0, width, height)
+      ctx.restore()
```
gre: only

```
      ctx.clearRect(0, 0, width, height)
```

is needed.

rest is not needed because you don't use any transformation.

src/components/extra/Snow.js
```diff
@@ -0,0 +1,123 @@
...
+    const canvas = canvasRef.current
+    const ctx = canvas.getContext('2d')
+
+    const flakes = times(
+      numFlakes,
+      () => new Flake(randomBetween(0, width, true), randomBetween(0, height, true)),
+    )
+
+    canvas.width = width
+    canvas.height = height
```
gre: to me you do not want to do that if it doesn't changes. setting again the width/height is not free and cause canvas to also reset (clear the rendering)
unless it's intended?

src/components/extra/Snow.js
```diff
@@ -0,0 +1,123 @@
...
+          flake.y = -flake.weight
+        }
+      })
+      requestAnimationFrame(updateFlakes)
+    }
+
+    const loopId = requestAnimationFrame(updateFlakes)
+
+    return () => {
+      if (loopId) {
```
gre: you don't need to if, it's always settled.

src/components/extra/Snow.js
```diff
@@ -0,0 +1,123 @@
...
+  const containerRef = useRef()
+
+  const [dimension, setDimension] = useState({ width: 0, height: 0 })
+
+  useLayoutEffect(() => {
+    const ro = new ResizeObserver(
+      throttle(() => {
+        setDimension(containerRef.current.getBoundingClientRect())
+      }),
+      250,
```
gre: i think a parenthesis was missed here, the 250 is to be a parameter of throttle i guess? because right now when you resize it's def not throttled :D

src/components/extra/Snow.js
```diff
@@ -0,0 +1,123 @@
...
+        ctx.beginPath()
+        ctx.arc(flake.x, flake.y, flake.weight, 0, 2 * Math.PI, false)
+        ctx.fillStyle = `rgba(255, 255, 255, ${flake.alpha})`
+        ctx.fill()
+
+        if (flake.y >= height) {
+          flake.y = -flake.weight
+        }
+      })
+      requestAnimationFrame(updateFlakes)
```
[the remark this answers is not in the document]
IAmMorrow: Yes this looked a bit too magical to me !

src/components/extra/Snow.js
IAmMorrow: Ok, good to know !

src/components/extra/Snow.js
IAmMorrow: LOL

src/components/extra/Snow.js
IAmMorrow: not intended but this hook will be re-renderer on dimension change most of the time, doesn't cost me to much to optimize it thought :)

gre: ok, we don't need to change it then. i didn't see it was only on dimension changes

src/components/extra/Snow.js
gre: same for ctx.save and ctx.restore !

IAmMorrow:
I added some opacity, seems to work pretty well in light and dark.

PR #2478, opened 2019-11-19
Fixes screen rendering on device frames

juan-cortes:
There were some styles linking directly to the ids defined in the path defs, this basically broke the positioning of the screens within the frame. 

### Type

UI Bug


gre:
to me btw, we really should think about ids as unique per mount instance. unless there are no variable part in the reused id sure, but it's a bit dangerous when we'll start displaying multiple svg at same time.

juan-cortes:
I have no idea how to do that @gre

PR #2489, opened 2019-11-22
created a script that check that we have wordings for every errors

IAmMorrow:
Added a CI script that will test:
- if we have errors exported in @LedgerHQ/errors that do not have a corresponding wording defines (or a missing title).
- if we have wording for errors that does not exist (anymore ?).

### Type

QA Script

### Context

LL-2047

Error wordings; CI scripts


gre:
we need a way to blacklist some error that we don't need in desktop project and/or are falling into generic case. to me you can put already:

- Error
- CustomError
- BluetoothRequired
- TransportWebUSBGestureRequired
- LedgerAPIError and LedgerAPIErrorMessage, interestingly I don't really see where it's used today so it can be ignored as well.

for DeviceSocketFail/LedgerAPI4xx, we just need to set `title: {{message}}` , basically we just forward the socket fail error at the moment so it displays on the UI.



For the remaining errors, we need to add the wording
- please first check if it's in mobile project and just import the wording from there
- otherwise we will ask DS at the end.

IAmMorrow:
@gre Here we are, only the errors lacking wording are now reported.

PR #2490, opened 2019-11-22
bootstrap tezos delegation

gre:


gre:
we'll merge this work so we can do internal testing.

PR #2493, opened 2019-11-25
WIP Manager v2. Initial work in progress version for pr draft

juan-cortes:
### Things that are todo
- Sorting by name, marketcap, dependencies
- ~Update all button logic~
- ~Loading flow is real bad, needs to perhaps skip the genuine check or do it all in one step instead of two~
- Lots of code cleanups that are needed
- Testing in other themes to see what breaks
- Possible performance boost for the list, it's a it weird

### Known issues
- ~The uninstalling/installing wording is broken and will always say "Uninstalling" right now~
- ~There's a highlight on the search input when selected, needs to be removed~

### Things that are supposedly working (from the hackathon code mostly)
- Dependency install/uninstall 😍 
- Searching by ticket/name
- Listing apps, sizes, support status, etc
- Individual update, install, delete
- Queueing
- DeviceStorage UI
- Experimental toggle access to the manager
- Device/Catalog toggle
- Tooltips

### Screenshots
image
image

### Type

Feature

### How to test this
You'll need a yalc'ed version of common with an added flag for the manager v2, this would no longer be needed if https://github.com/LedgerHQ/ledger-live-common/pull/409 gets merged.


gre:
we'll take that work in to test it more with everything else.

PR #2502, opened 2019-11-28
fixed error stack; fixed theme; tweaked display

IAmMorrow:
### Type

Bug Fix

### Context

N / A

App error screen

### DIFF
--- a/src/components/RenderError.js
+++ b/src/components/RenderError.js
@@ -24,8 +24,26 @@ import IconTriangleWarning from 'icons/TriangleWarning'
 // SERIOUSLY plz refactor to use <ResetButton>
 import { IconWrapperCircle } from './SettingsPage/ResetButton'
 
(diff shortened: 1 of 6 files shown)


src/components/RenderError.js
```diff
@@ -24,8 +24,22 @@ import IconTriangleWarning from 'icons/TriangleWarning'
 // SERIOUSLY plz refactor to use <ResetButton>
 import { IconWrapperCircle } from './SettingsPage/ResetButton'
 
-const printError = (error: mixed) => `${String(error)}
-${String((error && error.stack) || 'no stacktrace')}`
+const printError = (error: mixed) => {
+  const print = []
+
+  if (error.message) {
```
gre: please assume error could be falsy. `throw null` or shit like that. we never knows.

gre: curious that flow didn't complain btw

PR #2560, opened 2019-12-14
Update live-common to includes websocket workaround

gre:
includes https://github.com/LedgerHQ/ledger-live-common/pull/422

This is required for mobile android but we need to check all is fine to move on to this update.

### Type

Bug fix

### Context

LL-2084

Everything that involve a websocket connection with our script runners (genuine check, install, uninstall, firmware update)

- we need to test the normal feature works
- we need to test the error cases still behave like on develop branch
- test it against the normal script runner and the staging script runner


gre:
not working properly. more work will be needed

PR #2567, opened 2019-12-22
Complete Hungarian Translation

Stabil21:
_**Complete Hungarian Translation**_


gre:
that's awesome I would say let's merge it. I'll discuss with the team on monday, we will have to enable on Crowdin (we'll also try to enable a bunch of languages in a near future)

Stabil21:
I'm very glad you liked it, and I was able to help.

PR #2573, opened 2019-12-23
Updated Hungarian Translation

Stabil21:
Fixes in app.json


gre:
thanks. develop is now in sync with these. We can now directly follow up on Crowdin 🚀

PR #2599, opened 2020-02-20
(ManagerV2): add account banner issue + updating label

LFBarreto:
add account banner was showing in case pof updates, now fixed + updating label

### Type

Bug Fix, UI Polish

### Context

Manager v2


static/i18n/en/app.json
```diff
@@ -406,6 +406,7 @@
         "versionNew": "Version {{version}} (NEW{{newVersion}})",
         "installing": "Installing...",
         "uninstalling": "Uninstalling...",
+        "updating": "Updating...",
         "scheduled": "Scheduled",
```
dasilvarosa: Queued

static/i18n/en/app.json
```diff
@@ -406,6 +406,7 @@
...
         "update": "Update available",
         "updateAll": "Update All",
```
dasilvarosa: Update all

PR #2604, opened 2020-02-21
Spectron

nabil-brn:
### Description

- [x] Spectron setup
- Upgraded to v10 to make it compatible with electron v8
- Folder structure (po / spec / data / setup) to organize tests (W.I.P)
- [x] Page Object Pattern 
- Give an abstraction layer to avoid always maintain the test suite in case of SUT updates
- `this.app.client` is available in all page objects
- If webdriverio is bumped in spectron project, it will be easier to upgrade, the syntax is pretty similar: v4 ([current](https://github.com/LedgerHQ/ledger-live-desktop/blob/7909b768f41c4f890d7adbce31e30437790894d2/test-e2e/po/onboarding.page.js)) | v5 (https://webdriver.io/docs/pageobjects.html)

- Coverage
- [x] Onboarding (I think I go through 100% of the onboarding screens)
Capture d’écran 2020-03-10 à 20 48 33

- [ ] Visual regression testing (W.I.P)

### Give it a try
```
yarn dist
yarn jest tests/specs/onboarding.spec.js
```
Mar-10-2020 18-05-26

### Type
UI Testing Framework

New automation IDs have been set on some components, reviewing those IDs could be a good idea, maybe we can place/rename them in another way.


src/renderer/components/CurrencyBadge.js
```diff
@@ -95,7 +95,12 @@ function CurrencyBadge({ currency }: { currency: CryptoCurrency | TokenCurrency
         >
           {currency.ticker}
         </Box>
-        <Box ff="Inter" color="palette.text.shade100" fontSize={5} data-e2e="currencyBadge">
+        <Box
+          ff="Inter"
+          color="palette.text.shade100"
+          fontSize={5}
+          data-automation-id="currency-badge"
```
gre: I recommend to just use `id` and `className`s (classes)

- you benefit the fact react/ the DOM will warn you when there are duplicated ids
- it is a generic concept and valid for the purpose of identifying **unique** elements
- it's easier to select because you can just `#id`

nabil-brn: I agree with you, I didn't want to overlap potential dev ids, search & replace incoming then 👍.

Don't mind about this PR for the moment, I'm migrating those commits from lld-v2.

nabil-brn:
I rollbacked all the ci commits, this PR is ready to merge, let me know if you me to squash the commits ?

tests/data/onboarding/index.js
```diff
@@ -0,0 +1,25 @@
...
+  saveSeedTitle: "Write down your recovery phrase",
+  enterSeedTitle: "Enter your recovery phrase",
+  genuineModalTitle: "Genuine check",
+  genuinePinErrorTitle: "Didn't choose your own PIN code?",
+  genuinePinErrorDesc:
+    "Never use a device supplied with a PIN code. Please contact us when in doubt.",
+  genuineSeedErrorTitle: "Didn't save your recovery phrase by yourself?",
+  genuineSeedErrorDesc:
+    "Only save a recovery phrase that was displayed on your device and to you alone. Please contact us when in doubt.",
+  noDeviceTitle: "Don't have a Ledger device?",
```
gre: note for later. we shouldn't maintain this but dynamically reverse it with `i18n`. this will be a dev task.

gre:
I've squashed merged it.

PR #2610, opened 2020-02-24
added some tweaks on the sidebar starred list buttons

IAmMorrow:
tweaked a bit the star buttons, added hover effect and reworked paddings


gre:
No more drag & drop i'm afraid.
The feature was removed in context of putting the starred information back with the Account data, that way the stars now follow the normal account ordering.

If we want to reboot the "user defined ordering" in the future, the way to do it is via a new sorting mode that would keep account order "as is" and allow a way to drag & drop account between them. I say the priority for such feature is pretty low tho.

PR #2613, opened 2020-02-25
LL-2224 Move the NotifComponent to top-right corner when collapsed

juan-cortes:
Moved the `NotifComponent` a bit so that it's no longer out of the sidebar. I don't like this catch-all style inside `Box` (the reason why it was getting the left margin)  but we'll have to deal with it for now. Had to wrap it further to restyle it.
https://github.com/LedgerHQ/ledger-live-desktop/blob/19b092792100ec2944e2f55569054b91d8c30192/src/renderer/components/Box/Box.js#L84-L87 

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-2224

image

Enable the `DEBUG_UPDATE` flag to get the update debug UI visible, and set a state that has a notification component, then collapse the sidebar.

### DIFF
--- a/src/renderer/components/MainSideBar/index.js
+++ b/src/renderer/components/MainSideBar/index.js
@@ -246,7 +246,7 @@ const MainSideBar = () => {
                 iconActiveColor="wallet"
                 onClick={handleClickDashboard}
                 isActive={location.pathname === "/"}
-                NotifComponent={noAccounts ? undefined : UpdateDot}
+                NotifComponent={noAccounts ? undefined : <UpdateDot collapsed={collapsed} />}
                 disabled={noAccounts}
                 collapsed={secondAnim}
               />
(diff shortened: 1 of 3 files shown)


bscholving:
Can we have something like this?

Screenshot 2020-02-21 at 14 19 34

bscholving:
Can we have something like this?

Screenshot 2020-02-21 at 14 19 34

juan-cortes:
**YES WE CAN** (more or less).
I have no idea how to make it less abrupt for the opening animation, if this is ok... then we can go. Otherwise, maybe another dev can make it nicer 🤷‍♂ 

cc @bscholving 

### Closeup 
image

### Low def animation to see the jump
Feb-25-2020 20-16-31

bscholving:
Can we just avec a basic fade transition where the bullet disappear/appear ?  @juan-cortes

juan-cortes:
> Can we just avec a basic fade transition where the bullet disappear/appear ? @juan-cortes

@LFBarreto and I tried to get something animated and due to the way the collapse is notified to the dot the best we can do is to always have it on the top right corner, no animation. I can't spend more time on this 😖!

The original bug was the dot being outside of bounds, if we want to change the styling we'll have to address it in a separate issue unless someone else wants to take it.

asd

gre:
can you rollback to the version you shown me @juan-cortes I don't think we want the top right usage in unexpand mode.

please confirm @bscholving ; we need a simple solution first. no easy fix to offer a nice animation, at least we have a fix of the positionning

LFBarreto:
ezgif-7-3a1293839acf

heres the final result validated with @bscholving

PR #2625, opened 2020-02-26
remove text when button is in loading state

valpinkman:
### Type

UI Fix

Screen Recording 2020-02-26 at 15 14 04

### DIFF
--- a/src/renderer/components/Button.js
+++ b/src/renderer/components/Button.js
@@ -371,8 +371,9 @@ class Button extends PureComponent<
           <LoadingWrapper>
             <Spinner size={16} />
           </LoadingWrapper>
-        ) : null}
-        {children}
+        ) : (
+          children
+        )}
       </Base>
     );
   }
--- a/src/renderer/components/Button.js
+++ b/src/renderer/components/Button.js
@@ -262,6 +262,11 @@ const LoadingWrapper = styled.div`
   align-items: center;
   justify-content: center;
 `;
+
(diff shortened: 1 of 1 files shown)


gre:
this solution is not ideal and was previously discarded by Design because the button width blinks.

gre:
it would be great to make a gif of all diff possibility :
- the existing one
- your approach
- a variant version of the approach which would guarantee the width don't change (we can use opacity and keep absolute centering)

can you try the 3rd option? @valpinkman

juan-cortes:
> it would be great to make a gif of all diff possibility :
> 
> * the existing one
> * your approach
> * a variant version of the approach which would guarantee the width don't change (we can use opacity and keep absolute centering)
> 
> can you try the 3rd option? @valpinkman

That third option is literally what we used to have, the version I did hid the text and showed the spinner. We had that on production for some time I'm 92% sure of it :trollface:

valpinkman:
## Existing
CleanShot 2020-02-26 at 17 51 38

## Proposal 1
Screen Recording 2020-02-26 at 15 14 04

## Proposal 2
CleanShot 2020-02-26 at 17 47 58

juan-cortes:
Please let's just go back to proposal 2 for my mental health's sake.

valpinkman:
> Please let's just go back to proposal 2 for my mental health's sake.

Proposal 2 is accepted by @aramab

gre:
works fine. this really feels great btw.

PR #2630, opened 2020-02-26
LL-1118 Use user password to encrypt libcore db

MortalKastor:
Done by extending our brand new `InternalProcess` singleton to keep track of the state of libcore and feed it the password when needed.

- [x] Encrypt an unencrypted db
- [x] Work with an encrypted db with the user provided password
- [x] Remove password encryption (decrypt db)
- [x] Change encryption password
- [x] Fallback for the case the app password doesn't match libcore one, by dropping libcore db and start a new one encrypted with the app password.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1118
(reimplementation of #2142)

### Parts of the app affected

- Setting/changing a password: it now additionally affects libcore.
- Unlocking the app: it now also send the password to libcore for decryption.

#### Test plan

- Setting a password in the app should encrypt libcore db
- Disabling password should decrypt libcore db
- Changing the password should also be reflected in libcore db
- Unlocking the app should also unlock libcore
- Updating from an older version of Ledger Live with a password should encrypt libcore db with the password on first unlock

To check the current status of the libcore db encryption, just open the sqlite file with [DB Browser for SQLite](https://sqlitebrowser.org/), and use the defaults encryption settings (SQLCipher 4 defaults)

### DIFF
--- a/src/internal/live-common-setup.js
+++ b/src/internal/live-common-setup.js
@@ -1,6 +1,6 @@
 // @flow
 import "~/live-common-setup";
-import "./implement-libcore";
+// import "./implement-libcore";
 
 import { throwError } from "rxjs";
 import usbDetect from "usb-detection";
--- a/src/internal/live-common-setup.js
+++ b/src/internal/live-common-setup.js
@@ -1,6 +1,5 @@
 // @flow
 import "~/live-common-setup";
-// import "./implement-libcore";
 
 import { throwError } from "rxjs";
 import usbDetect from "usb-detection";
--- a/src/main/InternalProcess.js
+++ b/src/main/InternalProcess.js
@@ -92,16 +92,15 @@ class InternalProcess {
   changePassword(newPassword: string) {
     logger.info("Changing Libcore password", logType);
 
-    if (!this.libcoreInitialized) {
+    if (!this.libcoreInitialized || !this.process) {
       throw new Error("Can't change password if libcore is not unlocked first!");
     }
 
-    this.process &&
-      this.process.send({
-        type: "changeLibcorePassword",
-        currentPassword: this.password,
-        newPassword,
-      });
+    this.process.send({
+      type: "changeLibcorePassword",
+      currentPassword: this.password,
+      newPassword,
+    });
 
     this.password = newPassword;
   }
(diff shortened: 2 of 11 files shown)


src/internal/live-common-setup.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import "~/live-common-setup";
-import "./implement-libcore";
+// import "./implement-libcore";
```
gre: you can remove the code :)

src/main/InternalProcess.js
```diff
@@ -1,120 +1,212 @@
...
-          if (!msg) return;
-          if (process.env.INTERNAL_LOGS) console.log(msg);
-          try {
-            const obj = JSON.parse(msg);
-            if (obj && obj.type === "log") {
-              logger.onLog(obj.log);
-              return;
+      logger.info(`spawned internal process ${pid}`, logType);
+
+      this.process && // A bit stupid, but Flow complains otherwise
```
gre: would it help to make `process` a tmp variable of this scope so you don't need to check this? I think this happens because flow can't prove you didn't re-set this.process in another part of the code but if you do a local `const` it will not make you do this.

this would help on reducing the diff of this PR here.

src/main/InternalProcess.js
```diff
@@ -123,13 +215,13 @@ class InternalProcess {
...
 
-      logger.info(`ending process ${pid} ...`);
-      console.log(`ending process ${pid} ...`);
+      logger.info(`ending process ${pid} ...`, logType);
       this.active = false;
-      this.process.once("exit", () => {
-        resolve();
-      });
-      this.process.disconnect();
+      this.process &&
```
gre: same here you could have `const {process} = this` at the beginning of the function to prove to flow only once.

src/main/index.js
```diff
@@ -86,6 +86,21 @@ app.on("ready", async () => {
...
   });
 
+  ipcMain.handle("setLibcorePassword", async (event, { password }) => {
+    try {
+      await internal.setPassword(password);
+      if (!internal.process) {
+        await internal.start();
+      }
+    } catch (e) {
+      throw new Error("wrong password");
```
gre: should we make a custom error for this? it's displayed or used programmatically no?

src/renderer/components/IsUnlocked.js
```diff
@@ -91,6 +93,15 @@ const IsUnlocked = ({ children }: Props) => {
       try {
         if (!isAccountDecrypted) {
           await setEncryptionKey("app", "accounts", inputValue.password);
+          try {
+            // The password worked for db, it should be the same for libcore
+            await setLibcorePassword(inputValue.password);
+          } catch (e) {
```
gre: i'm not sure about the idea to generically catch all error cases. we should specifically expect some error cases no?

IAmMorrow:
Congrats for finally getting over this subject 👏 👏 👏 

The code looks 🆗  to me, but I believe this has to be seriously battle tested by QA before going any further !

src/main/InternalProcess.js
MortalKastor: Makes sense 👍

gre:
@Arnaud97234 can you write a task on LLC to reflect the issue we have found?

gre:
We'll close this PR for now. There are indeed complication on Windows that we saw, might be an issue with the underlying libcore library. discussion to continue internally.

PR #2632, opened 2020-02-26
(ManagerV2): search bar issue

LFBarreto:
Search bar on manager uses same component as accounts page + some refactoring and fixes on translation wording on this component
+ fix warning on dropdown compopnent

### Type

Bug Fix

### Test plan

Search bar in both accounts and manager page was changed: 
* Verify that search works
* in manager search input should be cleared on each tab changes

Drop-down should not have been impacted but still test that they work in both filters and sort on manager page 👍

### DIFF
--- a/src/renderer/components/DropDown.js
+++ b/src/renderer/components/DropDown.js
@@ -165,7 +165,17 @@ class DropDown extends PureComponent<Props> {
   };
 
   render() {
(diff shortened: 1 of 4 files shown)


gre:
I think the QA on this is quite important based on the modified code i see. so i'll set a QA flag.

PR #2633, opened 2020-02-26
fixed visual glitch on breadcrumb

IAmMorrow:
Repro: Receive flow => I don't have my device => Breadcrumb stepper has a visual glitch as described here: https://docs.google.com/spreadsheets/d/1IsuT6FQNaLrVZa_dViiTKp5NL1_lpqe8U9gM2s9k2Ns/edit#gid=228270868&range=A1

### DIFF
--- a/src/renderer/components/Stepper/Breadcrumb.js
+++ b/src/renderer/components/Stepper/Breadcrumb.js
@@ -31,6 +31,7 @@ const Bar: ThemedComponent<{
   height: 1px;
   left: ${p => p.start}%;
   position: absolute;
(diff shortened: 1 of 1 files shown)


IAmMorrow:
@gre this is not what I fixed in this PR, there was a visual glitch on the left (see the screenshot on the sheet)

I agree that the breadcrumb is also not behaving properly, but I guess it would need a proper rework because right now it feel a bit overcomplicated to me :)

gre:
@IAmMorrow ok makes sense. can you please create a jira task for what we discussed here? thanks

PR #2635, opened 2020-02-26
Adding dutch translation

momo258:


gre:
Hi @momo258 ,
I'm sorry this was not yet enough documented, but we are currently using Crowdin to do translations https://crowdin.com/project/ledger-wallet

There are already pending translations on Crowdin that are waiting to be proof read. I will discuss with @dasilvarosa who originally did them to see if we can also validate them so they get back in crowdin. 

I'll make sure to document that in a README.md next to the translation files in the future. This PR can be used later to double check if we get them matching but for now we'll close it.

Thanks

PR #2640, opened 2020-02-27
(ManagerV2): learn more link supportURL

LFBarreto:
ManagerV2

learn more link when we have an installed and not live supported app now links to the app.suppportURL provided by our API and fallback to the regular support apps page.

### Type

UI Polish

### Context

Manager v2

Once the support URL field is filled we can test it but this should work out of the box without any problem.

### DIFF
--- a/src/renderer/screens/manager/AppsList/AppActions.js
+++ b/src/renderer/screens/manager/AppsList/AppActions.js
@@ -84,7 +84,7 @@ const AppActions: React$ComponentType<Props> = React.memo(
     isLiveSupported,
     addAccount,
   }: Props) => {
-    const { name } = app;
+    const { name, supportURL } = app;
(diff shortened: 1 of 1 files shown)


gre:
@LFBarreto  I don't think we want to do this for now. we will only understand the implication when we put actual data, i'm not confortable having something that settle in the future, without having done a pretest check. (because it's not testable here)

I also don't remember exactly what we planned for `supportURL`. but i remember we had the possibility to have a `mailto:` kinda url too (but maybe it was more contactURL).

I think we should wait we do a "manager 2.1" with all new fields on the table and rediscuss this.

If you don't mind i think we should close for now and focus on sprint tasks that were planned.

LFBarreto:
@gre ok i discussed with fabrice for his team to fill in these fields and once they are ready we can test this out :+1: 
closing this for now

PR #2641, opened 2020-02-27
Add Experimental Explorers API settings

gre:
this would enable btc v3 explorers which is our upcoming version that fix bech32 issue for some users

Capture d’écran 2020-02-27 à 10 07 53

### Type

explorer migration

### Context

LL-2103

when you enable this feature, BTC accounts should still sync properly and all features related to it, like sending, should works the same.

### DIFF
--- a/src/renderer/experimental.js
+++ b/src/renderer/experimental.js
@@ -53,6 +53,13 @@ export const experimentalFeatures: Feature[] = [
   },
   {
     type: "toggle",
+    name: "EXPERIMENTAL_EXPLORERS",
+    title: "Experimental Explorers API",
+    description:
+      "Try an upcoming version of Ledger's blockchain explorers. Changing this setting may affect the account balance and synchronization as well as the send feature.",
+  },
+  {
+    type: "toggle",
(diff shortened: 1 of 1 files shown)


src/renderer/experimental.js
```diff
@@ -51,6 +51,13 @@ export const experimentalFeatures: Feature[] = [
     title: "Experimental device flow",
     description: "Automatically asks an Open App device permission in device flows",
   },
+  {
+    type: "toggle",
+    name: "EXPERIMENTAL_EXPLORERS",
+    title: "Experimental Explorers API",
+    description:
+      "Try an incoming version of Ledger explorers. Changing this value can affect the Account balance, synchronisation and send feature.",
```
dasilvarosa: "Try an upcoming version of Ledger's blockchain explorers. Changing this setting may affect the account balance and synchronization as well as the send feature.",

gre: thanks! done

gre:
we'll merge this. there seems to be some broadcast issue still. but it's experimental so it's expected.

PR #2643, opened 2020-02-27
reverted renderer process reuse to false

IAmMorrow:
Electron's allowRendererProcessReuse flag got its default value (false) deprecated in the V8. Turning it to true seems to change a lot of low level behaviors. As it doesn't really benefit us at the moment, I propose we turn it false (that will get rid of the deprecation warnings) and reconsider turning it on after the release.


gre:
we discussed and i agree it's safer to go back to this settings. it should be safe to do this and we'll test more in the next days.

PR #2644, opened 2020-02-27
Fix: The date on the tooltip on the price chart is one day off

LukeSugiura:
### Outcome

Screen Shot 2020-02-27 at 12 04 46 PM

### DIFF
--- a/src/renderer/components/Chart2/index.js
+++ b/src/renderer/components/Chart2/index.js
@@ -37,6 +37,7 @@ import React, { useRef, useLayoutEffect, useState, useMemo } from "react";
 import ChartJs from "chart.js";
 import styled from "styled-components";
 import Color from "color";
+import moment from "moment";
 
 import useTheme from "~/renderer/hooks/useTheme";
 import Tooltip from "./Tooltip";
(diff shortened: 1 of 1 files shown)


valpinkman:
Graph is broken:

CleanShot 2020-02-27 at 16 17 15

src/renderer/components/Chart2/index.js
```diff
@@ -115,28 +125,6 @@ const Chart = ({ height, data, color, renderTickY, renderTooltip, valueKey = "va
         display: false,
       },
       scales: {
-        xAxes: [
```
valpinkman: why remove this ?

src/renderer/components/Chart2/index.js
```diff
@@ -67,7 +67,15 @@ const ChartContainer: ThemedComponent<{}> = styled.div.attrs(({ height }) => ({
...
 
-const Chart = ({ height, data, color, renderTickY, renderTooltip, valueKey = "value" }: Props) => {
+const Chart = ({
+  height,
+  data,
+  color,
+  renderTickY,
+  renderTooltip,
+  valueKey = "value",
+  dateFormat = "MMM D",
```
juan-cortes: Where is this used?

LukeSugiura: Good question. I was wondering too. I saw it exists in Prop type but no component uses this prop. I'm just gonna delete it for now.

juan-cortes:
I'd say squash them all together and we're good 😕

PR #2647, opened 2020-02-27
(Managerv2): updated completion label

LFBarreto:
Manager v2

After a completed update the update list now shows an updated label

### Type

UI Polish

After a series of updates you should see a updated label on the succesfully updated apps

Make sure this label only shows in that case.


gre:
reminder to not double ✔️ approve from devs please, so we can make sure we need 1 dev approval and 1 qa approval.

if you want in future you can just comment :)

we need a @github feature to define groups in future 🙏

PR #2648, opened 2020-02-27
[Mac] Fix notarization failing for Apple IDs with more than 1 provider

MortalKastor:
As pointed out by @valpinkman, `yarn release` will fail on Mac at the notarization part if the Apple ID used is a member of multiple teams or organizations.

This fix that by hardcoding the provider to use Ledger's one (which, for historical reasons, has `EpicDreamSAS` as a shortname. It's _totally_ not a hack attempt on my part, promise 😇)

### Type

Bug Fix

Doing in a release.
On Valentin's Mac.


valpinkman:
🔥

PR #2650, opened 2020-02-27
fixed lock timeout

IAmMorrow:
fixed lock issues by making it simpler


LFBarreto:
Tested it works as intended :+1: 
@IAmMorrow  @gre I made a PR over this branch to implement a LockGuard component that should disable lock capabilities at critical points such as updates in manager 
https://github.com/IAmMorrow/ledger-live-desktop/pull/1

PR #2653, opened 2020-02-28
(DeviceAction): disconnect glitch after opening app

LFBarreto:
**DeviceAction**

Added .5s extra debounce when device disconnect after opening an app.

### Type
Bug Fix

Device connection modals and manager as well
It should not have side effect since this just adds extra debounce time on reducer state change in case of device disconnection.

try adding an account select the app.
connect your device following the instructions and check if after opening the app the modal goes to step 3 instead of showing the "connect your device animation"

### DIFF
--- a/src/renderer/components/DeviceAction/actions/app.js
+++ b/src/renderer/components/DeviceAction/actions/app.js
@@ -195,6 +195,10 @@ const useHook = (device: ?Device, appRequest: AppRequest): AppState => {
             // no debounce for allow event
             return empty();
           }
+          if (!s.device) {
+            // extra debounce for the disconnect event
+            return interval(2000);
+          }
           // default debounce (to be tweak)
           return interval(1500);
         }),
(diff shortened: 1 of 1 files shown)


src/renderer/components/DeviceAction/actions/app.js
```diff
@@ -195,6 +195,10 @@ const useHook = (device: ?Device, appRequest: AppRequest): AppState => {
             // no debounce for allow event
             return empty();
           }
+          if (!s.device) {
+            // extra debounce for the disconnect event
+            return interval(2000);
+          }
           // default debounce (to be tweak)
           return interval(1500);
```
gre: just change the 1500 to 2000 I think ?

LFBarreto: To me all this debouncing is a bit hacky to begin with i dont get why we receive a device disconnection after opening an app in the first place.

generalizing the debounce just worsen responsiveness on all other actions but we can set it at 2s all the time if you want.

gre: I agree that the hardware is hacky. it disconnect and reconnect when you enter an app. so yeah we have no other choices but to debouce this to not have the user seeing it.

LFBarreto: just set it by default to 2s :+1:

gre:
this is safe enough to merge.

PR #2654, opened 2020-02-28
make export operation work + remove fs from renderer

valpinkman:
Export operations was not working anymore

Removed FS calls from renderer and fix the export operations button

### Type

Fix

### Context

@gre 

Settings => Accounts => Export Operations


valpinkman:
OK on it @nabil-brn

PR #2657, opened 2020-02-28
do not show success if you cancel the save file dialog

valpinkman:
### Type

Fix

### DIFF
--- a/src/main/setup.js
+++ b/src/main/setup.js
@@ -34,9 +34,17 @@ ipcMain.handle("save-logs", async (event, path: { canceled: boolean, filePath: s
 ipcMain.handle(
   "export-operations",
   async (event, path: { canceled: boolean, filePath: string }, csv: string) => {
-    Promise.resolve().then(
-      () => !path.canceled && path.filePath && csv && fsWriteFile(path.filePath, csv),
-    );
+    try {
+      if (!path.canceled && path.filePath && csv) {
+        await fsWriteFile(path.filePath, csv);
+      }
+
+      if (path.canceled) {
+        throw new Error("canceled");
+      }
+    } catch (error) {
+      throw new Error("canceled");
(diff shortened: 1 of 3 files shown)


src/renderer/modals/ExportOperations/index.js
```diff
@@ -137,7 +139,7 @@ class ExportOperations extends PureComponent<Props, State> {
             ) : (
               <Box>
                 <IconWrapperCircle>
-                  <IconDownloadCloud />
+                  <IconDownloadCloud size={42} />
```
nabil-brn: #### Icon size
30px seems to be the correct ratio box/svg

```suggestion
                  <IconDownloadCloud size={30} />
```

#### Export operation
Error thrown when canceling the operation
```
Error occurred in handler for 'export-operations': Error: canceled
    at eval (webpack-internal:///./src/main/setup.js:44:11)
    at electron/js2c/browser_init.js:5822:52
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3851:42)
    at WebContents.emit (events.js:210:5)
```

---
2b7244d7 (HEAD -> fix/export-operations-csv-fixes, val/fix/export-operations-csv-fixes) do not show success if you cancel the save file dialog

src/main/setup.js
```diff
@@ -33,10 +33,21 @@ ipcMain.handle("save-logs", async (event, path: { canceled: boolean, filePath: s
...
+      if (!path.canceled && path.filePath && csv) {
+        await fsWriteFile(path.filePath, csv);
+        return true;
+      }
+
+      if (path.canceled) {
+        return false;
+      }
+
+      return false;
```
MortalKastor:
```suggestion
```

This part has no use ("if canceled return false, otherwise return false, then return false" 🤔)

PR #2661, opened 2020-03-02
LL-2237 (Manager): progress update hooked on observable

LFBarreto:
Manager:

progress bar for installs now uses the new observable synced hook from live-common
introduced here:

https://github.com/LedgerHQ/ledger-live-common/pull/540

### Type

Polish

### Context

LL-2237

### Parts of the app affected / Test plan

Manager

Try an install or many and check that the progress bar updates correctly

PR #2667, opened 2020-03-04
update device connect imports

valpinkman:
- removed the DeviceConnect logic from desktop and moved it to live-common
- Updated live-common to use the DeviceConnect logic from live-common

### Type

Code Improvement

### Context

LL-2223

Everywhere DeviceConnect is used (manager, txs, add account....)


gre:
I'm setting a HODL because I think https://github.com/LedgerHQ/ledger-live-desktop/pull/2661 needs to go first as the 12.2.1 came with some changes for Manager (not that it would be breaking but still great to bump in order)

gre:
@valpinkman can you rebase? why is it HODL?

gre:
@valpinkman sorry to ask again but you can rebase again (and probably squash into one commit), you no longer need to say "bump live common" because 12.5.0 is now on develop 🚀

valpinkman:
Will do @gre

PR #2684, opened 2020-03-05
LL-2269 Fixes bug in add account internal logic

juan-cortes:
With the introduction of some logic that was in charge of preselecting the first new account scanned when all others had already been added we introduced a bug that would mark as _checked_ the last account that was scanned regardless of whether it was added or not already. This resulted in the flow for add accounts being changed and showing the `Add Account` (because internally there was this **one** account to be added).

This fixes that internal counter and should show "Close" as the CTA when we will effectively add/create no accounts.

### Type

Bug Fix

### Context

https://ledgerhq.atlassian.net/browse/LL-2269

On the add account flow, when adding accounts 
- If any account will be created/added, the CTA should say add account/s.
- If no checkmarks are visible & selected, the CTA should say "Close" and end the flow.
- If all accounts have been added before & the last one is not empty, it should list it **and auto-select it**
- If all accounts have been added before & the last one **is empty** it should tell the user to receive funds on that account before creating a new one.

### Screenshot 
image

Example of a case where we'll kill the flow since we won't create more accounts.

### DIFF
--- a/src/renderer/modals/AddAccounts/steps/StepImport.js
+++ b/src/renderer/modals/AddAccounts/steps/StepImport.js
@@ -138,7 +138,9 @@ class StepImport extends PureComponent<StepProps> {
               setScannedAccounts({
                 scannedAccounts: [...scannedAccounts, account],
                 checkedAccountsIds: onlyNewAccounts
-                  ? [account.id]
+                  ? hasAlreadyBeenImported
+                    ? []
+                    : [account.id]
                   : !hasAlreadyBeenImported && !isNewAccount
                   ? uniq([...checkedAccountsIds, account.id])
                   : checkedAccountsIds,


juan-cortes:
The code that we had would only preselect the last one in that case. This comes from https://github.com/LedgerHQ/ledger-live-desktop/commit/d4a10c3fe3e1b3bb45dd1622180ee0364ff0b50a and it explicitly checks one account in that case. If we want something different we need a better definition of what the expected behavior is, because, to me, this is correct, we only preselect one the user can select the other one should he choose to 🤷‍♂ 

cc @gre

nabil-brn:
It could be correct, I've no idea which account we should preselect, but I mean at least we have to trigger input focus and checkbox preselection together on the same listed element, I might open a new issue ?

juan-cortes:
I'm not saying it's right either, I don't know. Let's wait for gre

gre:
The behavior is correct to me that we have focus on the first input. so you can do all with keyboard (input TAB check TAB input TAB check)  

in any case this discussion seems independent of the present fixes, isn't it?

PR #2690, opened 2020-03-06
LL-1878 Empty states for apps and accounts

juan-cortes:
### Empty state for no apps installed
Screenshot 2020-03-10 at 11 02 24

### Empty state for no accounts added
image

### Dev notes
Initially, the flag for showing this would be set to false to avoid triggering it for existing users, it will only be set to true in the case of a first onboarding where we `listApps` and find nothing. It will be set back to false ~on the `appOpExec` call from manager the first time we install an app. I couldn't find a cleaner way to do this.~ inside the manager once we see any apps installed for the first time, with `useEffect`.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1878

## About the `no-apps-empty-state`
- An existing user should **never** see the `no-apps-empty-state` after they've completed onboarding, regardless of whether they have apps installed or not.
- After the first onboarding, a new user will see the `no-apps-empty-state` **only** if the device connected during onboarding had no apps or we were unable to list apps on the device.
- After installing any app the `no-apps-empty-state` will be hidden, forever.
- Uninstalling all apps **will not** bring the `no-apps-empty-state` back.
- Restarting onboarding with the same or a different device **will not** bring the `no-apps-empty-state` back.

## About the `no-accounts-empty-state`
- If a user has no accounts added to live the should be redirected to the `Portfolio` screen and see the `no-accounts-empty-state`.
- If a user adds one or more accounts, they should no longer see the `no-accounts-empty-state`
- The `Accounts` sidebar menu item should be disabled and users should land in `Portfolio` any time the `no-accounts-empty-state` is visible.


src/renderer/screens/manager/Dashboard.js
```diff
@@ -33,9 +37,14 @@ const Dashboard = ({ device, deviceInfo, result, onReset }: Props) => {
...
-      command("appOpExec")({ appOp, targetId, app, devicePath: device.path }),
-    [device],
+    (appOp, targetId, app) => {
+      if (!hasInstalledApps && appOp.type === "install") {
+        dispatch(setHasInstalledApps(true));
+      }
+
+      return command("appOpExec")({ appOp, targetId, app, devicePath: device.path });
+    },
+    [device, dispatch, hasInstalledApps],
```
gre: I think we better not have the `exec` function getting recalculated each time hasInstalledApps changes. I think the `useAppsRunner`could badly glitch if this happen (probably in context of more than one install).

Also more generally I think it could be dangerous to react on an install action. I would instead react to seeing a apps instead `result.installed` (useEffect)

essentially, anytime a user plug a device with apps (even if it's another device) we unlock the situation

src/renderer/screens/onboarding/steps/GenuineCheck/index.js
```diff
@@ -76,12 +81,18 @@ const GenuineCheck = (props: StepProps) => {
...
-  const handleGenuineCheckPass = useCallback(() => {
-    updateGenuineCheck({
-      isDeviceGenuine: true,
-    });
-    setGenuineCheckModalOpened(false);
-  }, [updateGenuineCheck]);
+  const handleGenuineCheckPass = useCallback(
+    ({ result }: any) => {
+      if (!hasCompletedOnboarding && (!result || !result.installed.length)) {
+        dispatch(setHasInstalledApps(false));
```
gre: we might need to discuss more, but i wonder if we simply shouldn't **at any point in time** just track the "last seen app installed" and use that as a fact so if user ever re-uninstall all apps and he don't have accounts, we would correctly invite him to go back to Manager (and not add accounts, because we remembered you had no apps)

src/renderer/screens/dashboard/index.js
```diff
@@ -81,7 +84,9 @@ const DashboardPage = ({ saveSettings }: Props) => {
         totalCurrencies={totalCurrencies}
       />
       <Box flow={7}>
-        {totalAccounts > 0 ? (
+        {!hasInstalledApps ? (
+          <EmptyStateInstalledApps />
+        ) : totalAccounts > 0 ? (
```
gre: mmh to me this should be the first condition instead:

if user have totalAccounts, we really should display the normal portfolio.

it's only if user have no accounts that we then start displaying the relevant placeholder.

src/renderer/screens/accounts/index.js
```diff
@@ -83,17 +82,8 @@ const AccountsPage = ({
...
-        <TrackPage category="Accounts" accountsLength={accounts.length} />
-        <TopBannerContainer>
-          <UpdateBanner />
-          <MigrationBanner />
-        </TopBannerContainer>
-        <EmptyState />
-      </>
-    );
+  if (!accounts.length) {
+    return <Redirect to="/" />;
```
gre: I don't know about this. I think it's not good to be redirected to Portfolio when for instance, you just did a "Remove" on the last account.

src/renderer/screens/manager/Dashboard.js
juan-cortes: I initially had it as a `useEffect` to that exactly and couldn't get the setting to be stored for some reason so I resorted to this. If you think it should work like that I'll try to figure out why it wasn't doing it.

src/renderer/screens/dashboard/index.js
juan-cortes: will change or not once we all agree on a flow

src/renderer/screens/onboarding/steps/GenuineCheck/index.js
juan-cortes: will change or not once we all agree on a flow on monday

src/renderer/screens/accounts/index.js
juan-cortes: same

gre:
I feel we need a wording pass btw.

The two title

"Install an app on my device" VS "You don't have any accounts..." are inconsistent. Also personally when I read "You don't have any accounts..." I feel it's a bit aggressive / anxiogene. maybe just me?

juan-cortes:
wording still needs a review even if the flow is accepted, cc @dasilvarosa

static/i18n/en/app.json
```diff
@@ -179,11 +179,20 @@
   },
   "emptyState": {
     "dashboard": {
-      "title": "Install apps or add accounts",
-      "desc": "Open the Manager to install apps or add accounts if your Ledger device already has apps installed",
+      "title": "Install an app on my device",
```
dasilvarosa: Install apps on your device

static/i18n/en/app.json
```diff
@@ -179,11 +179,20 @@
...
       "buttons": {
-        "addAccount": "Add accounts",
-        "installApp": "Open Manager"
+        "installApp": "Install Apps",
+        "help": "Help"
+      }
+    },
+    "accounts":{
+      "title": "You don’t have any accounts…",
+      "desc": "Click on “Add account” to stat adding accounts in Ledger Live",
```
dasilvarosa: Add an account to start managing your crypto assets. You should have installed the apps for your crypto assets on your device.

static/i18n/en/app.json
```diff
@@ -179,11 +179,20 @@
...
+      "buttons": {
+        "addAccount": "Add account",
+        "installApp": "Install Apps on my device",
```
dasilvarosa: Go to Manager to install apps

static/i18n/en/app.json
```diff
@@ -179,11 +179,20 @@
...
+      "desc": "You need to install apps on your device to add accounts in Ledger Live. Go to the manager and install apps on your device.",
       "buttons": {
-        "addAccount": "Add accounts",
-        "installApp": "Open Manager"
+        "installApp": "Install Apps",
+        "help": "Help"
+      }
+    },
+    "accounts":{
+      "title": "You don’t have any accounts…",
```
dasilvarosa: Add an account to get started

juan-cortes:
Wording and screenshots reviewed and updated with @dasilvarosa.
cc @nabil-brn for qa for this

gre:
@juan-cortes can you squash the 2 commits?

juan-cortes:
> @juan-cortes can you squash the 2 commits?

you got it

src/renderer/screens/onboarding/steps/GenuineCheck/index.js
gre: ok so to my understanding you do not need to do this because it will happen anyway in the new `useEffect` inside AppsList. right?

PR #2691, opened 2020-03-06
Automatically hide/show empty token accounts when toggling the setting

juan-cortes:
instahide

We had a regression where toggling the setting for hiding/showing empty token accounts didn't reflect the changes without exiting the screen and coming back/refreshing. This triggers a reorder of accounts and therefor shows the changes instantly.

### Type

Regression fix

### DIFF
--- a/src/renderer/actions/settings.js
+++ b/src/renderer/actions/settings.js
@@ -7,6 +7,7 @@ import type { Currency } from "@ledgerhq/live-common/lib/types";
 
 import { setEnvOnAllThreads } from "./../../helpers/env";
 import type { SettingsState as Settings } from "./../reducers/settings";
+import { refreshAccountsOrdering } from "~/renderer/actions/general";
 
 export type SaveSettings = ($Shape<Settings>) => { type: string, payload: $Shape<Settings> };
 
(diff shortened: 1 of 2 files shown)


juan-cortes:
weird, I'll check

juan-cortes:
> So there was no fix needed for the card view in the end, right?

No, in fact, it's a matter of cache and React not rendering again the `TokensList`, that's I connected it to monitor that env. But @gre was saying that perhaps we are doing something _fishy_ by wrapping it in a reducer instead of directly calling `getEnv`. In any case, this fix should work for now, we can figure out that later.

gre:
when discussing internally with @juan-cortes , I've noticed something to improve in our code, on two aspects:

- `listSubAccounts` rely on that env, but as we need that env to be dynamic here, it does not enforce an easy way to re-render. we probably should have some sort of `useSubAccounts()` that `useEnv` and make sure it re-renders the UI.
- we seems to have a duplicated state in our settings store to track that settings and I'm not sure why. we shouldn't maintain our envs in both a user settings and a redux. but maybe i don't see the big picture. (i thought we were using localStorage to save envs, but maybe it was just experimental?)

MortalKastor:
> we probably should have some sort of `useSubAccounts()` that uses `useEnv` to make sure it re-renders the UI.

👍

> i thought we were using localStorage to save envs, but maybe it was just experimental?

Indeed, only experimental settings are saved to localStorage. This setting not being experimental and intended to be exported with the rest through LiveQR™️ we decided to have it in the settings store, but since it relies on an env variable to work, we have this duplicated state. I'm not saying it's the right thing to do, but that's how it happened.

nabil-brn:
Sorry @MortalKastor for dismissing your review, I didn't commit anything 🤨.

PR #2693, opened 2020-03-09
LL-2254

LFBarreto:
Support for :deciduous_tree:  accounts page

### Type

Feature

### Context

LL-2254

**ask me before testing**

Add a :deciduous_tree:  account and review the accounts page design.


src/renderer/ModalsLayer.js
```diff
@@ -44,7 +44,7 @@ const ModalsLayer = ({ visibleModals }: *) => (
       <BackDrop state={state}>
         {/* {// Will only render at the end of december
           isSnowTime() ? <Snow numFlakes={200} /> : null} */}
-        {visibleModals.map(({ name, data }, i) => {
+        {visibleModals.map(({ name, ...data }, i) => {
```
gre: what is the impact of this change on existing stuff? if you can pinpoint one example of thing that got fix with that to make sure QA covers it

LFBarreto: this should not impact the existing stuff since the props on our current modals aren't accessed at the root of each modal.
This is used however in the new delegation root once we use it on both tezos and :deciduous_tree:

src/live-common-set-supported-currencies.js
```diff
@@ -30,4 +30,5 @@ setSupportedCurrencies([
   "decred",
   "bitcoin_testnet",
   "ethereum_ropsten",
+  "tron",
```
gre: you meant to comment it back right?

LFBarreto: my bad sorry :hankey:

PR #2695, opened 2020-03-09
Fixes main logs to appear in exported logs + more db logs

gre:
if you input a wrong password and export logs, you will now see more info:

Capture d’écran 2020-03-09 à 14 36 44

not much to test and impact risk is low


gre:
rebased

PR #2700, opened 2020-03-10
LL-2267 Images: prevent right clicks

LFBarreto:
prevent right click on images

This should fix the issue where users could right click image resources and save them.

### Type

Bug Fix

### Context

LL-2267

Try right clicking on the logo in the password page or any logo on the manager catalog list or any baker image on a Tezos validator selection.

### DIFF
--- a/src/renderer/index.html
+++ b/src/renderer/index.html
@@ -1,97 +1,96 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
-  <head>
-    <meta charset="utf-8" />
-    <title><%= htmlWebpackPlugin.options.title %></title>
-    <script
-      async
-      src="https://resources.live.ledger.app/public_resources/analytics.min.js"
-    ></script>
-    <style>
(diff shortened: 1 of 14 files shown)


src/renderer/index.html
```diff
@@ -1,94 +1,97 @@
...
     </div>
-  </div>
-  <div id="modals"></div>
-</body>
-
-</html>
\ No newline at end of file
+    <div id="modals"></div>
+  </body>
+</html>
```
gre: please undo this file change, it is not related to the commit. (if it is needed, should be another commit)

gre: mmh, can we avoid the big diff? edit the file without prettier.

if we want to prettier it, we should do it in another "previous" commit to make commit more atomic.

LFBarreto: :+1:

gre:
squash the two commits please :)

PR #2702, opened 2020-03-10
LL-2259 Wording: sync error wording update + link added to error banner

LFBarreto:
Wording updated for SyncError 
+ added a support link to the description shown on the error banner.

### Type

wording

### Context

LL-2259

Sync errors


gre:
mmh i'm not sure how this is addressing LL-2259 
and it's a bit hard to understand what was fundamentally the problem but the initial complain was that the error displayed was something like HTTP 500 . to be confirmed with techsupport

LFBarreto:
@gre i could not find any translation text referring to & 500 error so i figured it was the SynError and updated the error banner in order for it to show an error support Link.

gre:
my point is, i think in the actual case where api is down, it's not going to be a "SyncError". let's discuss more.

gre:
We can get this in as a first step. I like that it still settle a better wording and link. (we'll dig more for the second problem of error precisions).

dasilvarosa:
The title "Synchronization error" is superfluous in this format imho.

gre:
it's the current format of title+description. we can't put just the description. we could just use a title, but sometimes it's not always enough informative.

nabil-brn:
Are we waiting for changes on this PR or is it ready to review ?

gre:
to me it's mergeable

PR #2704, opened 2020-03-12
Skip genuine check for dev with MOCK flag

juan-cortes:
Not sure I understood completely, but with this, adding the `MOCK` flag and being in dev mode will skip the genuine check from onboarding in order to allow @nabil-brn to write further automation tests. If someone else wants to take over further be my guest.


gre:
title and description does not match implementation, but i'm good with the implementation.

ultimately we should solve this by mocking the listapps itself at some point (like in the command itself), but this shortcut is good fo now

juan-cortes:
> title and description does not match implementation, but i'm good with the implementation.
> 
> ultimately we should solve this by mocking the listapps itself at some point (like in the command itself), but this shortcut is good fo now

I changed it by in-browser editing the file once @nabil-brn told me you wanted to use the `MOCK` flag instead of the new `SKIP_GENUINE_CHECK` I will update the title and description for history-sake

PR #2706, opened 2020-03-12
LL-2277 ~ LL-2300

LFBarreto:
Updated transaction details for :deciduous_tree: 

### Type

Feature

### Context

LL-2277 ~ LL-2300

Updated :deciduous_tree: operations details look into the associated issue LL-2300 for details


src/renderer/families/tron/Votes/index.js
```diff
@@ -51,6 +51,19 @@ const useTronSuperRepresentatives = () => {
   return sp;
 };
 
+export const useFormattedVotes = (
+  votes: ?Array<any>,
+  superRepresentatives: ?Array<any>,
+): Array<any> => {
+  return votes
```
gre: this could be useMemo ed?

LFBarreto: I guess but this is more of a data formatter than a provider I'll rename this function sorry 😅

src/renderer/families/tron/Votes/index.js
```diff
@@ -52,6 +52,17 @@ const useTronSuperRepresentatives = () => {
   return sp;
 };
 
+// @TODO move this to common
```
gre: can we try to keep live-common up to date each time we have these new functions? could be used on mobile

LFBarreto: i'll make a PR with this :+1:

PR #2711, opened 2020-03-13
Stellar

henri-ly:
Add Stellar to ledger-live-desktop.

Feature that was added was :
- Memo field for Stellar
- Add 2 confirmations fields

----- 

Missing : 
- Translation

Have to wait for libcore to integrate Stellar or can ask me for a custom builded libcore.


src/renderer/families/stellar/MemoTypeField.js
```diff
@@ -0,0 +1,53 @@
...
+    memoType => {
+      setSelectedMemoType(memoType);
+      onChange(bridge.updateTransaction(transaction, { memoType: memoType.value }));
+    },
+    [onChange, bridge, transaction],
+  );
+
+  useEffect(() => {
+    setSelectedMemoType(options.find(option => option.value === transaction.memoType));
+  }, [transaction.memoType, options]);
```
gre: mmh, this makes me thing that `selectedMemoType` should not be a state but it should be a variable. when you have a state that depends on another state but essentially don't need that state to evolve BUT to follow the first state, it is not a state, it should be a "derived data" calculated / memoized from the initial state. 

would it work if we would:
- remove line `const [selectedMemoType, setSelectedMemoType] = useState(options[0]);`
- replace it with just: `const selectedMemoType = options.find(option => option.value === transaction.memoType) || options[0]`
- remove completely this effect

src/renderer/families/stellar/MemoTypeField.js
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <Select
+      width={"156px"}
```
gre: minor style perf: `width="156px"`

src/renderer/families/stellar/MemoTypeField.js
```diff
@@ -0,0 +1,53 @@
...
+}: {
+  onChange: string => void,
+  account: Account,
+  transaction: Transaction,
+}) => {
+  invariant(transaction.family === "stellar", "MemoTypeField: stellar family expected");
+
+  const bridge = getAccountBridge(account);
+
+  const options = StellarMemoType.map(type => ({
```
gre: move this in the global module scope. i mean before the `cost MemoTypeField = ({` so it's a static variable.

src/renderer/families/stellar/MemoValueField.js
```diff
@@ -0,0 +1,47 @@
...
+  // We use transaction as an error here.
+  // It will be usefull to block a memo wrong format
+  // on the ledger-live mobile
+  const { transaction: memoError } = status.errors;
+  const { transaction: memoWarning } = status.warnings;
+
+  return (
+    <Input
+      warning={memoWarning}
+      error={memoError}
```
gre: totally minor here but to make it clearer i would directly inline the stuff (i'm not usually fan of rewriting the var name in spread, for readability), like

```
<Input
  warning={status.warnings.transaction}
  error={status.errors.transaction}
```

src/renderer/modals/OperationDetails/index.js
```diff
@@ -161,8 +164,8 @@ const mapStateToProps = (state, { operationId, accountId, parentId }) => {
   const mainCurrency = parentAccount
     ? parentAccount.currency
     : account && account.type !== "TokenAccount"
-    ? account.currency
-    : null;
+      ? account.currency
```
gre: some weird prettier style code change on this file.

static/i18n/en/app.json
```diff
@@ -1583,6 +1591,18 @@
...
       "title": "Network fees represent more than 10% of the amount"
+    },
+    "StellarWrongMemoFormat": {
+      "title": "Memo format is wrong"
+    },
+    "StellarMinimumBalanceWarning": {
+      "title": "A minimum balance is required to be kept in a Stellar account"
+    },
+    "StellarNewAccountMinimumTransaction": {
+      "title": "A new account requires at least 1 XLM to be activated"
```
gre: this must not be in the wording. wording should be 

```
"A new account requires at least {{amount}} to be activated
```

which interestingly makes this wording generic. to me we could drop Stellar prefix?


but also, isn't this already existing for another coin @juan-cortes  ?

static/i18n/en/app.json
```diff
@@ -1583,6 +1591,18 @@
     },
     "FeeTooHigh": {
       "title": "Network fees represent more than 10% of the amount"
+    },
+    "StellarWrongMemoFormat": {
+      "title": "Memo format is wrong"
+    },
+    "StellarMinimumBalanceWarning": {
+      "title": "A minimum balance is required to be kept in a Stellar account"
```
gre: to me the currency name should not be in the wording, it's not for translation

```
"A minimum balance is required to be kept in a {{currencyName}} account"
```


which interestingly makes this wording generic. to me we could drop Stellar prefix?

static/i18n/en/app.json
```diff
@@ -1583,6 +1591,18 @@
...
+    },
+    "StellarNewAccountMinimumTransaction": {
+      "title": "A new account requires at least 1 XLM to be activated"
+    },
+    "StellarSourceHasMultiSign": {
+      "title": "Please disable multisign to send XLM"
```
gre: same, XLM need to be a variable

PR #2712, opened 2020-03-13
new action, test if user is allowed to trigger ci

valpinkman:
Prevent a certain user from triggering CI (for example when crowdin push a million update like yesterday, we might go over the top with our CI times)


valpinkman:
Since the action is local to the project, we need to install dependencies first.
HODL until I set it up as a standalone action

PR #2713, opened 2020-03-13
Spectron

nabil-brn:
Copy pasta incoming ...


src/renderer/screens/accounts/AccountGridItem/index.js
```diff
@@ -51,7 +51,7 @@ class AccountCard extends PureComponent<Props> {
           style={{ display: hidden && "none" }}
           p={20}
           onClick={this.onClick}
-          data-e2e="dashboard_AccountCardWrapper"
+          data-automation-id="dashboard_AccountCardWrapper"
```
gre: i feel i'm repeating but imo it just should be `id` and classes (`className`), because these make sense beyond automation. the generic concept of identifying elements can be used for more than test. also for styling, and also helpful for a dev to inspect the dom. easier to write selectors for it,...

package.json
```diff
@@ -166,7 +166,7 @@
     "listr": "^0.14.3",
     "listr-verbose-renderer": "^0.6.0",
     "prettier": "^1.19.1",
-    "spectron": "^9.0.0",
+    "spectron": "^10.0.1",
```
gre: can you try to also update Jest and Babel Jest libs? cc @valpinkman can help

src/renderer/screens/accounts/AccountGridItem/index.js
nabil-brn: Yes of course ! This PR is a duplicate, we had trouble with CI execution so we had to test with another PR.
All will be done in the first one 👍,  sorry for the confusion.

package.json
nabil-brn: Will be done in the first PR too 👍

PR #2724, opened 2020-03-17
LL-2277 ~ LL-2291

LFBarreto:
Added :snowflake:  flow to :deciduous_tree: 

You can now select which resource you want to freeze for and the amount.

You can only select an amount without decimals and you can quick select the amount with helper buttons that show if you have at least 4 :deciduous_tree: available to freeze.

Once you've frozen your assets theres a timer on the next flow button.

+ Added a popover component
+ Added a Toggle Button component

### Type

Feature

### Context

LL-2277 ~ LL-2291

:deciduous_tree: accounts that can freeze some assets


src/renderer/components/InputCurrency.js
```diff
@@ -116,8 +118,9 @@ class InputCurrency extends PureComponent<Props, State> {
     }
   }
 
-  handleChange = (v: string) => {
-    const { onChange, unit, value, locale } = this.props;
+  handleChange = (val: string) => {
+    const { onChange, unit, value, locale, decimals } = this.props;
+    const v = decimals === 0 ? val.replace(/[.,]/g, "") : val;
```
gre: mmh this is covered by the sanitizeValueString for a unit that is decimal=0. can you explain the usecase here?

src/renderer/modals/Freeze/fields/AmountField.js
```diff
@@ -0,0 +1,184 @@
...
+        disableRounding: true,
+        showAllDigits: false,
+        showCode: true,
+      }),
+    [spendableBalance, defaultUnit],
+  );
+
+  /** show amount ratio buttons only if we can ratio the available assets to 25% or less */
+  const showAmountRatio = useMemo(
+    () => spendableBalance.gt(BigNumber(4 * 10 ** defaultUnit.magnitude)),
```
gre: why is this hardcoded to 4 ? can we have a more explicit constants?

PR #2725, opened 2020-03-17
Ll 2152/enable sentry again

IAmMorrow:
This PR should enable sentry again. It use the old (deprecated) implementation called "raven" (we should update that someday).

We need to ensure that we build the app with the SENTRY_URL envvar set or sentry won't work.

### DIFF
--- a/src/logger/logger.js
+++ b/src/logger/logger.js
@@ -73,37 +73,31 @@ export function enableDebugLogger() {
 }
 
 const captureBreadcrumb = (breadcrumb: any) => {
-  // FIXME
-  /*
   if (!process.env.STORYBOOK_ENV) {
     try {
-      if (typeof window !== 'undefined') {
-        require('sentry/browser').captureBreadcrumb(breadcrumb)
+      if (typeof window !== "undefined") {
+        require("~/sentry/browser").captureBreadcrumb(breadcrumb);
       } else {
-        require('sentry/node').captureBreadcrumb(breadcrumb)
+        require("~/sentry/node").captureBreadcrumb(breadcrumb);
       }
     } catch (e) {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ledger-live-desktop",
   "productName": "Ledger Live",
   "description": "Ledger Live - Desktop",
-  "version": "2.0.1",
+  "version": "2.0.1-nightly.20200316T2208",
```
gre: that's weird to see this in the diff 🤔

src/logger/logger.js
```diff
@@ -73,37 +73,31 @@ export function enableDebugLogger() {
...
-  /*
   if (!process.env.STORYBOOK_ENV) {
     try {
-      if (typeof window !== 'undefined') {
-        require('sentry/browser').captureBreadcrumb(breadcrumb)
+      if (typeof window !== "undefined") {
+        require("~/sentry/browser").captureBreadcrumb(breadcrumb);
       } else {
-        require('sentry/node').captureBreadcrumb(breadcrumb)
+        require("~/sentry/node").captureBreadcrumb(breadcrumb);
```
gre: to be checked the implication of this in the runtime load (does it create an extra load time?). i guess it's fine, just to know if you checked a bit that.

IAmMorrow: Yes it will as it's not bundled. should be pretty small thought. I implemented **raven** exactly as it use to be, didn't take any time to optimize. we should take some time to upgrade to **sentry** and to clean a bit the implementation :)

gre:
let's merge this and we'll test a bit the app is fine on develop.

PR #2727, opened 2020-03-18
Fixes LL-2282 Fixes #2680

### LINKED ISSUE #2680: [Windows] Ledger Live 2.0.0 and 2.0.1 fails to boot with Internal Process Error – ledger-core-node.node load fail
**N.B.** this issue only track **Windows**, for Linux, please check status on https://github.com/LedgerHQ/ledger-live-desktop/issues/2670

underlying reason is it can't seem to load `ledger-core-node.node` file.

We are investigating on it. We know it happens for a few users.

### Providing logs could help us

could you send your logs (SAVE on the error screen) to our support team and mention this github issue? https://support.ledgerwallet.com/hc/en-us/requests/new?ticket_form_id=248165  🙏  ( alternatively if you don't want to send the logs , you can see the JSON yourself on https://ledger-live-tools.now.sh/logsviewer )

### Technical investigation are welcomed

Ideally we need to identify what is the "common denominator" (is it special computer? related to the OS? related to an antivirus? related to accounts you could have before upgrading?). We usually need a way to reproduce the problem so we can understand what bug there is to fix.

---

One thing you can attempt is to remove (or rename) the folder of the "Ledger Live" user data. The folder that is saved by Ledger Live at **`%APPDATA%\Roaming\Ledger Live`** (NB: which is not the installation folder) – it should work if you paste that in a file browser. Otherwise if you see hidden files, it's usually under `C:\Users\<username>\AppData\Roaming\Ledger Live` )

gre:
pending for confirmation that this really fix the bug

### Context

LL-2282
#2680

for our QA:
- app correctly works (basically you can add btc accounts) on Linux, Windows, Mac

for our internal beta program we will do for our users that want to confirm with us the bug is fixes:
- confirm that the 2.0.1 does not work at boot for you.
- confirm that this build now works for you and you can add btc accounts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "@ledgerhq/hw-transport": "^5.11.0",
     "@ledgerhq/hw-transport-http": "^5.11.0",
     "@ledgerhq/hw-transport-node-hid-singleton": "^5.11.0",
-    "@ledgerhq/ledger-core": "^6.1.1",
+    "@ledgerhq/ledger-core": "^6.2.0-alpha.4",
     "@ledgerhq/live-common": "^12.6.1",
     "@ledgerhq/logs": "^5.11.0",
     "@tippy.js/react": "^3.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1333,10 +1333,10 @@
     "@ledgerhq/errors" "^5.11.0"
     events "^3.1.0"
 
-"@ledgerhq/ledger-core@^6.1.1":
-  version "6.1.1"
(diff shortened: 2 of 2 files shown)


nabil-brn:
Sync / Send / Receive
[BTC](https://blockstream.info/tx/c98153de05f2a3488294dd9bf653fb0e851af004446075bca5924c7777ea8b84)

✓ QA-ed

**Platform**
Windows 10

---
6a88e8a2 (HEAD -> pr/2727, github-desktop-gre/ll-2282) Fixes LL-2282 Fixes #2680

gre:
oops. sorry. we're not yet good for the actual fix. i'll reopen if we have an update on a new libcore version.

PR #2730, opened 2020-03-18
Ll 2313/screen fadeout bug

IAmMorrow:
This PR should fix the screen fading out bug.


gre:
- QA can check everything still works fine (around the app loading time)
- also check that error cases at boot (if you know how to make one?) still display fine

PR #2736, opened 2020-03-19
LL-2277

LFBarreto:
Fixes for :deciduous_tree: claim rewards and freeze modals for them to use GenericSendStepDevice

also fixed conditions for reaning rewards under 1 :deciduous_tree:  of balance

### Type

Feature

### Context

LL-2277

Freeze and Claim rewards modal


src/renderer/families/tron/AccountHeaderActions.js
```diff
@@ -28,8 +30,14 @@ type Props = {
 
 const AccountHeaderActions = ({ account, parentAccount }: Props) => {
   const dispatch = useDispatch();
-  const { tronResources } = account;
+  const unit = getAccountUnit(account);
+  /** min 1TRX transactions */
+  const minAmount = 10 ** unit.magnitude;
```
gre: ⚠️  if you change the account unit. it will not be 1 TRX !

moreover, I strongly recommend to hardcode this in a constant (outside of the component) – as well as any other usage of this.

it should probably be:

```js
const minAccount = BigNumber(10).pow(getCryptoCurrencyById("tron").units[0].magnitude)
```

src/renderer/families/tron/Votes/index.js
```diff
@@ -85,7 +86,11 @@ const Delegation = ({ account, parentAccount }: Props) => {
   const superRepresentatives = useTronSuperRepresentatives();
   const nextVotingDate = useNextVotingDate();
 
-  const { tronResources: { votes, tronPower, unwithdrawnReward } = {} } = account;
+  const unit = getAccountUnit(account);
+  /** min 1TRX transactions */
+  const minAmount = 10 ** unit.magnitude;
+
+  const { tronResources: { votes, tronPower, unwithdrawnReward } = {}, spendableBalance } = account;
```
gre: i'm not fan of this magic code, I think it would be way more readable to do

```js
  const { tronResources, spendableBalance } = account;
  const { votes, tronPower, unwithdrawnReward } = tronResources || {};
```

is there a reason why `tronResources` would be falsy? i assume it would always be here, and if so I recommend to do `invariant(tronResources, "")` because i don't understand how flow can be happy here.

src/renderer/modals/ManageTron/index.js
```diff
@@ -112,19 +112,22 @@ type Props = {
 const ManageModal = ({ name, account, parentAccount, ...rest }: Props) => {
   const dispatch = useDispatch();
   const mainAccount = getMainAccount(account, parentAccount);
+  const unit = getAccountUnit(mainAccount);
+  /** min 1 TRX for frozen and unfreeze transactions */
+  const minAmount = 10 ** unit.magnitude;
```
gre: same!

static/i18n/en/app.json
```diff
@@ -696,6 +696,7 @@
   "tron": {
     "voting": {
       "manageTP": "Manage Tron Power",
+      "warnEarnRewards": "You need at least 1TRX to start earning rewards",
```
gre: this must not be in the translation. it should be `You need at least {{amount}} to start earning rewards` and there is a missing space (and also I think it's `TRX 1`)

src/renderer/families/tron/Votes/index.js
LFBarreto: You can get no tron resources if you're in a token account i would not use invariant for this reason this piece of UI would trigger an error and stop the app from showing the token account page :+1:

src/renderer/families/tron/AccountHeaderActions.js
LFBarreto: Ok i'll make a PR for this :+1:

LFBarreto:
@gre i HOLD this one i'll wait on the new release of common to fix these issues and update hooks in the meantime :+1:

LFBarreto:
changes for this are included in #2741 closing this :+1:

PR #2739, opened 2020-03-19
show terms and conditions from the about settings page

valpinkman:
Adds the Terms and Condition Modal to the Settings > About page

### Type

Feat

### Context

LL-2311 

Settings > About > Terms and Conditions

image

image

### DIFF
--- a/src/renderer/modals/Terms/index.js
+++ b/src/renderer/modals/Terms/index.js
@@ -19,7 +19,7 @@ import Modal from "~/renderer/components/Modal";
 import ModalBody from "~/renderer/components/Modal/ModalBody";
 import { closeModal } from "~/renderer/actions/modals";
 
-const TermsModal = () => {
+const TermsModal = ({ showClose = false }: { showClose?: boolean }) => {
   const dispatch = useDispatch();
 
   const [markdown, error] = useTerms();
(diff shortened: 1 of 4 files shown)


valpinkman:
admire the cool wallpaper ;)

src/renderer/modals/Terms/index.js
```diff
@@ -26,73 +26,97 @@ const TermsModal = () => {
...
-        title={<Trans i18nKey="Terms.title" />}
-        render={() => (
-          <>
-            <TrackPage category="Modal" name="Terms" />
+    <Modal
+      name="MODAL_TERMS"
+      preventBackdropClick
+      centered
+      render={(
+        { data }: { data?: { showClose: boolean } }, // Why u no get the correct type ?
```
LFBarreto: Hi! :wave: i'm pretty sure you can retrieve data you send during the from dispatch event right from the component props.
I updated it not so long ago for the props to be accessible on the entry point of each modals have a look into `src/renderer/ModalsLayer.js`

You can then retrieve show close this way:
```javascript
const TermsModal = ({ name, showClose }: { name: string, showClose: boolean }) => {

const onClickClose = useCallback(() => {
    dispatch(closeModal(name));
  }, [dispatch]);
}

return (
    <Modal
      name={name} ...
```

valpinkman: will try that

PR #2740, opened 2020-03-19
Update live-common

gre:
Update live-common with latest requirements from @LFBarreto and also fixes for ZenCash "missing inputs"

### Type

lib update

### Context

LL-2126

that the BTC likes still work fine. unfortunately we don't have a clear scenario to reproduce the bug we are fixing here, but it was tested in a unit test way.


nabil-brn:
Sync / Send
[BTC](https://blockstream.info/tx/2d558742a3ad806878a999f099952f406c05f467881a4ad32e8990d392fbbeac)
[LTC](https://live.blockcypher.com/ltc/tx/4ade65b1702f47b5f4aae8a3d52887325dbb577471180e6468e780634879b7fa/)

Platform: macOS

✓ QA-ed

---
b0f735d8 (HEAD -> live-common-12-6-3, gre/live-common-12-6-3) Update live-common

gre:
one confirmation on an OS should be enough for this bump. let's merge and also address the windows bug (that is already in develop and that is likely not a windows bug, i actually repro on mac if you dont have app.json)

PR #2748, opened 2020-03-23
Spectron - post onboarding test coverage

nabil-brn:
Improved the onboarding test suite by adding some coverage on:
- genuine check
- password form
- analytics page
- Terms of use

Spectron now runs tests on bundled app and not a packaged one.

```
yarn
yarn build
yarn spectron
```

### Type

UI Testing coverage

### Context

Spectron

Some html element IDs


tests/data/onboarding/index.js
```diff
@@ -1,25 +1,28 @@
...
+
 export default {
   appTitle: "Ledger Live",
   duskColor: "#182532",
   darkColor: "#1c1d1f",
   lightColor: "#ffffff",
-  welcomeTitle: "Welcome to Ledger Live",
-  welcomeDesc:
-    "Let’s start by choosing how Ledger Live looks to you. You can change this later at any time in your settings.",
+  welcome: {
```
juan-cortes, line 10: Why are we testing against hardcoded data like this? I don't get it

gre: I agree. we should dynamically fetch the one from the en/app.json
we can literally import the JSON and use the json path to retrieve it.

gre: also to me, we should use it "in place". there is no need to have this tests/data folder
we must "collocate" the tests as much as possible to avoid looking at N different places.

nabil-brn: Actually there are different approaches: screenshots and text assertions

Given that screenshots are still not implemented, I'm using text assertions.

About the hard coded data, it's a topic that we should discuss on:
- Different languages support (i18n?)
- Is it relevant to use production data for expected test data ?
- Easiest way to maintain

Which strategy ?
100% Visual regression testing or a mix...

.github/workflows/bundle-app.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         run: yarn version --new-version=${{ steps.version.outputs.clean }}-pr.${{ steps.pr.outputs.pr }}
       - name: build the app
         run: yarn nightly
+      - name: run ui tests
+        run: yarn spectron
```
gre, line 31: can you undo this part?

PR #2750, opened 2020-03-24
removed logo when loaded

IAmMorrow:


gre:
let's not double ✅  please =D need QA confirmation

PR #2751, opened 2020-03-24
disable firm update button while installing apps + remount after update

valpinkman:
Disable firmware update button if user is installing/uninstalling apps
Remount manager after a firmware update

### Type

Feature

### Context

LL-2321

Manager


nabil-brn:
I need my devices with old firmwares, I'll get them tomorrow.
If someone already got them and could test, would be nice.

PR #2756, opened 2020-03-25
added poc for progressive update mechanism

IAmMorrow:
this PR feature a progressive update mechanism and a remote config system.

On app boot, the following data structure is fetched from github and loaded in memory: 

> https://github.com/LedgerHQ/ledger-live-desktop/blob/config/config.json

it then calculate a deterministic random local update threshold using the userId. If the threshold is inferior than the global one provided in the remote config, the update banner will appear. It is possible to define one global update threshold per platform (darwin, win32, linux ...).

Notes:

- the whole system is optional and if the config doesn't have the required data, the update will just appear normaly
- the deterministic random is based on a ${userId}-${version} seed, this way it's not always the same users that get the update first !

How to test ?

To test, you have to modify your package.json version number to something anterior to the last release (2.0.0 for instance), then you have to package the app (yarn dist).
Run the packaged app, and an update banner will appear (or not, 1 chance out of two with the actual config.json file)

Reset your userId in the localstorage if you want to reset the odds ;)


src/helpers/user.js
```diff
@@ -29,3 +31,16 @@ export const getUserId = () => {
   }
   throw new Error("user is only to be called from renderer");
 };
+
+export const shouldUpdateYet = (version: string, remoteConfig: any) => {
+  const userId = getUserId();
+  const generateNumber = seedrandom(`${userId}-${version}`);
```
gre: `Prando` is the lib we use on live-common that is less problematic, especially on diff JS env (seedrandom does not run on RN for instance. not that we would use here but i kinda feel this code could belong to https://github.com/LedgerHQ/ledger-live-common/blob/master/src/user.js which centralize the "derives from user id" logic)

src/renderer/components/Updater/Banner.js
```diff
@@ -51,12 +53,18 @@ const CONTENT_BY_STATUS = (quitAndInstall, reDownload, progress): { [string]: Co
...
 const UpdaterTopBanner = () => {
   const context = useContext(UpdaterContext);
+  const remoteConfig = useRemoteConfig();
 
   const reDownload = useCallback(() => {
     openURL(urls.liveHome);
   }, []);
 
-  if (context) {
+  if (context && remoteConfig && context.version) {
```
gre: ⚠️  this makes the release not resilient to network down problem or problem in the fetching config.

I think by default we always should suggest the auto update when fetching config fails. the challenge might be to not having it "flash" during the loading phase, so you might want to track if it's loading/error/loaded.

IAmMorrow: I fixed it by adding a "lastUpdatedAt" param that ensure that we tried at least once to fetch the remote config, sucessfully or not.

src/helpers/user.js
IAmMorrow: I replaced seedrandom with Prando

gre:
:pikasurprised:

gre:
that's a good feedback that i didn't anticipate, i think it's because I might have intentionally said "do it on ui side" as i thought was easiest and i didn't consider the concerned raised here.

IAmMorrow:
It's okay I will implement an "on user action" download that will settle this

IAmMorrow:
I added a banner that prompt "Ledger live {version} available for update" with a "Download now" button

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ledger-live-desktop",
   "productName": "Ledger Live",
   "description": "Ledger Live - Desktop",
-  "version": "2.1.0",
+  "version": "2.0.0",
```
MortalKastor:
```suggestion
  "version": "2.1.0",
```

(unless it was on purpose to help QA?)

IAmMorrow: not on purpose, I forgot it, I'm getting really slow

src/main/updater/index.js
```diff
@@ -18,6 +18,14 @@ export default (type: string) => {
       }
       break;
 
+    case "download-update":
+      if (!updater.downloadUpdate) {
+        logger.error(`Auto-update error: quitAndInstall called before init`);
```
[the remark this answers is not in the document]
IAmMorrow: damn me

PR #2757, opened 2020-03-25
Add hidden inputs to checkbox/switch components for better QA

juan-cortes:
In order to allow automation and testing libraries to work properly, we need to maintain the standard input dom elements (hidden in our case even if it's not ideal) so as to have access to things like `.checked` which wouldn't be available on a normal div element.

### Type

QA related improvements


gre:
to be checked that it doesn't messup with the "TAB" feature of the keyboard control

juan-cortes:
> to be checked that it doesn't messup with the "TAB" feature of the keyboard control

From my testing it's not tabbable

PR #2758, opened 2020-03-25
LL-2296 Remove track your order from no device flow

juan-cortes:
Drop the track your order option from the no device flow in onboarding.

Screenshot 2020-03-25 at 15 32 08

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2296


juan-cortes:
Other translations get updated automagically as far as I'm concerned, I don't think we have to manually update them. cc @gre

PR #2761, opened 2020-03-26
LL-2298 Refactor OnboardingFooterWrapper

juan-cortes:
This should allow @nabil-brn to write cleaner tests for onboarding according to my understanding of the situation. Some screens such as the password one on onboarding do have custom ids that will need to be set manually since the buttons are different. Do take a look.

### Type

QA feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2298

Onboarding in general, all steps with a footer should now use the `OnboardingFooter` component


gre:
lint?

juan-cortes:
> lint?
Indeed, I honestly didn't check ci at all for this one. I've refactored it a bit following the way we are doing this concept of `left` and `right` in other components. Anyway, it should pass tests now.

PR #2763, opened 2020-03-26
LL-2279 ~ LL-2335

LFBarreto:
(TRX): added base voting flow not editable yet but functional.

### Type

Feature

### Context

LL-2279 LL-2335

TRX accounts that can vote


src/renderer/modals/VoteTron/Body.js
```diff
@@ -0,0 +1,207 @@
...
+type StateProps = {|
+  t: TFunction,
+  device: ?Device,
+  accounts: Account[],
+  device: ?Device,
+  closeModal: string => void,
+  openModal: string => void,
+|};
+
+type Props = {|
```
LukeSugiura: i'm not a flow expert but can you just write like `type Props = OwnProps & StateProps`?

src/renderer/modals/VoteTron/Body.js
```diff
@@ -0,0 +1,207 @@
...
+  t,
+  stepId,
+  device,
+  closeModal,
+  openModal,
+  onChangeStepId,
+  params,
+  name,
+}: Props) => {
+  const [steps] = useState(createSteps);
```
LukeSugiura: useMemo?

src/renderer/modals/VoteTron/Body.js
LFBarreto: probably i copy pasted this part from other modals :sweat_smile:

src/renderer/modals/VoteTron/Body.js
LFBarreto: also copy pasted this without thinking but you're right on this :thinking:

src/renderer/modals/VoteTron/Body.js
```diff
@@ -0,0 +1,207 @@
...
+
+  return (
+    <Stepper {...stepperProps}>
+      <SyncSkipUnderPriority priority={100} />
+      <Track onUnmount event="CloseModalVote" />
+    </Stepper>
+  );
+};
+
+const C: React$ComponentType<OwnProps> = compose(
```
LukeSugiura: Is there any reasons that this is not directly exported?

src/renderer/modals/VoteTron/index.js
```diff
@@ -0,0 +1,56 @@
...
+import type { StepId } from "./types";
+type State = {
+  stepId: StepId,
+};
+
+const INITIAL_STATE = {
+  stepId: "castVotes",
+};
+
+class VoteModal extends PureComponent<{ name: string }, State> {
```
LukeSugiura: Why is this a class component?

src/renderer/modals/VoteTron/Body.js
LFBarreto: I tried but i think its the typing on those compose components that are a bit tricky :thinking:

src/renderer/modals/VoteTron/index.js
LFBarreto: same copy pasted the same way of doing stepper modals we do elsewhere. :sweat_smile:

src/renderer/modals/VoteTron/fields/VotesField.js
```diff
@@ -0,0 +1,260 @@
...
+      </IconContainer>
+      <VoteInput value={value} disabled={disabled} onChange={e => onChange(e.target.value)} />
+    </InputBox>
+  </Row>
+);
+
+const SR_THRESHOLD = 27;
+
+type Props = {
+  t: TFunction,
```
LukeSugiura: I rather use useTranslation hook inside this component instead of passing as a prop.

LFBarreto: I don't even use it i'll remove this :+1:

PR #2764, opened 2020-03-26
[Windows] fix for Windows Defender thinking installer is a virus

MortalKastor:
A bit of a shot in the dark but :mldletstry:

### Type

Bug Fix?
Moar like false positive work-around, amirite?

### Context

https://ledgerhq.atlassian.net/browse/LL-2345

Ledger Live 2.1.0 installer introduced downloading VC++ Redistributable as a fix to crash on startup for some user, but Windows Defender thinks the installer is a virus now...

image

Installer on Windows


gre:
@MortalKastor what's the strategy to test this?

- do we have to do a "real build"? (and not use the CI one right?)
- do we know exactly who internally have reproduced the issue so we can send them a build?

MortalKastor:
> * do we have to do a "real build"? (and not use the CI one right?)

Sadly, yes, version 2.1.0 wasn't detected as infected before we used the signed build.

> * do we know exactly who internally have reproduced the issue so we can send them a build?

Any Windows 10 with default settings reproduce the issue, https://www.virustotal.com/ can be used too.

gre:
@Arnaud97234 can you try to build that PR like a normal release (without pushing it) and test the .exe installer against that website? (and compare virus detection behavior with the 2.1.0)

Arnaud97234:
### Diagnostic result virustotal.com

*Current pre-release:*
https://www.virustotal.com/gui/file/6a52e82bcc6a2ff80408d3ddd82c1155b3a86e27b87c510f23c27115ea46b2f3/detection

*With the fix:*
https://www.virustotal.com/gui/file/718787e0f73725037b1e19e298b5ec0c19dcd3a5b8ab6aaa5f8134761129b5bc/detection

Looking at the result the trojan is not detected anymore by Microsoft.

PR #2772, opened 2020-03-28
Better error screens

gre:
to test it, you can add in src/renderer/screens/Manager/index.js in line 18 a `<OHBOY />`

and when you click Manager:

Capture d’écran 2020-03-31 à 18 57 02


IAmMorrow:
This is 🚀 🔥

PR #2787, opened 2020-04-01
use team-live as user to make the push to develop

valpinkman:
Nightlies were not be triggered because the user creating the commit inside the action, was "github", and does not have rights to push to develop

We are switching to the "Team Live" user 

I had to create a secrets to give to the action to authenticate team-live

### Type

CI


valpinkman:
The `cache` step on macOS is talking a huge time, so I removed it from the step (macOS only)

PR #2788, opened 2020-04-01
LL-2207 Blacklisting tokens feature

juan-cortes:
### Description
We introduce the possibility to blacklist a token and hide all its related accounts, this was a much-requested feature since the appearance of junk tokens that users didn't want to see in their portfolios.

### Screenshots

Screenshot 2020-04-01 at 19 41 47

image

Screenshot 2020-04-01 at 19 41 54

image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2207

Users are now able to blacklist tokens from the accounts screen by right-clicking on the account (both grid and row mode) and selecting 'Hide Token'. This can also be accessed when in an asset view for a particular token in the Account Allocation block. Currently, there's no other way of accessing the feature.

After a token is hidden it will appear in the Settings screen in the Accounts section. This new block will allow us to remove a token from the blacklist, making the accounts appear again after the next sync.


src/renderer/actions/settings.js
```diff
@@ -48,6 +48,18 @@ export const setHideEmptyTokenAccounts = (hideEmptyTokenAccounts: boolean) => as
 export const setSidebarCollapsed = (sidebarCollapsed: boolean) =>
   saveSettings({ sidebarCollapsed });
 
+export const blacklistToken = (tokenId: string) => (dispatch: Dispatch<*>) =>
+  dispatch({
+    type: "BLACKLIST_TOKEN",
+    payload: tokenId,
+  });
```
gre, line 55: you can just 

```js
(tokenId: string) => ({
    type: "BLACKLIST_TOKEN",
    payload: tokenId,
  });
```

src/renderer/actions/settings.js
```diff
@@ -48,6 +48,18 @@ export const setHideEmptyTokenAccounts = (hideEmptyTokenAccounts: boolean) => as
...
+
+export const showToken = (tokenId: string) => (dispatch: Dispatch<*>) =>
+  dispatch({
+    type: "SHOW_TOKEN",
+    payload: tokenId,
+  });
```
gre, line 61: same

src/renderer/modals/BlacklistToken/index.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+    [dispatch],
+  );
+
+  return (
+    <Modal
+      name="MODAL_BLACKLIST_TOKEN"
+      centered
+      render={({ data, onClose }) => (
+        <ModalBody
```
gre, line 28: can you try to split the component in two parts? the inside of render should better be a component itself. all the hooks can then move to that component too so you don't re-render the upper level Modal and it makes less nesting.

src/renderer/modals/BlacklistToken/index.js
```diff
@@ -0,0 +1,81 @@
...
+          preventBackdropClick
+          onClose={onClose}
+          title={t("blacklistToken.title")}
+          render={() => (
+            <Box>
```
gre, line 33: could also split out with a Body and a Footer 🤔

src/renderer/reducers/settings.js
```diff
@@ -172,6 +174,20 @@ const handlers: Object = {
     ...state,
     dismissedBanners: [...state.dismissedBanners, bannerId],
   }),
+  SHOW_TOKEN: (state: SettingsState, { payload: tokenId }) => {
+    const ids = state.blacklistedTokenIds || [];
```
gre, line 178: is there a reason why `|| []` is needed?

src/renderer/reducers/settings.js
```diff
@@ -284,6 +300,7 @@ export const autoLockTimeoutSelector = (state: State) => state.settings.autoLock
 export const shareAnalyticsSelector = (state: State) => state.settings.shareAnalytics;
 export const selectedTimeRangeSelector = (state: State) => state.settings.selectedTimeRange;
 export const hasInstalledAppsSelector = (state: State) => state.settings.hasInstalledApps;
+export const blacklistedTokenIdsSelector = (state: State) => state.settings.blacklistedTokenIds;
```
gre, line 303: (because typically here it's not used as a "maybe array" i imagine.

src/renderer/screens/settings/sections/Accounts/BlacklistedTokens.js
```diff
@@ -0,0 +1,132 @@
...
+  & > * {
+    letter-spacing: 0.1em;
+  }
+`;
+
+const BlacklistedTokens = () => {
+  const { t } = useTranslation();
+  const dispatch = useDispatch();
+  const sync = useBridgeSync();
+  useEffect(() => () => sync({ type: "SYNC_ALL_ACCOUNTS", priority: 5 }), [sync]);
```
gre, line 65: can you prefix with a code comment to explain what it does?

src/renderer/screens/settings/sections/Accounts/BlacklistedTokens.js
```diff
@@ -0,0 +1,132 @@
...
+
+  const onShowToken = useCallback(
+    tokenId => {
+      dispatch(showToken(tokenId));
+    },
+    [dispatch],
+  );
+
+  const blacklistedTokenIds = useSelector(blacklistedTokenIdsSelector).sort();
```
gre, line 74: ⚠️  .sort() is mutating the initial array in place.

either you keep the things sorted in the list in the first place, or you apply the sort somewhere else. I'm not necessarily see why we want it to be sorted btw, but if we do I reco to sort it in the reducer when you **add** a new token there (sort on ids doesn't make a lot of sense tho).

juan-cortes: I figured it made sense for the list to be a-z sorted, and it does make sense for erc20 tokens since they include the name in the id, not so much for these new tron ones which are numbers 🤷‍♂ I'll just drop the sort

src/renderer/screens/settings/sections/Accounts/BlacklistedTokens.js
juan-cortes: Sure

src/renderer/modals/BlacklistToken/index.js
juan-cortes: sure

src/renderer/modals/BlacklistToken/index.js
juan-cortes: sure, makes sense

nabil-brn:
@bscholving Do you think that we should change the order of those menus?
image

bscholving:
I think it's a good idea. @juan-cortes @nabil-brn @gre

PR #2789, opened 2020-04-02
Added a simple countermeasure against clipboard hijacking attacks

IAmMorrow:
A few milliseconds after copying an address, we now verify the clipboard and match its content to the original address. If it doesn't match, we display an error for the user to see.

image

QA: You can actually trigger it if you copy the address and then copy something else within the 300 ms 😄 I was able to do it with a clever alt tab

### DIFF
--- a/src/renderer/components/ReadOnlyAddressField.js
+++ b/src/renderer/components/ReadOnlyAddressField.js
@@ -35,6 +35,14 @@ const CopyFeedback = styled(Box).attrs(() => ({
   border-left: 1px solid ${p => p.theme.colors.palette.divider};
 `;
 
+const ClipboardSuspicious = styled.div`
+  font-family: Inter;
+  font-weight: 400;
+  font-style: normal;
+  font-size: 12px;
+  color: ${p => p.theme.colors.alertRed};
+`;
+
 const CopyBtn = styled(Box).attrs(() => ({
   bg: "palette.background.paper",
   color: "palette.text.shade100",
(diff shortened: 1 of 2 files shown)


IAmMorrow:
@gre Oh apparently I am supposed to throw a PotentialClipboardAttackError error as well, it doesn't seem to exist in Ledger Errors. Shall I create it ?

gre:
nah it's ok if it's not an error

gre:
that's super cool. can you check wording @dasilvarosa ? thanks

juan-cortes:
If I understand this correctly it will only detect if somehow the clipboard was changed immediately after copying, pretty cool but if the malware is in the browser or takes longer than the specified ms which is likely then this will do nothing. I guess this is what @LFBarreto was referring to when we talked about this initially about randomizing the time it took for the check. 😕

IAmMorrow:
It make no sense to wait a long time before checking, users will copy the address and alt+tab to paste it somewhere else straight away. We need to be fast if we want to alert them and unless you trigger an alarm the user will not see it.

I believe most of those malwares are generic and aren't targeting Ledger Live especially, they hook themselves to the clipboard and listen to new entries. If one of them looks like a crypto currency address, they simply push one of their own. 99% of the targets of such malwares are the exchanges. The exchanges cannot read the clipboard as they are sandboxed (unlike us).

Also please remember that if someone want to target us specifically, they have a thousand ways of doing it more efficiently.

IAmMorrow:
Also, if we increase the checking time too much, there will be false positive as users will eventually push something in their clipboard as well.

PR #2792, opened 2020-04-02
Fixes and live-common bump + pixel push

gre:
- Performance of 🌳 Add Accounts
- Fixes Votes to not block on a button and use a very reliable way to wait for Freeze's effect
- Added error in wordings (but still waiting for wording => internal slack tron-testing-feedbacks ongoing – do not look here)

please also check on one platform that adding accounts on BTC, XTZ, ETH, ETC, XRP, works ok (all accounts still discover) the performance of Add Accounts is done by **not always loading the transaction in the account**, the tradeoff is you will not see them until a second sync happens later


gre:
continuing on https://github.com/LedgerHQ/ledger-live-desktop/pull/2794

PR #2795, opened 2020-04-03
Fallback to USD when the user selected fiat counterValue is not supported

juan-cortes:
This pr introduces the automatic fallback to using USD as a counter value when the currently selected counter value is not reported as supported. Without this, users who have selected a non-supported fiat counter value will see their total balances and graphs as broken.

### Type

Feature

### Context

Slack conversation, no Jira task as far as I know

With the application stopped, edit your `app.json` and set the counter value to something not supported (but contemplated in https://github.com/LedgerHQ/ledger-live-common/blob/1a29a3bc5f533ee29e0a47c577a1a7e068546e0b/src/data/fiat.js#L23-L73 and then launch the application. It should automagically start using USD instead of VND or whatever non-supported fiat you chose.

### DIFF
--- a/src/renderer/reducers/settings.js
+++ b/src/renderer/reducers/settings.js
@@ -14,6 +14,7 @@ import { getEnv } from "@ledgerhq/live-common/lib/env";
 import { getLanguages } from "~/config/languages";
 import type { State } from ".";
 import { osLangAndRegionSelector } from "~/renderer/reducers/application";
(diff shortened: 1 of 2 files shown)


juan-cortes:
After reworking the feature with @gre feedback I'm sorry to ask @Arnaud97234 to test again and @gre to review again. We no longer dispatch the change but rather modify the data on read, and not hardcode the USD but fallback to initialState.

PR #2799, opened 2020-04-03
more khalil tron polish

IAmMorrow:


gre:
let's merge on develop and test from there. i'll need to polish some stuff regarding the device steps as agreed from this so it will be easier. we'll heavily test on develop.

PR #2802, opened 2020-04-03
Fixes device step to match exactly what happens on device

gre:
- this fixes the device step to match exactly what is shown in device.
- as agreed, the exception is that the SEND should not display any address because you are supposed to verify against where you send to, and showing the "From" address would be confusing here.
- The text are hardcoded in English because we must reflect the device wording. in future, we will try to generalize this back in live-common as suggested in https://github.com/LedgerHQ/ledger-live-common/issues/596
- **Also includes fixes of https://github.com/LedgerHQ/ledger-live-common/releases/tag/v12.13.0**

### Type

ui polish / security

### Context

🌳 

send

Capture d’écran 2020-04-03 à 21 28 47

send token

Capture d’écran 2020-04-03 à 21 35 20

freeze

Capture d’écran 2020-04-03 à 21 28 07

unfreeze

Capture d’écran 2020-04-03 à 21 22 53

votes

Capture d’écran 2020-04-03 à 21 33 36

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@ledgerhq/hw-transport-http": "^5.12.0",
     "@ledgerhq/hw-transport-node-hid-singleton": "^5.12.0",
     "@ledgerhq/ledger-core": "^6.2.1",
-    "@ledgerhq/live-common": "^12.12.2",
+    "@ledgerhq/live-common": "^12.13.0",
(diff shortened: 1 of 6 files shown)


FabriceDautriat:
@thomas-rj : can you pleae help review the screens above and compare with what is actually displayed  on the device ?

PR #2806, opened 2020-04-06
LL-2373 Export the black list of token ids on LiveQr

juan-cortes:
Add the setting when exporting accounts/settings

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2373

Once both PRs for blacklisting tokens are available on desktop and mobile, blacklist some on desktop, export the settings and see if they're blacklisted on mobile.


juan-cortes:
I forgot we had to do https://github.com/LedgerHQ/ledger-live-common/pull/605 so this pr will not be working until it's available on mobile.

PR #2838, opened 2020-04-09
removed downshift; replaced all use cases with tippy.js; upgraded tip…

IAmMorrow:
…py and react-tippy

QA:

test all the old downshift menus, basically all areas where you could click and expand a menu / selector (in the breadcrumb at the top, filtering, sorting ...)
test that the tooltips are still working properly :)


src/renderer/components/Breadcrumb/AccountCrumb.js
```diff
@@ -127,28 +121,36 @@ const AccountCrumb = () => {
     );
   }
 
+  console.log(items, processedItems);
```
valpinkman, line 124: oops ?

IAmMorrow: :sad:

PR #2860, opened 2020-04-16
Fixes windows virus false-positive issue due to electron-builder

gre:
Revert to older electron-builder because of https://github.com/electron-userland/electron-builder/issues/4793

### Type

bug fix

### Context

- https://github.com/electron-userland/electron-builder/issues/4793
- we might want to reborn in future the work of https://github.com/LedgerHQ/ledger-live-desktop/pull/2815

- this PR affects electron so we should check the overall app is fine.
- Check that builds on Linux / Mac / Windows are fine => which is essentially covered by the CI already!


gre:
it works

asohbat:
2.2.3 win
Untitled

asohbat:
> it works

it does not . 2.2.3 windows
Untitled

tomasz-s:
Just installed 2.2.3. Same problem. 
And, its new. I've never had this problem before.

PR #2862, opened 2020-04-16
Spectron testing with mocked device events

juan-cortes:
# Event mocking for dev & QA

In order to be able to complete all OK/KO cases of all flows for testing and development, we expose a way to mock events that require interaction with a real device. When running with the env variable `MOCK` you'll have access to `mockDeviceEvent` which receives one or more events that will be emitted to the subscribed observers triggering the associated actions.

### Dev
Simply open the console, `mockDeviceEvent` is available as a global window variable. Pass the snippet below or any event of your liking to that method.

### QA
As it's done in the `onboarding.spec.js` file, we need to get access to that method via a call to `const mockeDeviceEvent = getMockDeviceEvent(app)` afterward the functionality of the method is the same as for dev with the exception of having to `await` the result to not break the tests.

## What is this useful for?
It allows us to interact with all parts of the application without the need for a real device or real accounts. 

## Snippets per-flow
These are a series of snippets that allow you to reach the OK and KO ends of a device interaction for each of the flows. For the KOs, we can essentially use the name of any error and it will throw it, depending on the flow ones make more sense than others.

### Receive
- OK `{ type: "opened" }`
- KO `{ type: "error", error: { name: "WrongDeviceForAccount" } }`
- KO `{ type: "error", error: { name: "DisconnectedDevice" } }`

### Send
- OK `{ type: "opened" }`
- KO `{ type: "error", error: { name: "WrongDeviceForAccount" } }`
- KO `{ type: "error", error: { name: "UserRefusedOnDevice" } }`

### Add accounts
- OK `{ type: "opened" }`
- KO `{ type: "error", error: { name: "DisconnectedDevice" } }`

### Genuine Check / Manager
- OK (note it's not an array) We could make other profiles for devices available and fix the sizes of the apps returned by `mockListAppsResult` to test things like the storage breakdown in the manager. **Note** for devs in the console you will be able to run this snippet only because we've exposed the dependencies as globals when in mock.

  Refer to the signature of `mockListAppsResult` to see the doors it opens. But a tldr is the first parameter are the apps available in the catalog, the second is the list of installed apps.
    ```
    { type: "listingApps", deviceInfo: deviceInfo155 },  
    {  
      type: "result",  
      result: mockListAppsResult("Bitcoin", "Bitcoin", deviceInfo155),  
    }
   ```

- KO `{ type: "error", error: { name: "GenuineCheckFailed" } }`
- KO `{ type: "error", error: { name: "WrongDeviceForAccount" } }`

### Type

Feature

We also now have a little badge indicating we are in a mock mode in case we forget and suddenly add accounts and think we are rich. Not that it happened to me.
image

### DIFF
--- a/src/renderer/modals/Receive/steps/StepReceiveFunds.js
+++ b/src/renderer/modals/Receive/steps/StepReceiveFunds.js
@@ -26,6 +26,7 @@ import type { StepProps } from "../Body";
 import Modal from "~/renderer/components/Modal";
 import ModalBody from "~/renderer/components/Modal/ModalBody";
 import QRCode from "~/renderer/components/QRCode";
+import { getEnv } from "@ledgerhq/live-common/lib/env";
 
 const Separator = styled.div`
   border-top: 1px solid #99999933;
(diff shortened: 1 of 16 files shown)


src/renderer/modals/Receive/steps/StepReceiveFunds.js
```diff
@@ -155,23 +156,30 @@ const StepReceiveFunds = ({
...
-      }).toPromise();
-      if (address !== mainAccount.freshAddress) {
-        throw new WrongDeviceForAccount(`WrongDeviceForAccount ${mainAccount.name}`, {
-          accountName: mainAccount.name,
-        });
+      if (getEnv("MOCK")) {
+        setTimeout(() => {
+          onChangeAddressVerified(true);
+          transitionTo("receive");
+        }, 3000);
```
gre, line 163: (for later) we should have the opportunity to mock that part. it bypass the potential test of having WrongDeviceForAccount and also a delay during the device verification

maybe a future system to mock commands? anyway, not now

juan-cortes: we can still emit `{type:"error",error:{name:"WrongDeviceForAccount"}}` and trigger it 🤷

PR #2886, opened 2020-04-24
Bullrun mock run and mock helper

juan-cortes:
### Spectron side
Introduces a preliminary end-to-end run of the desktop application with Spectron going over all the flows without really asserting anything other than the fact that we are able to go over them. It currently covers the following flows: 
- Onboarding
- Access manager
- Firmware update
- Add accounts
- Send funds
- Receive
- Delegate
- Settings
- Partners
- Account migration

Some other flows like the currency specific voting flow ~or the account migration flow~ are not covered currently ~but I'll look into them probably during the weekend~. It involves so many changed files due to the need to provide missing identifiers that can be used to select an element from the Spectron side, and a bunch of tweaks.

### Manual side
When accessing the app in MOCK mode we now have a mock event-specific UI on the bottom left corner of the window allowing us to see a list of useful events that we might want to emit, a queue of the events that will be emitted once an observer subscribes to our mock event emitter, and a history of the emitted events via this mean.

Every section of this widget is collapsible, we can remove elements from the queue, and the quick events and history list are both clickable resulting in the target event being pushed back into the queue.
Since the events are too long, the full data is available on hove, and also on the below mentioned `window.mock.events.history`.
image

Additionally, we now have access to a `window.mock` object that contains all the data and methods we might need in case we want to do this from the console instead.

For Spectron runs, in order to avoid the widget colliding with the touch events, we can pass the env `DISABLE_MOCK_POINTER_EVENTS` flag which would essentially make this widget pass through the touch events, allowing us to see the queue of events while running a Spectron test suite.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2406

I'll write a test plan for the tests, later.


gre:
🚀

tests/specs/bullrun.spec.js
```diff
@@ -0,0 +1,237 @@
...
+    passwordPage = new PasswordPage(app);
+    analyticsPage = new AnalyticsPage(app);
+    portfolioPage = new PortfolioPage(app);
+    mockDeviceEvent = getMockDeviceEvent(app);
+
+    return app.start();
+  });
+
+  afterAll(() => {
+    return restoreUserData();
```
gre, line 40: a better duality would be to put the backupUserData() in beforeAll no?

On the pattern involved here itself. I think it would be way better to spawn Spectron with a specific user data for the run of this test.

to my mind, each single test should run in a standalone user data that gets created somewhere like in a /tmp/folder , that way it's not touching existing user data at all.

juan-cortes: The rimraf was what was present before I just added an imperfect backup, I'll look into a better way of doing this.

PR #2890, opened 2020-04-27
docker action first test

IAmMorrow:
This PR add a spectron test using a docker container on the CI github action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,3 +35,6 @@ jobs:
           yarn prepublish
       - name: test
         run: yarn ci
+      - name: Run spectron
+        id: spectron
+        uses: IAmMorrow/run-headless@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -35,3 +35,6 @@ jobs:
           yarn prepublish
       - name: test
         run: yarn ci
+      - name: Run spectron
+        id: spectron
+        uses: IAmMorrow/run-headless@v4
```
gre, line 40: (can we later move it to LedgerHQ? would need maybe a more obvious name, like `ledger-live-github-action-headless` or something)

valpinkman: If it works, you can move it 
https://github.com/LedgerHQ/actions

IAmMorrow: yep and we'll take the command to run as input

gre:
ok let's merge this for now. :goodenough:
and we can iterate in another PR for the small changes of actions

PR #2896, opened 2020-05-04
Fix broken develop without MOCK flag

juan-cortes:
This slipped through and basically breaks the app if not running on mock

### Type

Bug fix

### DIFF
--- a/src/renderer/components/DebugMock.js
+++ b/src/renderer/components/DebugMock.js
@@ -152,7 +152,8 @@ if (getEnv("MOCK")) {
   observerAwareEventLoop();
 }
 
-export const mockedEventEmitter = window.mock.events.emitter;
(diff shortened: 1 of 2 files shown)


src/renderer/components/DebugMock.js
```diff
@@ -152,7 +152,8 @@ if (getEnv("MOCK")) {
   observerAwareEventLoop();
 }
 
-export const mockedEventEmitter = window.mock.events.emitter;
+// $FlowFixMe
```
gre, line 155: Why would we have to flow fixme here?

juan-cortes: because if I don't, then every time I'm using the mockedEventEmitter I would have to check it exists, even though we know it does.

PR #2897, opened 2020-05-05
Device transaction config

juan-cortes:
~To be tested alongside https://github.com/LedgerHQ/ledger-live-common/pull/636 from live-common.~ Already bumped

If this works, on the device verification step of a device-involving operation we should see the fields pertinent to the operation. For instance, on a send for stellar with a memo, we should see the memo on the validating fields on the device.  If this works, we will have the ability to customize the rendering per field per currency, and determining what gets displayed on the app at the live-common level.

### Type

Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-2423

Once https://github.com/LedgerHQ/ledger-live-common/pull/636 is merged we need to go over all the send/delegate/vote/freeze/unfreeze/etc flows to make sure the fields displayed are correct. Sorry.


gre:
nice

gre:
@juan-cortes can you update the PR? live-comomn version for this has landed in develop.

PR #2900, opened 2020-05-07
LL-2394 (Tron): unfreeze success modal wording updated

LFBarreto:
After unfreezing some trx the success modal confirmation text should display the correct resource unfrozen for. (energy or bandwidth)

### Type

Wording

### Context

LL-2394

Unfreeze some trx chose energy -> success modal should refer to energy unfrozen same applies in the case of bandwidth

### DIFF
--- a/src/renderer/modals/Unfreeze/steps/StepConfirmation.js
+++ b/src/renderer/modals/Unfreeze/steps/StepConfirmation.js
@@ -29,6 +29,7 @@ const Container: ThemedComponent<{ shouldSpace?: boolean }> = styled(Box).attrs(
 function StepConfirmation({
   account,
   t,
+  transaction,
   optimisticOperation,
   error,
   theme,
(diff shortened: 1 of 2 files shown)


src/renderer/modals/Unfreeze/steps/StepConfirmation.js
```diff
@@ -42,7 +43,11 @@ function StepConfirmation({
         <SyncOneAccountOnMount priority={10} accountId={optimisticOperation.accountId} />
         <SuccessDisplay
           title={<Trans i18nKey="unfreeze.steps.confirmation.success.title" />}
-          description={multiline(t("unfreeze.steps.confirmation.success.text"))}
+          description={multiline(
+            t("unfreeze.steps.confirmation.success.text", {
+              resource: transaction && transaction.resource && transaction.resource.toLowerCase(),
```
srowhani, line 48: You can instead use optional chaining here

PR #2609, opened 2020-02-24
Refactor some components in TypeScript

JunichiSugiura:

PR #2912, opened 2020-05-13
Refactor some components in TypeScript

JunichiSugiura:
I accidentally deleted my remote branch for #2609 so this is the replacement.


src/renderer/components/AccountDistribution/Row.tsx
```diff
@@ -1,105 +1,59 @@
...
-  countervalue: BigNumber, // countervalue of the amount that was calculated based of the rate provided
-};
 
-type Props = {
-  item: AccountDistributionItem,
-  isVisible: boolean,
-};
+export interface AccountDistributionItem {
+  // [TODO] account: Account | TokenAccount;
+  account: any;
```
gre, line 38: i'm not sure we should merge this (lower our type coverage) **until** we migrate live-common tho?

JunichiSugiura: We should migrate from ledgerjs -> ledger-live-common -> ledger-live-desktop/mobile I think.

gre:
this establishes the technically possibility of this. we will focus on Ledgerjs / live-common first during the next months.

PR #2925, opened 2020-05-18
LL-2511 Cosmos Undelegation Flow

JunichiSugiura:

PR #2926, opened 2020-05-19
LL-2512 LL-2511

LFBarreto:
added :atom: redelegation and undelegation flows

includes #2925 

### Type

Feature

### Context

LL-2512 LL-2511

:atom: accounts with delegated assets.
this uses mocks for now more details on the main issue.


src/renderer/families/cosmos/UndelegationFlowModal/steps/Amount.js
```diff
@@ -28,16 +28,19 @@ export default function StepAmount({
       onUpdateTransaction(tx =>
         bridge.updateTransaction(tx, {
           ...tx,
-          validators: [{ ...tx.validators[0], ...validatorFields }],
+          validators:
+            tx.validators && tx.validators.length > 0
+              ? [{ ...tx.validators[0], ...validatorFields }]
+              : [{ ...validatorFields }],
```
JunichiSugiura, line 34: this can be just `: [validatorFields],`

src/renderer/families/cosmos/UndelegationFlowModal/steps/Confirmation.js
```diff
@@ -24,6 +24,8 @@ export default function StepConfirmation({
 }: StepProps) {
   const { t } = useTranslation();
 
+  console.log(transaction, account);
```
JunichiSugiura, line 27: remove

src/renderer/families/cosmos/UndelegationFlowModal/fields/Amount.js
```diff
@@ -28,7 +28,10 @@ export default function AmountField({
   const { t } = useTranslation();
   const unit = getAccountUnit(account);
 
-  const initialAmount = useMemo(() => (validator ? validator.amount : BigNumber(0)), [validator]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const initialAmount = useMemo(() => (validator ? validator.amount : BigNumber(0)), [
```
gre, line 32: i'm not sure why we need a memo for this (if there is a business logic related to it, it worth documenting, but also i feel we use memo wrong if it's the case as react don't guarantee you memo won't get recalculated 🤔 ). there is a overuse of memo in our app.

LFBarreto: i updated the logic here maybe its more clear that way we need to update the "initialAmount" variable each time the source validator is changed a sideEffect is indeed a better way to do things

PR #2928, opened 2020-05-19
wip regression test with jest image snapshots

valpinkman:
### Type

Feat

### Context

LL-2438

Tests

This PR setup jest-image-snapshot as a way of doing regression testing on the screen of Ledger Live. 

What this PR is NOT: a full suite of test using jest-snapshot-testing.

I believe it's pretty easy to keep on improving the test using the bullrun and this PR. QA team here you go !!!

### DIFF
--- a/tests/specs/bullrun.spec.js
+++ b/tests/specs/bullrun.spec.js
@@ -4,23 +4,28 @@ import ModalPage from "../po/modal.page";
 import GenuinePage from "../po/genuine.page";
 import PasswordPage from "../po/password.page";
 import AnalyticsPage from "../po/analytics.page";
-import PortfolioPage from "../po/portfolio.page";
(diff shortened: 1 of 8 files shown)


tests/specs/bullrun.spec.js
```diff
@@ -30,9 +35,8 @@ describe("Bullrun", () => {
     genuinePage = new GenuinePage(app);
     passwordPage = new PasswordPage(app);
     analyticsPage = new AnalyticsPage(app);
-    portfolioPage = new PortfolioPage(app);
+    // portfolioPage = new PortfolioPage(app);
```
gre, line 38: why is it commented?

valpinkman: because it was not used anywhere

PR #2929, opened 2020-05-19
live-common 12.23.0 + libcore

gre:
- libcore bump & LL-2156: disable SyncToken usage => check all coins are fine, especially tezos.
- LL-2484: disrupt ClubCoin => no longer should appear in list
- LL-2546: analytics to count number of account in currency => nothing to test. we'll postpone to testing with Ben.
- LL-2522: bch_on_bitcoin_segwit invalid path recovering derivation => if you send BCH to a BTC address that is from FIRST segwit bitcoin account address it will start appearing as a unknown. **Low Effort on testing this, I made it work on my side**. At least check BCH derivation is fine.


gre:
HODL. there is a regression in libcore and tezos. waiting for a fix.

gre:
unHODLed

gre:
ok as this is the only issue, and it's not just about this coin but all terminated coins, we'll merge this bump and we'll fix it in a future pr.

PR #2933, opened 2020-05-27
Exchange: buy crypto

IAmMorrow:
It just works ™


src/main/window-lifecycle.js
```diff
@@ -94,5 +94,10 @@ export async function createMainWindow({ dimensions, positions }: any, settings:
     mainWindow = null;
   });
 
+  mainWindow.webContents.on('new-window', (event, url) => {
+    event.preventDefault();
+    shell.openExternal(url)
```
gre, line 99: ⚠️  make sure to check the url starts with `https://` or `http://` but nothing else.

IAmMorrow: roger that

IAmMorrow: all good :)

PR #2937, opened 2020-05-28
add screenshots testing to firmware update flow

valpinkman:
Fix flow for Screenshot Testing, and add screenshots for firmware update flow


valpinkman:
> any idea why yarn.lock got committed but nothing changes in package.json ?

Probably a `yarn install` after I did a `yalc add...`

PR #2939, opened 2020-05-29
Fix DelegateAmountCell rendering when amount==0

gagbo:
React prefers a `return null` over a bare `return` and that breaks the UI when reaching that branch

### Type
Bug Fix

### Context
Internal Slack channel


gagbo:
No idea why CI would fail for such a change :(

PR #2958, opened 2020-06-03
Test e2e - Enable dev mode

Arnaud97234:
- [x] Check Portfolio empty state
- [x] Enable developper mode from Experimental settings
- [ ] Verify that testnet app are available in the manager (TODO)
- [x] Verify that testnet currency is available from add account
- [x] Add tBTC accounts
- [x] Check Portfolio with accounts
- [x] Check Accounts page

**Mocks improvement requested**

- Check testnet app available in the Manager: Currently when using mocks, the developer mode has no effect on the apps available in the Manager (tBTC always available, even when Dev mode is OFF).
- Release notes and Terms of use modals are displayed over each other.

### Type

UI end to end test

### Context

https://ledgerhq.atlassian.net/browse/LL-2481

Empty state, Portfolio with accounts, Settings, Manager, Add Account, Accounts

`jest tests/specs/dev-mode.spec.js`


gre:
This PR was forgotten. @Arnaud97234 & @machard => can you both work together to try to extract the essence of what this was doing so we can apply it to the existing & more recent UI testing?

PR #2931, opened 2020-05-20
LL-2548 Cosmos

JunichiSugiura:

PR #2964, opened 2020-06-08
LL-2548 Cosmos (#2931)

LFBarreto:
:atom:

### Type

Feature


gre:
@valpinkman @LFBarreto can you both work together tomorrow to see if this Github Actions feedback make sense and how to fix it if it's a "bad case"?

unfortunately there is not enough coverage of screenshot yet, but we'll need to try to see what happens when we play with accounts and screenshot it.

src/renderer/components/Delegation/ValidatorRow.js
```diff
@@ -221,6 +227,37 @@ const ValidatorRow = ({
 
   const itemExists = typeof value === "number";
 
+  const InputComponent =
```
gre, line 230: this is not a component. it's an element. should be lower cased. can be called `const input =`

PR #2985, opened 2020-06-11
Cosmos feedback fixes

LFBarreto:
Polishes mainly on styling.

### Type

UI Polish

### Context

Gre Google doc

Atom accounts

### DIFF
--- a/src/renderer/families/cosmos/DelegationFlowModal/fields/ValidatorsField.js
+++ b/src/renderer/families/cosmos/DelegationFlowModal/fields/ValidatorsField.js
@@ -23,6 +23,7 @@ import type {
   CosmosDelegationInfo,
   CosmosMappedValidator,
 } from "@ledgerhq/live-common/lib/families/cosmos/types";
+import { formatCurrencyUnit } from "@ledgerhq/live-common/lib/currencies";
 
 import { openURL } from "~/renderer/linking";
 import Box from "~/renderer/components/Box";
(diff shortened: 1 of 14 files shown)


src/renderer/families/cosmos/DelegationFlowModal/fields/ValidatorsField.js
```diff
@@ -265,14 +263,18 @@ const ValidatorField = ({
     ],
   );
 
+  const formatMax = +formatCurrencyUnit(unit, max);
```
gre, line 266: there is an accidental `+` here?

LFBarreto: Not really in that case :sweat_smile: i need a formatted number to check if the available max is lower than 0.
The component is shared with tron so it makes for weird stuff like this :shrug:

gre: ok, we really shouldn't "parse the formatter" , i don't understand why we don't do somethign else like `max.times(10 ** unit.magnitude).toNumber()` ?


i don't like the `+something` pattern at all because it looks like a typo, i prefer we explicitly use parseInt if it makes sense, here we should try to not format and parse again

gre: I forgot to ask this, but i'm curious why we actually use floating value. we should try to stay in the "satoshis" domain of units. to me we only need to go back to "decimals" domain at the parser and formatter time, which are two extremity. forcing us to do so will be a good practice in a future consideration of a switch to `bigint` – which will not have "floating points".

PR #2987, opened 2020-06-11
Cosmos/Tron changes to formatCurrencyUnit

LFBarreto:
(cosmos/tron): change usage for formatCurrencyUnit to CurrencyUnitValue that integrates locales

### Type

Bug Fix

Try and check if the formatting for most of the amounts on cosmos and tron are correctly localized


LFBarreto:
@gre i'll create a function for it then that only returns the string but yeah it works it still is JS :sweat_smile:

gre:
just beware that we can't do this if you have react hooks in it. especially not with conditions like this

```
    selectedValidator &&
    CurrencyUnitValue({
```

react will badly break as soon as selectedValidator goes `null` and back to defined.

PR #3005, opened 2020-06-16
Select currency order tokens after currencies

LFBarreto:
(SelectCurrency): reorder tokens after currencies on the select lists

tokens are appended to the cryptocurrencies list after marketcap sorting

localhost_8080_webpack_index html_theme=dusk

point #81 of cosmos ICC

### Type

UI Polish

on add accounts or any currency selector the ordering should reflect those changes

### DIFF
--- a/src/renderer/components/SelectCurrency.js
+++ b/src/renderer/components/SelectCurrency.js
@@ -14,6 +14,7 @@ import CryptoCurrencyIcon from "~/renderer/components/CryptoCurrencyIcon";
 type Props<C: Currency> = {
   onChange: (?C) => void,
   currencies: C[],
+  tokens: C[],
   value?: C,
   placeholder?: string,
   autoFocus?: boolean,
@@ -29,6 +30,7 @@ const SelectCurrency = <C: Currency>({
   value,
   placeholder,
   currencies,
+  tokens,
   autoFocus,
   minWidth,
   width,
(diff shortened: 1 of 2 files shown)


src/renderer/components/SelectCurrency.js
```diff
@@ -14,6 +14,7 @@ import CryptoCurrencyIcon from "~/renderer/components/CryptoCurrencyIcon";
 type Props<C: Currency> = {
   onChange: (?C) => void,
   currencies: C[],
+  tokens: C[],
```
gre, line 17: I don't really understand the problem at stake here. I think the feedback you received is based on a OLD testing that involved not yet having the countervalues and the proper marketcap ording for cosmos.

mmh to me there need to be a better way to solve this or maybe a better typing if we want to move to this. the select starts to take two list now, it feels hard to use.

to me the fixes need more internal discussion and probably it's something to solve in `useCurrenciesByMarketcap` instead (it could have a second parameter to precise you want to bail out the tokens) because we will need that not only on desktop, and maybe not only for this specific add accounts?

LFBarreto: The issue is that some Ethereum tokens are mixed in between the other coins and make it a bit confusing and hard on the selection of regular coins (which should be prioritised)
We can indeed update the logic in live-common i would need to look a bit more into it.

gre: ok, to me we should postpone this task to a sprint task, it's out of scope of the new coin we integrate and of ICC feedback to me.

PR #3016, opened 2020-06-18
LL-2589 (Onboarding): change nanoX step wording

LFBarreto:
Updated nanoX onboarding recovery wording

The final recovery wording will be applied with live-common changes

### Type

Wording

### Context

LL-2589

Test the onboarding flow the step wordings for recovery should be the same between nanoS and nanoX


gre:
conflicts.

PR #3021, opened 2020-06-22
POC Notification system

juan-cortes:
Weekend project that aims to kickstart the proposal of having ephemeral notifications on the app to transmit some message to our users. The current system is pretty easy to plug into, it relies on a shared context and we can emit/dismiss notifications from anywhere that uses it. What I mean by ephemeral is that there's no memory of the notifications after we launch the app for now, perhaps we could keep track of dismissed ids to do single use notifications too.

image
image

### Type

Feature

### How to emit/consume/dismiss notifications
```
const { items, dismiss, add } = useContext(NotificationsContext);
```
Everything is exposed from the notification context, we can then list notifications, emit them from pertinent events such as broadcasting, or price changes, etc.

### TODO/Wishlist
- Types of notifications (error/warning/success)
- Icon support
- Different positioning on screen/size
- Return an id from the add to autogenerate instead of having to provide the id

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -13,6 +13,7 @@ import "tippy.js/dist/svg-arrow.css";
 
 import type { State } from "~/renderer/reducers";
 import StyleProvider from "~/renderer/styles/StyleProvider";
+import NotificationsProvider from "~/renderer/components/Notifications/NotificationsProvider";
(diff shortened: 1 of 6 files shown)


juan-cortes:
Closing since this is now 5 months old

PR #3030, opened 2020-06-24
LL-2756 Promo Carousel feature

juan-cortes:
image
image
image
image

The carousel will iterate through the provided slides except from when the user's cursor is on it, hover should trigger a pause of the play but still allow the parallax effect on the banner itself. Each banner is independent and should have its own logic as to what to do with the CTA and what content to display, any current banner can serve as a template for further slides.

There is currently no means to not show a slide depending on the user's data or actions but this should be the natural next step on this feature since there will be slides that should not be presented to all users. Think swap for not supported ip addresses for instance.

### Type

Feature

### Context
https://ledgerhq.atlassian.net/browse/LL-2756

The carousel is in the portfolio, make sure it doesn't go crazy, or affect performance. 
Navigation should work, ctas should work, we are pending a copy check of the wording.
Also check the empty states for the dashboard (no accounts/no apps) do **not** show the carousel.


juan-cortes:
> Cool stuff. Some lint issues

Cleaneup up, basically forgot to delete the import of the dead delegate banner.

juan-cortes:
Chris requested the the carousel is not shown if we are in the placeholder for the portfolio, adding the hodl label until that's implemented.

PR #3031, opened 2020-06-25
LL-2601 Add a warning when we have txs pending confirmation

juan-cortes:
imageimage

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2601

Pay special attention to this breaking the value ticker (the animated number component) when we change currencies, countervalues, sync etc. I made the total balance inline but maybe that breaks something else, hopefully not. 🤞

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -33,14 +33,7 @@
   dependencies:
     tunnel "0.0.6"
 
-"@babel/code-frame@^7.0.0":
-  version "7.10.3"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -33,7 +33,14 @@
   dependencies:
     tunnel "0.0.6"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
```
gre, line 36: Mmh can you avoid commit this yarn.lock? There is no reason it updates.

PR #3032, opened 2020-06-25
fixed account settings in row alignment problem

IAmMorrow:
image

### DIFF
--- a/src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
+++ b/src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
@@ -262,12 +262,12 @@ export default ConnectedAccountSettingRenderBody;
 
 export const Container: ThemedComponent<{}> = styled(Box).attrs(() => ({
   flow: 2,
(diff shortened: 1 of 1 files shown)


src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
```diff
@@ -262,12 +262,12 @@ export default ConnectedAccountSettingRenderBody;
 
 export const Container: ThemedComponent<{}> = styled(Box).attrs(() => ({
   flow: 2,
-  justify: "space-between",
```
gre, line 265: i think it was renamed `justifyContent` , but your solution works too

IAmMorrow: that design system opportunity will be perfect to decide if we actually want to get rid of the <Box /> paradigm

gre: hell yeah

PR #3038, opened 2020-06-29
remove uninstall disclaimer

gre:
remove firmware update banner that says apps will be uninstall since they are now reinstalled.


gre:
included in sibling pr

PR #3045, opened 2020-07-01
Remember last seen device to have a dynamic Blue Dot on the Manager tab if a firmware or apps update available

valpinkman:
Feature: Remember last seen device to have a dynamic Blue Dot on the Manager tab if a firmware or apps update available.

Capture d’écran 2020-07-03 à 10 41 33

**How does it work?**

- each time we list apps (genuine check / opening manager) or install apps/firmware => we save the state of the "last seen device" including: the model (nano s / nano x / blue), the firmware version, the list of apps installed.
- At the boot of Ledger Live, we fetch the available firmware and apps for the last seen device to present or not a  🔵  will appear on the **🔧 Manager** tab.

**Scenarios**

- For a new user, going through the onboarding genuine check will saves its state, so a  🔵  will appear on the **🔧 Manager** tab **only** if an app is outdated (or firmware outdated)
- For an existing user, a 🔵  will appear on the **🔧 Manager** tab. _this happens because we don't yet know what is your device_. Entering Manager would make the  🔵  disappear if there are no available update.
- Changing the provider to make a firmware update available will dynamically make the  🔵  to appear on the **🔧 Manager** tab. turning it off will make it disappear (unless an app is outdated)
- When you enter Manager, and you apply an update,  🔵  will disappear automatically.

**NB**

- this does not need a device to be plugged to work.
- tradeoff: plugging another device will **NOT** reflect the 🔵 state reality until you reopen Manager.


src/renderer/actions/settings.js
```diff
@@ -88,3 +88,9 @@ export const dismissBanner = (bannerKey: string) => ({
 
 export const setPreferredDeviceModel = (preferredDeviceModel: DeviceModelId) =>
   saveSettings({ preferredDeviceModel });
+
+// TODO VAL: should I use saveSettings here ????
```
gre, line 92: i prefer to have explicit action because more reducer could listen to them so all good here.

src/renderer/components/DeviceAction/index.js
```diff
@@ -219,6 +221,7 @@ const DeviceAction = <R, H, P>({
 const mapStateToProps = createStructuredSelector({
   reduxDevice: getCurrentDevice,
   preferredDeviceModel: preferredDeviceModelSelector,
+  lastSeenDevice: lastSeenDeviceSelector,
```
gre, line 224: i don't understand why deviceAction need them. to me we potentially just need to call a function to "refresh". and even that, i don't think we should do it in DeviceAction but just a side effect as we do for onboarding & manager.

src/renderer/reducers/settings.js
```diff
@@ -68,9 +75,12 @@ export type SettingsState = {
   loaded: boolean, // is the settings loaded from db (it not we don't save them)
   hasCompletedOnboarding: boolean,
   counterValue: string,
+  // TODO: DEPRECATE THESE
   preferredDeviceModel: DeviceModelId,
```
gre, line 79: to me this one is the only one that is not deprecated. I would keep this one and only this one as discussed yesterday.
preferredDeviceModel would "move" independently of the other DeviceModelInfo.

preferredDeviceModel can be refreshed very often, because we quickly can know what is the model plugged. on the other end, the apps + deviceInfo is a costy info to get so we only refresh it in a few cases (entering manager essentially)

src/renderer/reducers/settings.js
```diff
@@ -338,6 +358,8 @@ export const dismissedBannerSelectorLoaded = (bannerKey: string) => (state: Stat
 export const hideEmptyTokenAccountsSelector = (state: State) =>
   state.settings.hideEmptyTokenAccounts;
 
+export const lastSeenDeviceSelector = (state: State) => state.settings.lastSeenDevice;
```
gre, line 361: to me there is no reason we need this "raw" selector but i'm not seeing the selector that tells if blue is on or off 🤔

src/renderer/components/DeviceAction/index.js
valpinkman: gotcha

src/renderer/reducers/settings.js
valpinkman: ahah still a wip

src/renderer/hooks/useShouldDisplayBlueDot.js
```diff
@@ -0,0 +1,80 @@
...
+    ])
+      .then(([fw, outdatedApp]) => {
+        if (cancelled) return;
+
+        if (fw || outdatedApp) {
+          setDisplay(true);
+          return;
+        }
+
+        setDisplay(false);
```
gre, line 67: `setDisplay(Boolean(fw || outdatedApp))`

src/renderer/hooks/useShouldDisplayBlueDot.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    if (!dmi) {
+      setDisplay(true);
+      return cancel;
+    }
+
+    const { deviceInfo, modelId, apps } = dmi;
+
+    if (!deviceInfo || !modelId) {
```
gre, line 50: that should not be possible. looking at your type, these are never falsy.

src/renderer/hooks/useShouldDisplayBlueDot.js
```diff
@@ -0,0 +1,80 @@
...
+  const outdated = apps.map(app => {
+    const currApp = compatibleAppVersionsList.find(e => e.name === app.name);
+    return currApp && semver.gt(currApp.version, app.version);
+  });
+
+  return outdated.some(Boolean);
+}
+
+function useShouldDisplayBlueDot(dmi: ?DeviceModelInfo): boolean {
+  const [display, setDisplay] = useState(false);
```
gre, line 35: initial value could be `!dmi`

src/renderer/screens/manager/AppsList/index.js
```diff
@@ -96,10 +107,17 @@ const AppsList = ({ firmware, deviceInfo, result, exec, t, render, appsToRestore
...
-  const hasOutdatedAppsOrFirmware = !!firmware || state.installed.some(a => !a.updated);
+  // Save last seen device
   useEffect(() => {
-    reduxDispatch(setHasOutdatedAppsOrFirmware(hasOutdatedAppsOrFirmware));
-  }, [reduxDispatch, hasOutdatedAppsOrFirmware]);
+    const dmi = {
+      modelId: device.modelId,
+      deviceInfo,
+      apps: state.installed.length
+        ? state.installed.map(({ name, version }) => ({ name, version }))
```
gre, line 116: you don't need the ternary here

src/renderer/screens/onboarding/steps/GenuineCheck/index.js
```diff
@@ -82,6 +82,20 @@ const GenuineCheck = (props: StepProps) => {
...
         dispatch(setHasInstalledApps(false));
       }
+
+      if (result) {
+        const { deviceModelId: modelId, deviceInfo, installed } = result;
+        dispatch(
+          setLastSeenDeviceInfo({
+            modelId,
+            deviceInfo,
+            apps: installed.length ? installed.map(({ name, version }) => ({ name, version })) : [],
```
gre, line 92: (same)

src/renderer/screens/onboarding/steps/GenuineCheck/index.js
```diff
@@ -82,6 +82,20 @@ const GenuineCheck = (props: StepProps) => {
...
+          }),
+        );
+      }
+
+      console.log("result", result);
```
gre, line 97: console log can be removed

PR #3046, opened 2020-07-02
live-common & libs upgrade

gre:
This is a maintenance upgrade. there are no changes but just upgrading libraries to be all in sync and minimize the dup overload.

notable changes: node-hid@^1.3.0

This **does not** include Electron part which is covered by other PR. so it should be "less impacting".


gre:
i'll remove tag important to focus all effort on remaining PR #3045

PR #3071, opened 2020-07-10
setup deep linking feature

valpinkman:
First draft for a deep linking feature

### Type

feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2600


valpinkman:
for QA: 
- if app is not open, it will open it **only**, not navigating directly
- if app is open: magic

Also, validators on delegate not implemented

PR #3072, opened 2020-07-11
Cosmos unified operations

gre:
LL-2762

contains work of https://github.com/LedgerHQ/ledger-live-common/pull/774

use this PR only for validation but this would need that live-common to be merged and a new live-common bump to happen.


gre:
i'll reopen a pr

PR #3073, opened 2020-07-11
Fix for Add account select placeholder

khalilbenihoud:
Fixed accessibility issue on: 
- Add account select placeholder
- Modal Stepper

**Current:**
before

**After the fix:**
after

**Current:**
before

**After the fix:**
Screenshot 2020-07-16 at 10 27 45


src/renderer/components/DropDownSelector.js
```diff
@@ -10,14 +10,14 @@ import Box from "~/renderer/components/Box";
 export const DropDownItem: ThemedComponent<{ isActive: boolean }> = styled(Box).attrs(p => ({
   borderRadius: 1,
   justifyContent: "center",
-  ff: p.isActive ? "Inter|SemiBold" : "Inter",
+  ff: p.isActive ? "Inter|SemiBold" : "Inter|SemiBold",
```
LFBarreto, line 13: just ```ff: "Inter|SemiBold"``` would be enough here

PR #3079, opened 2020-07-15
Fix alignment on icons

khalilbenihoud:
Pixel Polish on the icons alignment inside the call-to-actions.

Current:
current

After the fix:
fix


gre:
why is this tagged important. it's not a release blocker.

PR #3087, opened 2020-07-22
LL-2885 Update link/wording for lostSeed link

juan-cortes:
### Type

Wording

### Context

https://ledgerhq.atlassian.net/browse/LL-2885

Trigger a firmware update and check the wording and link are correct, although I don't see how this one could go wrong.

### DIFF
--- a/src/config/urls.js
+++ b/src/config/urls.js
@@ -43,7 +43,11 @@ export const urls = {
     nanoX: "https://support.ledger.com/hc/en-us/articles/360013349800",
     blue: "https://support.ledger.com/hc/en-us/articles/360005885733",
   },
-  lostPinOrSeed: "https://support.ledger.com/hc/en-us/articles/360000609933",
+  lostPinOrSeed: {
+    nanoS: "https://support.ledger.com/hc/en-us/articles/360000609933",
+    nanoX: "https://support.ledger.com/hc/en-us/articles/360019097794",
+    blue: "https://support.ledger.com/hc/en-us/articles/360000609933",
+  },
   maxSpendable: "https://support.ledger.com/hc/en-us/articles/360012960679",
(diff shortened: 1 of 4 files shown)


juan-cortes:
No qa was assigned to this

gre:
@juan-cortes can you attach the new screenshot so it pass CI?

juan-cortes:
@nabil-brn you are 100% correct, I took it as simply changing the wording and not as making it dynamic as clearly stated in the jira task. I'll address, good QA :beau:

juan-cortes:
@nabil-brn @gre Updated making the URL dynamic depending on the device model. For the Blue, there doesn't seem to be any specific URL for the lost-seed case so I'm just linking to the manual. If you have a better suggestion please update the URL directly with a suggested edit, that would be the quickest (and avoid me having to jump back into this branch), thank you!

dasilvarosa:
> For the Blue, there doesn't seem to be any specific URL for the lost-seed case so I'm just linking to the manual. If you have a better suggestion please update the URL directly with a suggested edit

Please maintain the Ledger Nano S article for the Blue. That's more relevant than the Blue manual.

juan-cortes:
> Could we fix the cursor on hover?
> Comment from @dasilvarosa
> 
> > The link in the release notes (blog post) does not change the cursor in hover state to show that it is clickable

To me it's a totally unrelated issue, this is for the firmware update lost seed link, then again, I went to check it and indeed we don't put the pointer cursor, I've changed it, please check @Arnaud97234 

image

juan-cortes:
> > For the Blue, there doesn't seem to be any specific URL for the lost-seed case so I'm just linking to the manual. If you have a better suggestion please update the URL directly with a suggested edit
> 
> Please maintain the Ledger Nano S article for the Blue. That's more relevant than the Blue manual.

Sure, done

PR #3093, opened 2020-07-27
added a new account selector component

IAmMorrow:
image

This PR feature a state hook for a currency + account combo selector. It can be seen working for the buy crypto account select


src/renderer/components/PerCurrencySelectAccount/index.js
```diff
@@ -0,0 +1,218 @@
...
+  onChange,
+  t,
+  currency,
+  mandatoryTokens,
+  ...props
+}: Props) => {
+  const availableAccounts = useMemo(() => {
+    if (currency.type === "TokenCurrency") {
+      const mainAccounts = getAccountsByCurrency(accounts, currency.parentCurrency);
+      return mainAccounts.map((acc: Account) => {
```
juan-cortes, line 99: This should be using `accountWithMandatoryTokens` instead, just provide the currency and map the accounts, something like `mainAccounts.map(a=>accountWithMandatoryTokens(a, [currency])` would return all accounts with the needed subaccounts without destroying existing ones

src/renderer/components/PerCurrencySelectAccount/index.js
```diff
@@ -0,0 +1,218 @@
...
+  useEffect(() => {
+    setSelectedValue(
+      availableAccounts.length ? availableAccounts[0] : { account: null, subAccount: null },
+    );
+  }, [availableAccounts]);
+
+  const [searchInputValue, setSearchInputValue] = useState("");
+
+  const renderValue = ({ data }: { data: Option }) => {
+    console.log("DATA: ", data);
```
juan-cortes, line 129: Remove the console log

src/renderer/components/PerCurrencySelectAccount/index.js
```diff
@@ -0,0 +1,218 @@
...
+            account: option.account,
+            subAccount: option.subAccount,
+          });
+        }
+        return result;
+      }, []),
+    [searchInputValue, availableAccounts, currency],
+  );
+
+  console.log("SELECTED VALUE: ", selectedValue);
```
juan-cortes, line 188: Same

src/renderer/components/PerCurrencySelectAccount/index.js
gre: My understanding is that accountWithMandatory would indeed simplify some code here and as you will have to do the same on mobile is better as shared code.

But at the same time this seems to be implemented in a different paradigm where values are actually tuples with sub accounts? I'm ok with that PR solution that seems better to keep the account reference and avoid bad surprise.
However that generateTokenAccount must be moved to live-common. Please provide a PR to make it there.

src/renderer/components/PerCurrencySelectAccount/index.js
```diff
@@ -0,0 +1,214 @@
...
+    }
+    return getAccountsByCurrency(accounts, currency).map((acc: Account) => ({
+      account: acc,
+      subAccount: null,
+    }));
+  }, [accounts, currency]);
+
+  const [selectedValue, setSelectedValue] = useState();
+
+  useEffect(() => {
```
gre, line 120: That effect is dangerous because it assumes the accounts prop never update.

This needs to use a ref to statically store the initial value because I think you want it to be just at init?

```
const initialAcc = useRef(availableAccounts)
```

juan-cortes:
Let's get this merged and the show moving 🚀

src/renderer/components/PerCurrencySelectAccount/index.js
IAmMorrow: Yep that's exactly how I've seen it. I find it less convenient to mutate accounts. Account generation was marked for factorisation in live common already. accountWithMandatoryToken should use this func too !

src/renderer/components/PerCurrencySelectAccount/state.js
```diff
@@ -0,0 +1,120 @@
...
+      : { account: null, subAccount: null };
+
+    return {
+      currency: defaultCurrency,
+      account,
+      subAccount,
+    };
+  });
+
+  const { currency, account, subAccount } = state;
```
gre, line 72: when looking at the state currently involved, i think there can still be a case where allAccounts is more updated than the internal state. I mean, there is clearly two single source of trust regarding the reference to account

a solution to this problem is to do this instead:

```js
const account = allAccounts.find(a => a.id===accountId)
const subAccount = (account && account.subAccounts || []).find(a => a.id===subAccountId)
```

where `accountId` and `subAccountId` would be what is actually stored as a state in this hook.

in case of Buy feature, you probably didn't see any issue because we don't care about account updates & it doesn't impact the feature. In case of swap, it's way more dangerous (e.g. balance can updates)

IAmMorrow: Yep but what if the subaccount doesn't actually exist ?

nabil-brn:
Hide token feature affects the balance shown by the select input, it'll always display `0`.
Should we display the real balance, `0` or `Hidden `?

juan-cortes:
Blacklisting a token should not even allow showing the token in the list at all. 
There's no logic in a user hiding a token and then wanting to buy/swap or otherwise interact with it whatsoever 

cc @aramab @gre

gre:
the feedback is legit but i'm not sure how we want to solve it

- one way is to explicitly tell the user this token has been hidden by him.
- another way is to make select work regardless, but way dangerous and not even sure if feasible technically.
- another way is just to let this happening and considering it for later. (this is my recommendation, this is a too rare case to be taken too much dev effort, let's ship things first)

juan-cortes:
>_Some checks were not successful_

PR #3099, opened 2020-07-29
LL-2995 Carousel rework

juan-cortes:
Group

image

Rework the carousel to make it less prominent on the screen without, it now has a smaller footprint, blends more with the rest of the content and most importantly the dismiss will be permanent until we have a new announcement to show. A user can toggle the carousel's visibility from the settings screen too if they want to.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2995

### Test plan
Make sure the carousel can be dismissed via the X CTA on the banner itself. The action that dispatches the setting change is the primary button "Close" so if the user doesn't confirm, the carousel will still be visible upon navigation. 
Make sure the setting is updated correctly in Settings/General and that the visibility can be toggle with it.


src/renderer/components/Carousel/index.js
```diff
@@ -143,21 +195,41 @@ const Carousel = ({
   }, [index, slides.length]);
 
   const onDismiss = useCallback(() => {
-    dispatch(setDismissedCarousel(true));
-  }, [dispatch]);
+    setWantToDismiss(true);
+  }, []);
```
valpinkman, line 199: are you sure you don't want to add `setWantToDismiss` to the dependencies of `useCallback` ?

src/renderer/components/Carousel/index.js
```diff
@@ -143,21 +195,41 @@ const Carousel = ({
...
+
+  const onUndo = useCallback(() => {
+    setWantToDismiss(false);
+  }, []);
```
valpinkman, line 203: same here

src/renderer/components/Carousel/index.js
juan-cortes: Pretty sure, according to the [faq on hooks](https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often) referring to a `useState` setter function's presence in the dependencies of a `useCallback`:

> _(The identity of the setCount function is guaranteed to be stable so it’s safe to omit.)_

juan-cortes:
image
@aramab @gre latest wording from figma

Arnaud97234:
I validated that PR because everything listed above is correct, but we don't have the auto-switch between element anymore.

juan-cortes:
Then it's not correct, ill check again in a bit

PR #3104, opened 2020-08-03
LL-2685 Add buy cta to notEnoughGas

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-2685

Clicking on the action "Buy ETH" button should dismiss any open modal and navigate to the `Buy` feature screen where the user can start a buy flow for the selected currency.  The selected currency (in this case only eth would trigger the UI) should be pre-selected in the buy screen. This pre-selection will not work if you are already in the buy screen.

### DIFF
--- a/src/renderer/components/BuyButton.js
+++ b/src/renderer/components/BuyButton.js
@@ -1,0 +1,32 @@
+// @flow
+
+import React, { useCallback } from "react";
+import { Trans } from "react-i18next";
+import Button from "~/renderer/components/Button";
+import { useHistory } from "react-router-dom";
+import { closeAllModal } from "~/renderer/actions/modals";
+import { useDispatch } from "react-redux";
+import type { CryptoCurrency } from "@ledgerhq/live-common/lib/types";
+
+const BuyButton = ({ currency }: { currency: CryptoCurrency }) => {
+  const history = useHistory();
+  const dispatch = useDispatch();
+
+  const onClick = useCallback(() => {
+    dispatch(closeAllModal());
+    history.push({
+      pathname: "/exchange",
+      state: {
+        defaultCurrency: currency,
+      },
+    });
+  }, [currency, dispatch, history]);
+
+  return (
+    <Button mr={1} primary inverted onClick={onClick}>
+      <Trans i18nKey="buy.buyCTA" values={{ currencyTicker: currency.ticker }} />
+    </Button>
+  );
+};
+
+export default BuyButton;
--- a/src/renderer/screens/exchange/index.js
+++ b/src/renderer/screens/exchange/index.js
@@ -2,6 +2,7 @@
 import React, { useState } from "react";
 import styled from "styled-components";
 import Box from "~/renderer/components/Box";
+import { useLocation } from "react-router-dom";
 import type { ThemedComponent } from "~/renderer/styles/StyleProvider";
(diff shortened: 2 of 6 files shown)


juan-cortes:
Discussing the actual UI of this 😪

src/renderer/screens/exchange/index.js
```diff
@@ -38,7 +42,7 @@ const Exchange = () => {
       </Box>
       <TabBar tabs={tabs.map(tab => t(tab.title))} onIndexChange={setActiveTabIndex} />
       <Card grow style={{ overflow: "hidden" }}>
-        {tabs[activeTabIndex].component}
+        <Component defaultCurrency={maybeDefaultCurrency} />
```
gre, line 45: i think a more generic way would be to pass `{...location?.state}`

juan-cortes: 🤷 whatever you say chief, updated!

src/renderer/screens/exchange/index.js
```diff
@@ -18,17 +19,19 @@ const Container: ThemedComponent<{ selectable: boolean, pb: number }> = styled(B
...
-    component: <History />,
+    component: History,
   },
 ];
 
 const Exchange = () => {
   const [activeTabIndex, setActiveTabIndex] = useState(0);
+  const location = useLocation();
   const { t } = useTranslation();
+  const Component = useMemo(() => tabs[activeTabIndex].component, [activeTabIndex]);
```
gre, line 34: this do not need to be a memo, there is nothing to memorize, we can directly do `Component = tabs[activeTabIndex].component`

src/renderer/components/BuyButton.js
```diff
@@ -0,0 +1,35 @@
...
+import { Trans } from "react-i18next";
+import Button from "~/renderer/components/Button";
+import { useHistory } from "react-router-dom";
+import { closeAllModal } from "~/renderer/actions/modals";
+import { useDispatch } from "react-redux";
+import type { TokenCurrency, CryptoCurrency } from "@ledgerhq/live-common/lib/types";
+import { getCryptoCurrencyById } from "@ledgerhq/live-common/lib/currencies";
+
+// NB Defaulting to ethereum since it's the only one throwing the NotEnoughGas currently
+const BuyButton = ({ currency: _c }: { currency?: CryptoCurrency | TokenCurrency }) => {
```
gre, line 13: there are I think two problem in this:

- the fact it's optional and infer it's "ethereum" is dangerous
- the fact it's `| TokenCurrency` does not seem covered, as you don't pass anything on other hand but just default to ethereum, it is not tested. 

TLDR: can we just `currency: CryptoCurrency` ?

- ethereum_classic could throw NotEnoughGas (as we converge to same impl), other new coin could in the future
- on the other side, you know what coin is involved as it's basically `mainAccount.currency`

src/renderer/screens/exchange/index.js
juan-cortes: Makes sense, addressed

src/renderer/components/BuyButton.js
juan-cortes: Indeed, it was originally optional because the first design had the cta inside the amount field, and the amount field didn't have the currency and I was lazy. With it being in the footer and having direct access to the currency, this is cleaner. Addressed.

PR #3115, opened 2020-08-11
update libcore and live-common

gre:
- libcore 3.5.0 https://github.com/LedgerHQ/lib-ledger-core/blob/master/CHANGELOG.md#350
- live-common 13.7.0 https://github.com/LedgerHQ/ledger-live-common/releases/tag/v13.7.0

⚠️  not mergeable due to regression in libcore. we can use this PR to test more things.

- https://ledgerhq.atlassian.net/browse/LLC-696


gre:
@Arnaud97234 @nabil-brn  this PR is clearly not mergeable because regression on Tezos, however, it would be interesting to assess there are no other regression, **especially on Bitcoin and bitcoin forks** and around the UTXOs subject. I recently revealed that excluding some UTXO would have bugs using latest libcore: like basically you uncheck an utxo and it sometimes would not see the other possibility to build the transaction even when you would be able to. (exemple: you have 3 utxo of 100 each, you try to spend 50. you uncheck one, it starts saying "Not enough balance" even tho the 2 other utxos could be of use)

gre:
a new libcore has been available. we'll retest the latest stuff

PR #3119, opened 2020-08-12
LL-2676 Auto select currency on  account buy cta click

juan-cortes:
When in an account for a currency that allows us to buy funds through LLD, clicking on the Buy button should preselect that currency and account. In order to do this we've introduced a default currency and default account concept to the new account selector, also fixes a regression in the cta to buy from a token send where the parent ETH account didn't have enough funds, the introduction of the new component broke the auto currency select in that case. This addresses that.

### Type

Feature + Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-2676

Any time we click on a buy cta, it should pre select the currency and account on the target form.


gre:
please rebase develop

juan-cortes:
@gre done

PR #3120, opened 2020-08-13
Switch BCH address formatting to cashaddr

### LINKED ISSUE #980: Cashaddr format for Bitcoin Cash

- Ledger Live **version_here**
- Platform: **windows OR mac OR linux**

gre:
Fixes #980
depends on https://github.com/LedgerHQ/ledger-live-common/pull/813

### Type

feature

### Context

LL-2806

- all existing bch accounts will be formatting the address as cashaddr. However, a synchronisation will be required in order for the account to update that state, receive flow might temporarily tell "something wrong happened" if the sync is not waited, it's a limited tradeoff here.
- when the sync is complete, it is expected that all operations will have their inputs, outputs formatted as cashaddr
- it is also expected that a Receive will correctly format the device as cashaddr and the UI as well.
- the cashaddr format chosen **do not includes the `bitcoincash:`** prefix because the device didn't. It also makes things more concise, cashaddr spec clearly says that this can be omitted by wallets.
- the impact of live-common needs to be taken into consideration as all Receive flow has been reworked to allow the bridge to expose a custom `receive`. It also impact the Buy Crypto as it's also doing a receive.
- **It is recommended to QA this PR first so we can know if we can merge the live-common PR and do a bump there, before merging this current PR.**


gre:
as it passed QA, i'll merge this now otherwise it's going to collide with incoming PRs =)

gre:
@chaintip fixes #980

PR #3123, opened 2020-08-17
Use Device type from live-common

juan-cortes:
Stop using the export from the Device reducers in favour of the one exported from live-common actions.

### Type

Code Quality improvement

### Testing
I only tested the send, receive, manager install+uninstall flows, not sure if the QA effort on this needs to touch all flows or not.


gre:
as this was solved as part of swap, we'll keep it in swap for now as this work as uncomplete

PR #3124, opened 2020-08-17
Better UI for searchable selects

juan-cortes:
This is the result of a proposal the @bscholving made regarding searchable selectors in the app, it should be plug and play and not break any existing selectors. Basically if it's searchable and the list is displayed we change the UI to look like on the screenshots. If not, we keep it as is.

image
image
image
image
image

### Type

UI Polish

### Context

Slack conversation

Any flow that has a selector, check it didn't break.


Arnaud97234 (changes requested):
### Bug found
Add account: When the drop down list is displayed the placeholder and the Search text are both displayed one hover the other.

image

juan-cortes:
I'll fix it tomorrow, other selects look good on your end? @Arnaud97234

juan-cortes:
Current version should have the add accounts flow fixed @Arnaud97234

Arnaud97234 (approved):

PR #3133, opened 2020-08-25
bump libcore version 6.8.6

LFBarreto:
Should fix Cosmos regressions

test Cosmos and Eth send flow


Arnaud97234 (approved):
Cosmos and ETH are OK. Clear cache is OK. Works properly on Winddows.

PR #3147, opened 2020-09-01
update and migration banner now appear on top of carousel

IAmMorrow:


Arnaud97234 (changes requested):
### Bug found

1) When no banner displayed padding top should be smaller.
image
_____

2) When a banner is displayed the padding between the banner and the carousel should be smaller.
image

valpinkman (approved):

Arnaud97234 (approved):

PR #3151, opened 2020-09-02
add xpub warning on advanced account settings (bitcoin family)

valpinkman:
Adds and info banner into Account => Settings => Advanced warning about the xpub use. 

Only for Bitcoin family

### Type

Feat

### Context

LL-2039

### DIFF
--- a/src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
+++ b/src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
@@ -153,8 +153,6 @@ class AccountSettingRenderBody extends PureComponent<Props, State> {
 
     const account = this.getAccount(data);
 
-    console.log({ account });
-
     const usefulData = {
       xpub: account.xpub || undefined,
       index: account.index,


Arnaud97234 (dismissed):

src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
```diff
@@ -147,6 +152,9 @@ class AccountSettingRenderBody extends PureComponent<Props, State> {
     if (!data) return null;
 
     const account = this.getAccount(data);
+
+    console.log({ account });
```
juan-cortes, line 156: Remove, right?

juan-cortes (changes requested):

src/renderer/modals/SettingsAccount/AccountSettingRenderBody.js
juan-cortes:
```suggestion
```

PR #3156, opened 2020-09-04
on lock screen, using the keyboard will automatically focus the passw…

IAmMorrow:
…ord input if it's not


gre (approved):
autoFocus wouldn't work? I'm fine with this solution otherwise.

Arnaud97234 (approved):

PR #3169, opened 2020-09-09
LL-3172 (AccountPage): add an exchange CTA

LFBarreto:
Group exchange action into one drop down select.

localhost_8080_webpack_index html_theme=null

### Type

UI Polish

### Context

LL-3172

Account pages

### DIFF
--- a/src/renderer/screens/account/AccountHeaderActions.js
+++ b/src/renderer/screens/account/AccountHeaderActions.js
@@ -3,7 +3,7 @@
 import React, { useCallback } from "react";
 import { compose } from "redux";
 import { connect } from "react-redux";
-import { withTranslation } from "react-i18next";
+import { withTranslation, Trans } from "react-i18next";
 import styled from "styled-components";
 import type { Account, AccountLike } from "@ledgerhq/live-common/lib/types";
 import Tooltip from "~/renderer/components/Tooltip";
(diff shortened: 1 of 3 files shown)


src/renderer/screens/account/AccountHeaderActions.js
```diff
@@ -87,6 +83,47 @@ const AccountHeaderActions = ({ account, parentAccount, openModal, t }: Props) =
     });
   }, [currency, history, mainAccount]);
 
+  // List of available exchange actions
+  const actions = [
+    ...(availableOnExchange
```
juan-cortes, line 88: Since there's no unified list of crypto used for all exchange features, we'll need individual checks per action (buy/swap/sell) but I'll adapt this on swap side if needed.

src/renderer/screens/account/AccountHeaderActions.js
```diff
@@ -97,7 +134,27 @@ const AccountHeaderActions = ({ account, parentAccount, openModal, t }: Props) =
           ) : null}
 
           <ReceiveAction account={account} parentAccount={parentAccount} onClick={onReceive} />
-          {availableOnExchange ? <BuyActionDefault currency={currency} onClick={onBuy} /> : null}
+          {actions && actions.length > 0 && (
```
juan-cortes, line 137: I believe we agreed on not using the logical `&&` operator to return jsx conditionally because we fell into crashes that were a nightmare to locate (in the generated asar file) and use ternaries with a null fallback cc @gre

juan-cortes (changes requested):

src/renderer/screens/account/AccountHeaderActions.js
LFBarreto: Ho yes indeed when you have a single return statement it can do a React render error :+1:

src/renderer/screens/account/AccountHeaderActions.js
LFBarreto: For this one we might need to port the feature flagging of exchange features into live-common :+1:

juan-cortes (approved):

Arnaud97234 (approved):

PR #3180, opened 2020-09-21
live-common 14.7.0

gre:
- ensure that there is no regression on the manager entrance (as device flow logic has changed)


gre:
#3192 replaces this PR.

PR #3185, opened 2020-09-23
PoC of Websocket Bridge

gre:
This implements an escape hatch to allow Ledger Live to expose the device over WebSocket for web app that requires the device but when U2F/webusb is not possible / not working.

This allows a web app to link the user to, for instance `ledgerlive://bridge?appName=Ethereum&origin=mydomain.com` which will open the Bridge on Ledger Live.

**Here is basically how it works:**

- user open via an intent (using URI scheme) to use a specific appName and an origin, Ledger Live will open a modal that requests the user to connect the device and open the correct app:

Capture d’écran 2020-09-23 à 09 04 07

- Once the user open the app on his device, a WebSocket server on port 8435 starts (like `ledger-live proxy` CLI) and will allow clients to run APDU to the device. The client must be run from the same domain as provided by origin, if provided.
- The modals is establishing the a bi directionnel channel to the device. As soon as the user close that websocket channel, the modals turns into the success final step and the server is stopped.
- When the user closes the modal, the WebSocket server will also be stopped.

### To be determined / To be implemented

- [x] the origin, which is uses to deny websocket client that wouldn't be on same domain, is not yet implemented. It's also not clear if we want or not to allow it to be optional.
- [ ] The fact the device can BLOCK on apdus makes me think we might want to BLOCK the modal each time an apdu run. Typically we don't want to allow a web app to start doing something with the device (like a verify address) and to let the user close the modal and go do something else, it's both bad in term of UX and security.
- [ ] It it not clear yet on the behavior of two clients consuming at the same time, we likely want to deny the second connection
- [ ] UX: to be implemented the various cases where the WebSocket server can fail to be created (e.g. port in use). probably reuse the GenericError component.
- [ ] What should we do when the user disconnect the device during the bridge. Think about what it means for user land too.
- [ ] validate that URI scheme is enough and drop the experimental row.

### Note on user land integration

For a web page it would mean:

(1) Adding some kind of `<a href="ledgerlive://bridge?appName=Ethereum&origin=mydomain.com">Open Ledger Live Bridge</a>` to the web app
(2) on click of that bridge (or not even on click), actively polling for websocket availability.
(3) when the WebSocket is connected, you can use it to communicate to the device.

Please check a full example at https://ledger-live-tools.now.sh/bridgetest

Source code: https://github.com/LedgerHQ/ledger-live-common/blob/master/tools/src/demos/bridgetest/index.js

### Type

escape hatch

### Context

LL-3391

### How to test

For testing the URI scheme, you need the app builded. You can grab one of the build that our CI will generate on this PR.

### DIFF
--- a/src/internal/commands/index.js
+++ b/src/internal/commands/index.js
@@ -17,6 +17,7 @@ import testCrash from "./testCrash";
 import testInterval from "./testInterval";
 import appOpExec from "./appOpExec";
 import initSwap from "./initSwap";
+import websocketBridge from "./websocketBridge";
 import { commands as bridgeProxyCommands } from "~/renderer/bridge/proxy-commands";
 
 export const commandsById = {
@@ -39,6 +40,7 @@ export const commandsById = {
   initSwap,
   testCrash,
   testInterval,
+  websocketBridge,
 };
 
 export type Commands = typeof commandsById;
--- a/src/internal/commands/websocketBridge.js
+++ b/src/internal/commands/websocketBridge.js
@@ -1,0 +1,137 @@
+// @flow
+import { Observable } from "rxjs";
+import { log, listen } from "@ledgerhq/logs";
+import { open } from "@ledgerhq/live-common/lib/hw";
+import WebSocket from "ws";
+
+type Input = {
+  origin: ?string,
+  deviceId: string,
+};
+type Result = { log: string };
+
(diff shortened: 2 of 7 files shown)


gre:
work is sent in https://github.com/LedgerHQ/ledger-live-desktop/pull/3257

PR #3191, opened 2020-09-24
LL-3337 fix deeplinking uri schemes

valpinkman:
deeplinking now works with all type of uri schemes
- ledger-live:portfolio
- ledger-live://portfolio (or any number of slashes you want)

Does not crash on `ledger-live:` anymore

### Type

Bug Fix

### Context

LL-3337


gre (approved):

Arnaud97234 (approved):
It's OK for me.
Should use `ledgerlive:` instead of `ledger-live:`

src/renderer/hooks/useDeeplinking.js
```diff
@@ -47,9 +46,12 @@ function useDeepLinkHandler() {
 
   const handler = useCallback(
     (event: any, deeplink: string) => {
-      const [, path] = deeplink.split("ledgerlive://");
+      const { pathname, searchParams } = new URL(deeplink);
+      const query = Object.fromEntries(searchParams);
 
-      const { url, query } = querystring.parseUrl(path);
```
gre, line 52: there is one thing i didn't spot: the logic no longer verify the domain is ledgerlive. I think it's necessary when we will start supporting more than just ledgerlive://

not blocking this merge, but we will need to be aware for future

PR #3192, opened 2020-09-24
@ledgerhq/cryptoassets

gre:
depends on https://github.com/LedgerHQ/ledger-live-common/pull/888

assert that the application still works and especially that:
- ERC20 transaction works
- TRC* transaction works
- ASA transaction works


Arnaud97234 (approved):
- Send ERC20 OK
- Send TRC10 / TRC20 OK
- Send ASA OK

PR #3198, opened 2020-09-29
LL-3207 Prevent closing the coin control modal if it has errors

juan-cortes:
To avoid a deadlock after an invalid selection in the coin control modal, disable closing said modal if the status reports errors.
This is not what was specified on the Jira task but I believe it makes more sense to not allow closing the coin control modal if the transaction would become invalid than allowing to re open it.

### Type

Bug Fix
### Context

https://ledgerhq.atlassian.net/browse/LL-3207

- Open the coin control modal and make it so the transaction throws an error
- Try to close the modal, it should not be possible
- Select some UTXOs to make the transaction valid again
- Close the modal

### DIFF
--- a/src/renderer/families/bitcoin/CoinControlModal.js
+++ b/src/renderer/families/bitcoin/CoinControlModal.js
@@ -62,14 +62,15 @@ const CoinControlModal = ({
   const error = errorKeys.length ? status.errors[errorKeys[0]] : null;
 
   const returning = (status.txOutputs || []).find(tx => !!tx.path);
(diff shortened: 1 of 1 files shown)


juan-cortes:
wtf screenshot not loading the text :trollface:

gre (commented):
Ok with this solution. Did we address somewhere else the fact we need to use utxos.length for accessing the screen? I'm not sure it was in LL-3207 but we mentioned it somewhere. Cc @IAmMorrow

Arnaud97234 (approved):

PR #3199, opened 2020-09-29
LL-3422 Preload the portfolio behind the onboarding

juan-cortes:
Instead of loading the onboarding OR the dashboard, load both. Since the onboarding is absolutely positioned and takes the whole screen we can't see any difference. Once the onboarding is finished, the other content becomes visible and has everything already rendered an in place

### Type

UI Polish (and testing)

### Context

https://ledgerhq.atlassian.net/browse/LL-3422

- Launch the bullrun
- Record the bullrun
- Go frame by frame and try to see one where after the confetti screen the no-accounts illustration hasn't loaded but the text has loaded. If you don't find one, we good.

### DIFF
--- a/src/renderer/components/OnboardingOrElse.js
+++ b/src/renderer/components/OnboardingOrElse.js
@@ -13,12 +13,14 @@ type Props = {
 const OnboardingOrElse = ({ children }: Props) => {
   const hasCompletedOnboarding = useSelector(hasCompletedOnboardingSelector);
   const onboardingRelaunched = useSelector(onboardingRelaunchedSelector);
+  const showOnboarding = !hasCompletedOnboarding || onboardingRelaunched;
 
-  if (!hasCompletedOnboarding || onboardingRelaunched) {
-    return <Onboarding />;
-  }
-
-  return children;
+  return (
+    <>
+      {showOnboarding ? <Onboarding /> : null}
+      {children}
+    </>
+  );
 };
 
 const ConnectedOnboardingOrElse: React$ComponentType<Props> = memo(OnboardingOrElse);


gre:
I would be careful with this approach because I anticipate there might be problem. It sounds like any effect we have on the app will now happen on the onboarding.

The first thing I'm thinking of is that this will likely glitch the analytics, triggering events during the onboarding.

I understand the goal is to preload the images but maybe there is another way. If we confirm there are no issue, we can keep it but it's not trivial to QA this current PR.

juan-cortes:
> I would be careful with this approach because I anticipate there might be problem. It sounds like any effect we have on the app will now happen on the onboarding.
> 
> The first thing I'm thinking of is that this will likely glitch the analytics, triggering events during the onboarding.
> 
> I understand the goal is to preload the images but maybe there is another way (like rendering in background the only parts we know create issues). If we confirm there are no issue, we can keep it but it's not trivial to QA this current PR, and we don't have a lot of qa buffer this week.

Fair enough for the analytics part. There should be no other side effects since we'd have no accounts and no apps installed so there would be no sync, no graphs, nothing else. Perhaps not triggering the analytics if it's not visible or something like that. Maybe leave it for now and retackle it at some other point.

Arnaud97234 (changes requested):
### Bug found
Regression.
"Terms of use" modal shouldn't be displayed before the onboarding.
ezgif com-gif-maker

### How to reproduce
- Clean LL user data
- Start LL

### Expected behaviour
The "Terms of use" modal should be displayed after onboarding success screen (After the confetti)

juan-cortes:
> ### Bug found
> Regression.
> "Terms of use" modal shouldn't be displayed before the onboarding.
> ezgif com-gif-maker
> 
> ### How to reproduce
> * Clean LL user data
> * Start LL
> 
> ### Expected behaviour
> The "Terms of use" modal should be displayed after onboarding success screen (After the confetti)

Fair enough, then we can kill this approach as far as I'm concerned, I couldn't get the image to preload without pre-rendering it. Perhaps an even hackier way where we render it somewhere else and move it, but not on this PR. Closing it.

PR #3202, opened 2020-09-30
Add tracking events for swap feature

juan-cortes:
Added some events to the feature. The task lists "listApps and failed listApps" as events, no idea how we want to do that in this context so I didn't. The rest of the events should be ok.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3543

Don't know how to test this.

### DIFF
--- a/src/renderer/modals/Swap/SwapBody.js
+++ b/src/renderer/modals/Swap/SwapBody.js
@@ -12,6 +12,8 @@ import { useDispatch } from "react-redux";
 import { updateAccountWithUpdater } from "~/renderer/actions/accounts";
 import { addPendingOperation, getMainAccount } from "@ledgerhq/live-common/lib/account";
 import addToSwapHistory from "@ledgerhq/live-common/lib/swap/addToSwapHistory";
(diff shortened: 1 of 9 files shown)


src/renderer/modals/Swap/SwapBody.js
```diff
@@ -81,6 +82,7 @@ const SwapBody = ({
       title={<Trans i18nKey="swap.modal.title" />}
       render={() => (
         <>
+          <TrackPage category="Swap" name={`ModalStep${activeStep + 1}`} />
```
gre, line 85: It's better to use the name of the step in case order changes in future

juan-cortes: Sure, makes sense

juan-cortes: In fact, `activeStep` is not even a number but rather the id already, so this would've done something like `ModalStepfinished1` which makes no sense. Fixed

gre (dismissed):

gre:
@juan-cortes  conflicts

gre (approved):

PR #3205, opened 2020-10-01
Swap - Rework the getProviders logic for better privacy

juan-cortes:
### Description

Before the launch of the feature to the public this logical rework was due, it involves no longer fetching the available providers at app launch and only doing so after the user has accepted the KYC (ip sharing for geo-locking of the feature) checkbox and only inside the swap feature itself. The cached providers will now be in the settings internal state and therefor persisted between launches unlike the previous version where they were volatile and cleared on app close. This PR also includes some code cleanup to be closer to the structure of LLM when possible for easier switching between the two environments.

### Type

Feature

### Context

Internal discussion

All the swap feature should work the same as before, make sure that you first have to accept the KYC checkbox before any providers are fetch, and that before that no CTA on specific accounts to access swap are visible. If we reach a "_Swap feature not available_" state then the CTAs would once again be missing since we can't offer the feature.


gre:
lgtm to me. tested a swap on Windows.

PR #3206, opened 2020-10-01
Adds the swap banner for the portfolio

juan-cortes:
Adds the banner cta to access the swap feature to the carousel with the fancy parallax and all. Note that I put it in third place, taking the previous position held by staking Cosmos, since it's the logical place after the internal ones (I think)
banner

### Type

Feature

Make sure the banner shows, and links to the feature.


gre:
screenshot like a boss

gre:
tested. I think it's good enough. 

one general feedback on the banner: the click area around the arrows is still too small, it's hard to click on them and i am often "click jacked" when clicking too near to the border.

PR #3218, opened 2020-10-07
LL-2907 (AddAccounts): token conditional cta added)

LFBarreto:
(AddAccounts): token conditional cta added)

ezgif-3-ce617e0f0627

### Type

Feature
### Context

LL-2907

Add token accounts flow. refer to given jira tasks for cta conditions.


src/renderer/modals/AddAccounts/steps/StepChooseCurrency.js
```diff
@@ -54,25 +57,97 @@ const StepChooseCurrency = ({ currency, setCurrency }: StepProps) => {
...
+  const onTokenCta = useCallback(() => {
+    if (parentTokenAccount) {
+      onCloseModal();
+      dispatch(
+        openModal(
+          "MODAL_RECEIVE",
+          tokenAccount // if already has token receive directly to it
+            ? {
+                account: tokenAccount,
+                parentAccount: parentTokenAccount,
```
gre, line 104: by the presence of this `{account,parentAccount}` pair, I have the feeling we assume that the parentAccount is account of account,
but it is not necessarily the case look at the current logic you have above

```
  const parentTokenAccount =
    isToken &&
    parentCurrency &&
    flatAccounts.find(acc => {
      const c = getAccountCurrency(acc);
      return c.id === parentCurrency.id;
    });

  const tokenAccount =
    isToken &&
    flatAccounts.find(acc => {
      const c = getAccountCurrency(acc);
      // $FlowFixMe
      return c.id === currency.id;
    });
```

we seem to take the first match of the current and token but independently so it will pick the first, regardless of it one belong to the other.

My second feedback regarding this is I think we can be much more optimized and minimal if we just implement it like this:


```
function findFirstTokenAccount (token, accounts) {
  for (const parentAccount of accounts) {
    for (const account of accounts.subAccounts) {
      if (...) {
        return { account, parentAccount } 
      }
    }
  }
}
```

function that would make a lot of sense to be in live-common.

There are a lot of places where we tend to prefer using the flattenAccounts but it is often minimal and better to stick to the nested version.

(pseudo code, missing some part, but you get the idea)

gre (changes requested):
my understanding of the logic shows a potential issue and also an algorithm that could be a bit simplified

src/renderer/modals/AddAccounts/steps/StepChooseCurrency.js
LFBarreto: I updated it accordingly i'll create a PR on live-common with this function :+1:

Arnaud97234 (approved):

gre (approved):

Arnaud97234 (approved):

PR #3221, opened 2020-10-07
LL-3207 Improved coin control modal

juan-cortes:
Second attempt at a better coin control modal including the feedback from Charles, which hopefully leaves everyone happy. Or as happy as we can be with a stacked modal on the app :trollface: 

image
image
The coin control modal is only available on accounts with some UTXO, so we can't access the second modal on an empty account for instance.
image
When there are errors in the transaction, we don't display them here. For instance, this state would normally show the "amount is required" error, but we can't fix that from the second modal, so we just not show it. You can tell there's an error because of the footer missing information.
image
Prevent disabling the last UTXO, this fixes the issue of closing the modal and not being able to open it again because all UTXO are excluded for the account. This does not break the previous logic of not allowing the selection of unconfirmed coins or whatever was present.

image
When doing a send max, adding/excluding UTXO will automatically adapt the balance in footer (like before) without breaking.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3207

- Follow the screenshots
- Try to reach a locked state like before
- Maybe improve the wording, I might not have chose the best words.


gre (dismissed):

Arnaud97234 (dismissed):
### Bug found

Design should be different for utxo whether they are Excluded (replaceable) or Selected (Only one remaining).

image
image

The replaceable utxo is excluded and can't be selected
____

image
image

 The replaceable utxo is the only one selected. Maybe the utxo block shouldn't be greyed out when it's the last one unchecked.

juan-cortes:
I did nothing to the ui but merely the flow. If this is a problem, then we need to agree on how it should look, I don't care either way so @gre weigh in

gre:
when activating "Allow to be replaceable" everything is supposed to be checkable, so indeed there is a bug in the UI here. not saying this PR has regressed on this but this is important to fix it, maybe in the same PR as we are here.

juan-cortes:
Ah, gotcha, I misunderstood, it's supposed to not show as disabled because it's selectable
It's just a matter of the states taking priority, I'll take a look. Is that what you meant @Arnaud97234 i misunderstood

juan-cortes:
After the call with @Arnaud97234 I think the behaviour should be OK now, will need dev review for the changes though @gre

gre (approved):

Arnaud97234 (approved):

PR #3227, opened 2020-10-07
LL-3550 Add currency name to add account flow tracked events

juan-cortes:
### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3550

Don't know how to validate these since the exported logs for tracked events don't include extra properties cc @gre


juan-cortes:
I initially added the `currencyName` to the first step but I think that since it's a track page it will not send another event after we select the currency so it's useless. Please confirm @gre this is the case

gre:
yeah i think we can only focus on the step 2+ because nothing really started on step1

gre (approved):

Arnaud97234 (approved):

PR #3246, opened 2020-10-21
fix ctrl-r not working when app crashes

machard:
Fixing the fact that ctrl-r was not working when the app was crashed

### Type

Bug Fix

write something wrong, see the error page, and hit ctrl-r

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -1,5 +1,5 @@
 // @flow
-import React from "react";
+import React, { useEffect } from "react";
 import { hot } from "react-hot-loader/root";
 import { Provider } from "react-redux";
 import type { Store } from "redux";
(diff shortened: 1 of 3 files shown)


gre:
This needs to be documented in the code.
**But this the correct behavior** to block the user to be able to Ctrl R during a critical error case. Because critical error case usually mean a crash at boot and a complete failure to set up the internal process. Ctrl R being only reloading the renderer, allowing it would permit user to bypass the problem but then experience many broken parts (literally all device actions are broken, etc. )

gre (changes requested):
Check comment

machard:
Ok. So users can refresh when not in the crash screen, and dev can do it all the time ?

gre (approved):

valpinkman (approved):

nabil-brn (changes requested):
I'm able to [**cmd** + **R**] with a prod build (`yarn dist`) on a crash page.
Oct-30-2020 13-07-56

---
13c61762 (HEAD -> fix/ctrl-r-not-working-when-fatal-error-occured, machard/fix/ctrl-r-not-working-when-fatal-error-occured) only allow reload on crash screen if in __DEV__

nabil-brn (approved):
✓ QA-ed

---
fe1c7232 (HEAD -> fix/ctrl-r-not-working-when-fatal-error-occured, machard/fix/ctrl-r-not-working-when-fatal-error-occured) fix handler subscription

PR #3248, opened 2020-10-21
Integrate latest ethjs evolution

gre:
- This works fully migrate ETH to use the JS implementation.
- It also uses latest work of the ETH api (staging, which is testable with Experimental Explorers) but is still backward compatible with production (which is without Experimental Explorers).
- reworked the gas price component to use the range provided by network (if experimental on)

it is great to confirm this is good to go so we can merge the live-common PR.

### Type

update

### Context

https://github.com/LedgerHQ/ledger-live-common/pull/912

all accounts are migrated to ETHjs.
gas price works and you can do ETH/ERC20 tx without an issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@ledgerhq/hw-transport-http": "^5.26.0",
     "@ledgerhq/hw-transport-node-hid-singleton": "^5.26.0",
     "@ledgerhq/ledger-core": "6.9.1",
-    "@ledgerhq/live-common": "^15.5.0-ethjs.5",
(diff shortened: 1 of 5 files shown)


Arnaud97234 (changes requested):
ETH optimistic fees tx should be type of "Fees" instead of "Send"
image

nabil-brn (approved):
✓ QA-ed

Tested ETH, ERC20, BTC, ALGO, ASA, TRX, TRC20

---
Ledger Live 2.15.0-sha.a8450c0

valpinkman (approved):

gre:
thx for feedback. fixed now.

PR #3252, opened 2020-10-22
allow buy flow without device

machard:
Allow to buy without device

### Type

Feature

### Context

https://ledgerhq.atlassian.net/secure/RapidBoard.jspa?rapidView=62&projectKey=LL&modal=detail&selectedIssue=LL-3688&quickFilter=150

1. Test buy flow with a ledger
2. Test buy flow without a ledger


gre (approved):

gre:
i'll let you check with Chris if the design is good & then we can send to 🇶🇦

machard:
flow 1, checking that the device is connected (required new translations strings):
Capture d’écran 2020-10-23 à 09 59 40
Capture d’écran 2020-10-23 à 09 59 50

flow 2, checking address (same same as receive flow):
Capture d’écran 2020-10-23 à 10 00 21
Capture d’écran 2020-10-23 à 10 00 30

Arnaud97234 (changes requested):
In the second device interaction the following use case is not covered:
- User click "Don't have your device?"
_Result: User lands on the modal where the address is displayed_
We should display 2 CTA: "Continue" and "Verify", if user click on "Verify" the device flow should restart and the address should be displayed on the device

Arnaud97234 (commented):
@cthepot-ledger If it's ok on your end then for me it's a go

gre:
@Arnaud97234 we actually rescheduled this for a **future** sprint (not the one that started today) because we will need design for mobile to do the same too.

cthepot-ledger:
@machard some UI changes requires on second modal

cthepot-ledger:
GO for me after 2 loops.

nabil-brn (changes requested):
The iframe is blank when I select BTC or ETH and I get this error in the console:
`TypeError: Cannot read property 'BTC' of undefined        instrument.ts:129`

---
51415b6a3 (HEAD -> LL-3688, machard/LL-3688) Merge branch 'develop' into LL-3688

nabil-brn (approved):
✓ QA-ed

---
2.20.0-sha.5bc7693a6ca079ab410a93adfdc151377187e8c1

gre (approved):

PR #3264, opened 2020-10-29
Let it snow

juan-cortes:
This can be tested with the `SNOW_EVENT=1` env, or by trusting and waiting for the date.
Should only be visible when modals are out, shouldn't affect anything else.

Kudos to @IAmMorrow , I just brought it back

### Type

UI Treat


gre (approved):

nabil-brn (approved):
✓ QA-ed

---
4117b727 (HEAD -> snow, juan/snow) Let it snow

PR #3265, opened 2020-10-29
LL-3774 Add a warning banner for ongoing scam campaigns

juan-cortes:
image

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-3774

- Test that the notice shows, test that it can be dismissed.


gre (approved):

gre:
i'm ok with this, but it's probably the time to also have it auto fetched using the config.json. we'll rediscuss later.

src/renderer/components/banners/OngoingScams.js
```diff
@@ -0,0 +1,33 @@
...
+    content={{
+      message: <Trans i18nKey="banners.ongoingScams" />,
+      Icon: TriangleWarning,
+      right: (
+        <Link id="modal-migrate-accounts-button" onClick={() => openURL(urls.banners.ongoingScams)}>
+          <Trans i18nKey="common.learnMore" />
+        </Link>
+      ),
+    }}
+    bannerId={"migrate"}
```
gre, line 30: wrong ids

gre (changes requested):
mmh check the ids. we seems to have "migrate" in the name. is it expected?

src/renderer/components/banners/OngoingScams.js
juan-cortes: No, it's an oversight on my part, good catch

juan-cortes:
> i'm ok with this, but it's probably the time to also have it auto fetched using the config.json. we'll rediscuss later. how does it work with the BCH banner?

There's a priority and only one is visible at a time, once you dismiss you see the next visible one. This was confirmed to be the expected behaviour with @cthepot-ledger granted it's not ideal since they might miss it but it's what we have.

juan-cortes:
Updated the screenshot to keep ci happy

nabil-brn (approved):
✓ QA-ed

---
Ledger Live 2.15.0-sha.680cdd0

PR #3270, opened 2020-11-03
LL-3816 (Analytics) add source for tracking page events

LFBarreto:
(Analytics): add optional source property from navigation routing to all track page events

All tracking page events can now send an optional source property.
This should happen everywhere we have an UI triggering a routing update since all history.push have been updated with that extra param.

### Type

Analytics
### Context

LL-3816

Using the console test out navigation from sidebar you should see the `source: "sidebar"` property on the event. same for other navigation.


gre:
HODL for discussion in LL. especially around the fact using the buttons event should be enough to identify user's behavior.

gre (dismissed):

Arnaud97234 (commented):
From sidebar:
- [x] Portfolio
- [x] Accounts
- [x] Buy
- [x] Swap
- [x] Compound
- [x] Manager
- [x] Experimental Feature

From asset allocation:
- [x] Portfolio asset allocation
- [x] Account

From accounts page:
- [x] Account
- [x] subAccount

From account page:
- [x] SubAccount
- [x] Buy
- [x] Swap

From breadcrumb:
- [x] Portfolio
- [x] Accounts
- [x] Account
- [x] SubAccount

From operation details:
- [x] Account
- [x] subAccount

From topBar:
- [x] Settings

Arnaud97234 (changes requested):
- Can't access to manager from add account modal (if app missing on the device)
- No event when accessing to Settings from Experimental flag in the sidebar

Arnaud97234 (approved):

PR #3274, opened 2020-11-04
fix spectron not running

machard:
Spectron was not runnning because of https://github.com/electron-userland/spectron/issues/174#issuecomment-319242097

### Type

bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-3276

yarn spectron (needs yarn start running in an other terminal)

### DIFF
--- a/src/main/window-lifecycle.js
+++ b/src/main/window-lifecycle.js
@@ -100,7 +100,7 @@ export async function createMainWindow({ dimensions, positions }: any, settings:
 
   loadWindow();
 
-  if (__DEV__ || DEV_TOOLS) {
+  if ((__DEV__ || DEV_TOOLS) && !process.env.DISABLE_DEV_TOOLS) {
     mainWindow.webContents.openDevTools();
   }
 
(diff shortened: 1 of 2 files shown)


machard:
Actually there is a problem when running on mac because of the screenshot size.
When running on a computer with a retina screen, the size of the screenshot is doubled.
I tried to resize capture on the fly to the size of the existing captures but it does not pass.
onboarding-1-get-started-diff

nabil-brn (approved):
> Actually there is a problem when running on mac because of the screenshot size.
> When running on a computer with a retina screen, the size of the screenshot is doubled.
> I tried to resize capture on the fly to the size of the existing captures but it does not pass.
> onboarding-1-get-started-diff

UI tests are meant to run on CI inside docker containers, that's why the resolution is not the same, but it's actually a good idea to standardize those screenshots to be able to develop tests faster and make the debugging easier.

✓ QA-ed

Running on windows and macOS

---
08644169 (HEAD -> fix-spectron-run, machard/fix-spectron-run) fix spectron not running

machard:
I am looking at https://github.com/SeleniumHQ/docker-selenium#standalone and https://www.electronjs.org/docs/tutorial/using-selenium-and-webdriver

otherwise it would need to run the tests locally in a docker environment as well

gre (approved):

PR #3277, opened 2020-11-05
live-common 15.7.x

gre:
- should fixes recent TRON issues
- check that some compound / erc20 txs still works (some refinement in gasLImit calculation)
- LL-3842
- use supplyMax when creating transaction in supply
- LL-3835 added CompoundLowerAllowanceOfActiveAccountError

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@ledgerhq/hw-transport-http": "^5.28.0",
     "@ledgerhq/hw-transport-node-hid-singleton": "^5.28.0",
     "@ledgerhq/ledger-core": "6.9.1",
-    "@ledgerhq/live-common": "^15.7.6",
+    "@ledgerhq/live-common": "^15.7.7",
     "@ledgerhq/logs": "^5.28.0",
     "@tippyjs/react": "^4.0.2",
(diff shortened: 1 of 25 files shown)


valpinkman (approved):

Arnaud97234 (changes requested):
### Bug found
Just after submitting votes the the voted ongoing period should be "Just started" instead of "common.timeAgo.seconds"

IMG_1041

Arnaud97234:
- [x] LL-3842: Already approved account should redirect to supply

Arnaud97234 (changes requested):
### Bug found
- Approve an account that is already Approved and Supplied
- Enter a specific amount (disable No limit)

The following error is displayed. As explained by @LFBarreto, in that case it shouldn't be possible to approve. The error raised is not understandable by the user.

Screenshot 2020-11-05 at 21 47 39

gre:
Luiz provided a better wording @Arnaud97234

Arnaud97234 (commented):
image

OK

Arnaud97234 (approved):
Some fixes are remaining but they are minor, we could fix them in another PR

PR #3280, opened 2020-11-06
try to run tests easily the same way locally and on the CI

machard:
I am trying to run electron in docker when using spectron in order to be able to run tests quickly the same way locally and on the CI.
(otherwise there is an issue with captures comparaisons).

electron and web driver run in docker. spectron on the host speaks to docker/electron and docker/electron speaks to webpack on the host. Pbm: node_modules is mounted in the container so the libs with native code break.

idea : copy package.json in app/package.json, mount only .webpack in app/.webpack, and do npm install in the dockerfile (tradeoff : run docker-compose up —build after a deps update but if only code or tests change you just run yarn spectron and it stills runs asap)

unknown: I just hope captures will work since the jest plugin that handles images lives in the host and electron in the container. but i think its handled. 

### Type

tests run improvements

### Context

Story spectron

yarn spectron


machard:
Youpi it's working. And it's quick to iterate now when writing tests.
I updated the reference screenshots since orginal size was 1023x767 which is weird.
But now, these screenshots will pass the tests locally for everyone.

Left to do:
- update the github webflow

Nice to have
- update spectron and electron and make a PR to spectron to avoid mokeypatching and overriding webdriverOptions like that.

machard:
it works!

4 tests to fix
It feels like getMockDeviceEvent hangs sometimes locally
Speed ups possible (https://github.com/marketplace/actions/docker-layer-caching on ci and change docker build context to .)

valpinkman (approved):

gre (approved):

PR #3282, opened 2020-11-06
Mock swap spectron

juan-cortes:
image

## Changes to the bullrun
This pr introduces two changes to the bullrun that should make local testing easier for devs and qa. 
- The ability to run the bullrun with disabled screenshots, meaning the run should be faster and not fail in those checks. To run without screenshots simply call `yarn spectron -no-screenshots`. This should have no impact whatsoever in the github actions run, which would still take screenshots by running without the `-no-screenshots` flag.
- The possibility to filter some currencies or steps by editing the `currencies` and `steps` variables respectively, this is particularly useful to test a specific flow, or develop it, by just running through the mvp steps instead of the full run. For instance for the swap I just needed ETH and BTC, and the onboarding, and add accounts steps, which means the run would complete without manager, send, receive, delegate, discreet, etc. You get the point.

## About the swap flow
We also now have the swap flow covered. It's just a bullish run, as it should be, so no error handling, no history even (this could be easily added by clicking on the tab I would assume). The swap generated is obviously not real since we don't want to be moving funds around and paying fees on every pr.

### Type

Feature

### DIFF
--- a/src/renderer/components/DebugMock.js
+++ b/src/renderer/components/DebugMock.js
@@ -6,6 +6,7 @@ import { getEnv } from "@ledgerhq/live-common/lib/env";
 import Text from "~/renderer/components/Text";
 import { ReplaySubject } from "rxjs";
 import { deserializeError } from "@ledgerhq/errors";
+import { fromTransactionRaw } from "@ledgerhq/live-common/lib/transaction";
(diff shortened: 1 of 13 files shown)


juan-cortes:
Since @machard seems to be working on a pretty involved rework of the bull run and this hasn't been merged in almost three weeks, I see no benefit in updating it. Please @machard extract the swap feature out of this branch and merge it into yours.

PR #3284, opened 2020-11-06
fix text in compound banner on manage modal

valpinkman:
### Type

Bug Fix

### Context

LL-3847


gre:
looks like it is included in https://github.com/LedgerHQ/ledger-live-desktop/pull/3287

PR #3287, opened 2020-11-10
fix/wordings

valpinkman:


gre (dismissed):

Arnaud97234 (changes requested):
On desktop the amount shown in the disclaimer corresponds to the deposited funds on the account while it should be the amount Approved.
The sentence is correct but the value is not: For instance the account below is fully approved and the value deposited = 6.001 DAI, On mobile the behaviour is ok.

image

IMG_440903775774-1

Arnaud97234 (approved):

Issue #1240, opened 2018-07-22
Support connecting to the user's own full node


durbanpoison:
- [x] I have checked this feature was not yet requested.

The way Ledger Live works today is that the wallet is getting its balances from a central server run by a third party. This is convenient because it means the user doesn't have to run their own full node for each coin they have in their wallet. But this convenience comes at the cost of some security and privacy guarantees that full nodes provide.

I would like to be able to connect my Ledger Live app to my own full nodes, either nodes running locally on my computer or on a remote server. These full nodes might only be running traffic through Tor, so the wallet application should also support connecting to these nodes through Tor.

#### Part of the application

Settings

attakorns:
Agreed! This would be a very useful feature!

kravens:
We need this, my node my rules, not my node not my rules!

2pac1:
This is a must.
Electrum is lightning years ahead of ledger live.
There should be an option to force the use of TOR + option to pin it to your NODE
Why would Ledger HQ keep logs of your IP address and what UTXOs are you checking!? Come on

Thasugeas:
Please allow this, even without TOR, it would be fine!

System-Glitch:
Any news on this?

althea16:
Isn't this already by possible by just using your Ledger with Wasabi or Electrum, for example, which are connected to your own node??

kravens:
Many users have indeed switched to Electrum or Wasabi to connect their own node to the hardware-wallet, including me.

MikeD10375:
This feature request has been open for years now with no movement on the issue. What is the problem? Is Ledger making money from this data? I would rather not leave Ledger products/services, however since we are being ignored I may have no choice.

Jacksper13:
Following up on this

MikeD10375:
I've had no success using Ledger Live Desktop on my Ubuntu 20.04.1 laptop. I've been using Ledger Live on my Android phone in the meanwhile.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

System-Glitch:
#3291 adds support for [satstack](https://github.com/ledgerhq/satstack). I think this addition will allow this issue to be closed.

PR #3291, opened 2020-11-12
LL-3081 Satstack feature

juan-cortes:
Fixes https://github.com/LedgerHQ/ledger-live-desktop/issues/1240

## Testing notes

- testing the app in dev mode do not test the feature correctly. It should be test in production mode. If you still want to do it in dev mode, you must run `lss` command on the terminal when being in the User Data folder where the lss.json has been written, otherwise SatStack won't find the dev's config.


src/renderer/modals/FullNode/FullNodeBody.js
```diff
@@ -0,0 +1,167 @@
...
+        }
+      }
+    }
+    prefill();
+  }, [satStackAlreadyConfigured]);
+
+  useEffect(() => {
+    if (scannedDescriptors) {
+      try {
+        saveLSS({ node: { ...nodeConfig }, accounts: scannedDescriptors });
```
gre, line 85: why do we `...nodeConfig` it ?

src/renderer/modals/FullNode/FullNodeBody.js
```diff
@@ -0,0 +1,167 @@
...
+        if (maybeNodeConfig && maybeNodeConfig.node) {
+          setNodeConfig(maybeNodeConfig.node);
+        }
+      }
+    }
+    prefill();
+  }, [satStackAlreadyConfigured]);
+
+  useEffect(() => {
+    if (scannedDescriptors) {
```
gre, line 83: the condition feels a bit weird.
this sounds like it could run if you edit nodeConfig and have scannedDescriptors settled. But i guess it works since scannedDescriptors changes as second step.
But then, what about the Edit mode? Should we save it when user just "fixes" its config but without needing to scan any new descriptors?

In term of flow, I have feeling that we have two places where we should save:
- after full config (node + descriptors) (for both first time and re-edit time)
- after node config connection succeed validation and **if it's edit mode** (🤔  even tho we could allow to save it even on the first run, but it would make the set up partially complete (missing accounts descriptor) so probably to just do it in edit mode)

src/renderer/modals/FullNode/FullNodeBody.js
```diff
@@ -0,0 +1,167 @@
...
+    const errors = validateRPCNodeConfig(nodeConfig);
+    // NB All fields are required, why do we need these errors?
+    return !errors.length;
+  }, [nodeConfig]);
+
+  useEffect(() => {
+    // Prefill the node data for edit?
+    async function prefill() {
+      if (satStackAlreadyConfigured) {
+        const maybeNodeConfig = await loadLSS();
```
gre, line 73: maybeLSSConfig, it's not just node config in this object.

src/renderer/modals/FullNode/helpers.js
```diff
@@ -0,0 +1,55 @@
...
+  stringifySatsStackConfig,
+  parseSatsStackConfig,
+} from "@ledgerhq/live-common/lib/families/bitcoin/satstack";
+import type { SatsStackConfig } from "@ledgerhq/live-common/lib/families/bitcoin/satstack";
+import path from "path";
+const lssFileName = "lss.json";
+
+export const saveLSS = async (lssConfig: SatsStackConfig) => {
+  const userDataDirectory = resolveUserDataDirectory();
+  const filePath = path.resolve(userDataDirectory, lssFileName);
```
gre, line 16: I think filePath should be on main thread side. There will never be a reason we want the config file to be somewhere else so I would move out this logic (lssFileName, userDataDirectory, filePath) to be in where the invoked functions are implemented.

src/renderer/init.js
```diff
@@ -29,6 +29,7 @@ import dbMiddleware from "~/renderer/middlewares/db";
 import createStore from "~/renderer/createStore";
 import events from "~/renderer/events";
 import { setAccounts } from "~/renderer/actions/accounts";
+import { loadLSS } from "~/renderer/modals/FullNode/helpers";
```
gre, line 32: I don't like the fact this loadLSS is in a hidden place like modals/FullNode/helpers. I would prefer we put it next to the other loaders, probably in renderer/storage.js itself.

src/renderer/modals/FullNode/helpers.js
```diff
@@ -0,0 +1,55 @@
...
+
+  if (filePath) {
+    const configStub = {
+      node: { host: "", username: "", password: "" },
+      accounts: [],
+    };
+    const maybeExistingConfig = (await loadLSS()) || configStub;
+    const updated = editSatsStackConfig(maybeExistingConfig || { accounts: [] }, lssConfig);
```
gre, line 24: I don't understand why we `|| { accounts: [] }` with the configStub logic just done before.

src/renderer/modals/FullNode/helpers.js
```diff
@@ -0,0 +1,55 @@
...
+    await ipcRenderer.invoke("generate-lss-config", filePath, stringifySatsStackConfig(updated));
```
gre, line 25: with the current logic, this could fail because you have silented the error case of writing the file.
either don't try {} catch {} on the other side OR do a condition here.

I tend to think we must not try catch on other side because you would need to display an error on the UI, right?

src/renderer/modals/FullNode/helpers.js
```diff
@@ -0,0 +1,55 @@
...
+    setEnvOnAllThreads("SATSTACK", true);
+  }
+};
+
+export const removeLSS = async () => {
+  const userDataDirectory = resolveUserDataDirectory();
+  const filePath = path.resolve(userDataDirectory, lssFileName);
+
+  if (filePath) {
+    await ipcRenderer.invoke("delete-lss-config", filePath);
```
gre, line 35: same. this might fail.

src/renderer/modals/FullNode/helpers.js
```diff
@@ -0,0 +1,55 @@
...
+
+export const loadLSS = async (): Promise<?SatsStackConfig> => {
+  const userDataDirectory = resolveUserDataDirectory();
+  const filePath = path.resolve(userDataDirectory, lssFileName);
+
+  if (filePath) {
+    const satStackConfigRaw = await ipcRenderer.invoke("load-lss-config", filePath);
+    try {
+      const config = parseSatsStackConfig(satStackConfigRaw);
+      setEnvOnAllThreads("SATSTACK", true);
```
gre, line 48: same.

src/renderer/modals/FullNode/helpers.js
```diff
@@ -0,0 +1,55 @@
...
+      return config;
+    } catch (e) {
+      // Bad data from the file for instance.
```
gre, line 51: well, are you sure this happens? because you try catch on other side. also we never should catch(e) without a log.

src/renderer/modals/FullNode/steps/StepAccounts.js
```diff
@@ -0,0 +1,133 @@
...
+const Accounts = () => {
+  // FIXME Not using the AccountList componenent because styles differ quite a bit, we should unify.
+  const accounts = useSelector(accountsSelector);
+  const [numberOfAccountsToScan, setNumberOfAccountsToScan] = useState(10);
+  const currency = getCryptoCurrencyById("bitcoin");
+  const bitcoinAccounts = accounts.filter(a => getAccountCurrency(a) === currency);
+
+  const onUpdateNumberOfAccountsToScan = useCallback(value => {
+    const newNumberOfAccounts = parseInt(value) || 1;
+    if (newNumberOfAccounts < 1000) {
```
gre, line 41: with suggestion above of maxlength, you would not need it, but if you want it safe, this logic should be part of the fallback logic.

src/renderer/modals/FullNode/steps/StepAccounts.js
```diff
@@ -0,0 +1,133 @@
...
+
+const Accounts = () => {
+  // FIXME Not using the AccountList componenent because styles differ quite a bit, we should unify.
+  const accounts = useSelector(accountsSelector);
+  const [numberOfAccountsToScan, setNumberOfAccountsToScan] = useState(10);
+  const currency = getCryptoCurrencyById("bitcoin");
+  const bitcoinAccounts = accounts.filter(a => getAccountCurrency(a) === currency);
+
+  const onUpdateNumberOfAccountsToScan = useCallback(value => {
+    const newNumberOfAccounts = parseInt(value) || 1;
```
gre, line 40: please give 10 to second parameter of parseInt.
I suggest to fallback to 1 as soon as the value is `isNaN()`, `!isFinite()` or `< 1`.
and also to actively deny an edit that would make it higher than 1000 as you seem to do before => maybe just use "maxlength 3" on the input?

src/renderer/modals/FullNode/steps/StepConnectDevice.js
```diff
@@ -0,0 +1,151 @@
...
+        deviceId: device.deviceId,
+        currency,
+        limit: 10,
+      })
+        // $FlowFixMe I don't know what you want.
+        .pipe(reduce((acc, item) => acc.concat({ descriptor: item }), []))
+        .subscribe(descriptors => {
+          setScanStatus(connectionStatus.SUCCESS);
+          setScannedDescriptors(descriptors);
+        });
```
gre, line 52: ⚠️ you must handle error case of the subscription which can happen for various reasons. Like sleep mode of device, quitting the app on the device, etc..

src/renderer/modals/FullNode/steps/StepConnectDevice.js
```diff
@@ -0,0 +1,151 @@
...
+  setError: Error => void,
+}) => {
+  const currency = getCryptoCurrencyById("bitcoin");
+  const [device, setDevice] = useState(null);
+  const [scanStatus, setScanStatus] = useState<ConnectionStatus>(connectionStatus.IDLE);
+
+  useEffect(() => {
+    if (device) {
+      const currency = getCryptoCurrencyById("bitcoin");
+      command("scanDescriptors")({
```
gre, line 42: ⚠️  you must keep a subscription param and unsubscribe when the effect is gone.

src/renderer/modals/FullNode/steps/StepConnectDevice.js
```diff
@@ -0,0 +1,151 @@
...
+        limit: 10,
+      })
+        // $FlowFixMe I don't know what you want.
+        .pipe(reduce((acc, item) => acc.concat({ descriptor: item }), []))
+        .subscribe(descriptors => {
+          setScanStatus(connectionStatus.SUCCESS);
+          setScannedDescriptors(descriptors);
+        });
+    }
+  }, [device, setScannedDescriptors]);
```
gre, line 54: would it be possible for `device` to actually mutate and trigger again the scan?

src/renderer/modals/FullNode/steps/StepConnectDevice.js
```diff
@@ -0,0 +1,151 @@
...
+  scannedDescriptors,
+}: {
+  onClose: () => void,
+  onStepChange: FullNodeSteps => void,
+  scannedDescriptors: any,
+}) => {
+  const onOpenUserDataFolder = useCallback(() => {
+    const userDataDirectory = resolveUserDataDirectory();
+    shell.showItemInFolder(userDataDirectory);
+  }, []);
```
gre, line 131: why don't we have an existing component that do exactly that? It seems we repeat ourself with the same button that exists in settings?

src/renderer/modals/FullNode/steps/StepDisconnect.js
```diff
@@ -0,0 +1,43 @@
...
+
+const Disconnect = () => (
+  <Box>
+    <Illustration />
+  </Box>
+);
+
+export const StepDisconnectFooter = ({ onClose }: { onClose: () => void }) => {
+  const onConfirmDisconnect = useCallback(() => {
+    removeLSS();
```
gre, line 27: - this is async
- this might fail

src/renderer/modals/FullNode/steps/StepSatStack.js
```diff
@@ -0,0 +1,189 @@
...
+import { urls } from "~/config/urls";
+
+const SatStack = ({
+  satStackDownloaded,
+  setSatStackDownloaded,
+}: {
+  satStackDownloaded: boolean,
+  setSatStackDownloaded: boolean => void,
+}) => {
+  const latestStatus = useObservable(statusObservable, { type: "satstack-disconnected" });
```
gre, line 24: mmh I don't think the initial state should be `{ type: "satstack-disconnected" }`
it should be actually null/undefined and in that case you would render a spinner because we don't actually know yet the status. (see #sat-stack discussion)

src/renderer/screens/settings/sections/Accounts/FullNodeStatus.js
```diff
@@ -0,0 +1,64 @@
...
+const naiveMapping = {
+  ko: ["satstack-disconnected", "node-disconnected", "invalid-chain"],
+  ok: ["ready"],
+  pending: ["syncing", "scanning", "initializing"],
+};
+
+const FullNodeStatus = () => {
+  const { t } = useTranslation();
+  const latestStatus = useObservable(statusObservable, { type: "satstack-disconnected" });
+  const satStackAlreadyConfigured = getEnv("SATSTACK");
```
gre, line 37: `useEnv`

src/renderer/modals/FullNode/steps/StepSatStack.js
```diff
@@ -0,0 +1,189 @@
...
+};
+
+export const StepSatStackFooter = ({
+  satStackDownloaded,
+  onClose,
+}: {
+  satStackDownloaded: boolean,
+  onClose: () => void,
+}) => {
+  const latestStatus = useObservable(statusObservable, { type: "satstack-disconnected" });
```
gre, line 174: same.

src/renderer/reducers/settings.js
```diff
@@ -383,6 +385,8 @@ export const swapProvidersSelector = (state: Object) => state.settings.swapProvi
 
 export const showClearCacheBannerSelector = (state: Object) => state.settings.showClearCacheBanner;
 
+export const fullNodeEnabledSelector = (state: Object) => state.settings.fullNodeEnabled;
```
gre, line 388: I don't understand why we need to track this boolean and where we would not be able to replace that usage by a `useEnv("SATSTACK")`

gre (dismissed):

src/internal/commands/scanDescriptors.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import type { Observable } from "rxjs";
+import { from } from "rxjs";
+import { scanDescriptors } from "@ledgerhq/live-common/lib/families/bitcoin/descriptor";
+import type { AccountDescriptor } from "@ledgerhq/live-common/lib/families/bitcoin/descriptor";
+import type { CryptoCurrency } from "@ledgerhq/live-common/lib/types";
+
+type Input = {
+  deviceId: string,
+  currency: CryptoCurrency,
```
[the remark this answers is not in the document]
juan-cortes: Sure, it's bitcoin anyway

src/renderer/modals/FullNode/FullNodeBody.js
```diff
@@ -0,0 +1,167 @@
...
+
+  const [numberOfAccountsToScan, setNumberOfAccountsToScan] = useState(10);
+  const [nodeConfig, setNodeConfig] = useState<RPCNodeConfig>({
+    host: "",
+    username: "",
+    password: "",
+    notls: false,
+  });
+  const [scannedDescriptors, setScannedDescriptors] = useState();
+  const patchNodeConfig = useCallback((patch: any) => setNodeConfig({ ...nodeConfig, ...patch }), [
```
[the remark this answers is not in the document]
juan-cortes: Seems like yes, it can

src/renderer/modals/FullNode/FullNodeBody.js
juan-cortes: Because it's wrong

src/renderer/modals/FullNode/FullNodeBody.js
juan-cortes: Fair enough

src/renderer/modals/FullNode/helpers.js
juan-cortes: Makes sense

src/renderer/init.js
juan-cortes: Moved them

src/renderer/modals/FullNode/helpers.js
juan-cortes: Me neither, probably trying to make flow happy at some point

src/renderer/modals/FullNode/helpers.js
juan-cortes: Not try/catch on the main thread anymore, all errors will be thrown, let's see how that goes

src/renderer/modals/FullNode/helpers.js
juan-cortes: removed all catches, errors will be thrown

src/renderer/modals/FullNode/steps/StepAccounts.js
juan-cortes: Removed it, added the , 10 to parse int

src/renderer/modals/FullNode/steps/StepConnectDevice.js
juan-cortes: good catch

src/renderer/modals/FullNode/steps/StepDisconnect.js
juan-cortes: So what do you want me to do about it?
If it fails to delete... then whelp, we tried 🤷

src/renderer/modals/FullNode/FullNodeBody.js
juan-cortes: it will save in both steps after initial setup and only after descriptor import on the initial setup

src/renderer/modals/FullNode/steps/StepConnectDevice.js
juan-cortes: We do, had to tweak it to be generic, but we do now.

src/renderer/modals/FullNode/steps/StepDisconnect.js
gre: yeah it's ok; we can keep it fire and forget

gre (commented):
all looks good to me.

- we'll need to wait wordings / url
- conflicts on yarn.lock

gre (dismissed):

onyb:
### Some test plans

This is in addition to testing the dedicated UX.

#### Synchronization

|      | balance ok | transaction history ok (+ amount, fees, addresses) | sum of ops ok | utxos ok |
|:--|:-------:|:-------:|:-------------:|:------:|
| **legacy** | ☐  | ☐ | ☐ | ☐ |
| **segwit** | ☐ | ☐ | ☐ | ☐ |
| **native segwit** | ☐ | ☐ | ☐ | ☐ |

#### Fee estimation

- [ ] ok

#### Transactions from one account to another within Ledger Live

- [ ] successful
- [ ] transaction visible under both destination and recipient accounts.
- [ ] confirmations updated

#### Transactions between different pairs of account types

|      | legacy | segwit | native segwit |
|:--|:-------:|:-------:|:-------------:|
| **legacy** | ☐  | ☐ | ☐ |
| **segwit** | ☐ | ☐ | ☐ |
| **native segwit** | ☐ | ☐ | ☐ |

#### Coin control

- [ ] UTXOs have confirmations.  ⚠️ known bug ⚠️

#### Bitcoind

- [ ] v0.20.0 ok
- [ ] v0.21.0 ok  (unreleased, but let's test anyway)

#### satstack

- [ ] can connect to bitcoind on the same machine
- [ ] can connect to bitcoind on a different machine
- [ ] works on Linux
- [ ] works on Mac OS
- [ ] works on Windows

gre:
@juan-cortes can you merge develop in? Thanks

juan-cortes:
> @juan-cortes can you merge develop in? Thanks

Develop merged in

gre (approved):

Arnaud97234 (changes requested):
@onyb @gre 

- Bitcoin 4 (Native Segwit): `xpub6CwXeXS3Y59ByJgRQGDMro3UESKvkBtdwB8EfNX6Z3jCSryVu51Em8PUaBpJcFBLgbWSv8jDpnBFWRbTpquDMYdcq1aAdzNFnPhHj88qrM6`
tx `20282d7a7cccac8225669e13718a00e8eba69367f52bd12314ac85fabe0a3469` is from changelly. The "From" address is missing. 

- Bitcoin 3 (Segwit): `xpub6DVf7AmjyPfAEpWjZGUXZf2YTADcK8Q69AfSyCFPg3duqnWChdj3R3es17KLN85HXGUDCP3gfu3DAXUY7bgUTMezSLz3eaq5dsQybYVkP8q`
tx '2c80b773f88ab907e986a9856bbb05716d26bdc91506e9d0f03d604c16eaab05' was made from an account that is not known by satstacks, the from address is present in the receive operation.

- On satstack side, All transactions made after I setup my personal node appear with a date "1970/01/01" (Tx made with satstack or regular LL explorer).
Pending tx status is "confirmed" with satstack while it shouldn't. Also the amount of confirmations is huge while it shouldn't.

Satstack:
image

Ledger explorer:
Screenshot from 2020-11-26 20-14-43

You can use the enclosed lss.json file to investigate:
[lss.json.zip](https://github.com/LedgerHQ/ledger-live-desktop/files/5604475/lss.json.zip)

gre:
Could it be a bug in satstack?

The missing from is weird but not that critical.
The timestamp issue is a bit more annoying, especially for the operation list and the balance history.

dasilvarosa:
The number of confirmations seems to be the block in which the transaction was included.

gre:
@juan-cortes can you investigate on possible reason of this?
- what is in the Account data that makes it glitch
- following the rabbit hole, what would be the specific HTTP call & data that would lead to such bad Account data (my intuition is one of the http call don't have correct info, and we can compare with Ledger's explorer)

gre:
ok actually not easy to test on our side unless we get your full set up. will take time to investigate, we will plan next week.

=> going to shift the settings to Experimental settings

- the settings must be in experimental. at the end of the list.
- once you have lss.json enabled, it also should show you the "Experimental used" bottom-left label so we are consistent with the other experimental.

@juan-cortes can you work on this? 👍

onyb:
**Missing `from` field:**

A full node can only know about your own addresses, therefore Changelly's address is empty. Whenever possible, satstack will try to infer it from whatever information it has.

I guess we need to include some UX fix here:
"if satstack enabled and no from address, display an info explaining why."

**Crazy number of confirmations:**

Looks like an integer overflow to me.

**Wrong timestamps:**

@Arnaud97234 Can you confirm if you see this only for unconfirmed transactions? What happens if you clear cache after the transaction has some confirmations?

gre:
@juan-cortes please merge develop in.

juan-cortes:
> @juan-cortes please merge develop in.

Monday boss, it's too late now

gre:
(mmh still conflicting, also lint issue)

static/i18n/en/app.json
```diff
@@ -245,6 +245,134 @@
...
+              "header": "Could not reach full node",
+              "description": "Please verify that your node is reachable and that the connection settings you entered are correct."
+            },
+            "invalid-chain": {
+              "header": "invalid-chain title",
+              "description": "invalid-chain description for tooltip"
+            },
+            "initializing": {
+              "header": "Loading",
+              "description": ""
```
gre, line 351: wording

juan-cortes:
> (mmh still conflicting, also lint issue)

Merged again, fixed lint issue

PR #3300, opened 2020-11-20
Fixes errors to be visible again in Sync state

gre:
**remaining:**

- [x] fixes triangle black issue
- [ ] add small test to cover this: it can run a specific app.json which have one account, and using MOCK=crash_123 it should show an error on the portfolio header as well as on the accounts screen.

## What is this?

We have a regression in which all errors in sync would no longer appear in Synchronization error but would be displayed as **🕐 Paused**, which is only supposed to be legit for bad cases or cases where you are doing Send flow and the sync is actively "paused".

## How to test

- You can cut your network
- You can use mock with `MOCK=crash_123`

NB: an error is **designed** to be hidden for the user until the algorithm consider the last successful sync was too old. So turning network off **will not instantly show the error**, this is the normal behavior. See https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/651460609/High+level+states+of+the+app+synchronized+sync+error+

### DIFF
--- a/src/renderer/bridge/BridgeSyncContext.js
+++ b/src/renderer/bridge/BridgeSyncContext.js
@@ -34,6 +34,7 @@ export const BridgeSyncProvider = ({ children }: { children: React$Node }) => {
       onUnusualInternalProcessError();
     }
     logger.critical(error);
+    return error;
   }, []);
 
   return (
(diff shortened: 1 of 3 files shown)


Arnaud97234 (commented):
Fixed:
- [x] https://ledgerhq.atlassian.net/browse/LL-3115

gre:
@Arnaud97234 arrow should be fixed and we can postpone the actual test to a future PR (cojoin with @machard 's test PR)

Arnaud97234 (approved):

PR #3304, opened 2020-11-20
LL-3521 Restore hide tokens setting section and add toggle

JunichiSugiura:
Screen Shot 2020-11-20 at 16 50 11

### Type
Bug Fix (sort of)

### Context
LL-3521

Settings -> Accounts -> Hide Tokens

### DIFF
--- a/src/renderer/screens/settings/sections/Accounts/BlacklistedTokens.js
+++ b/src/renderer/screens/settings/sections/Accounts/BlacklistedTokens.js
@@ -1,0 +1,132 @@
+// @flow
+import {
+  SettingsSection as Section,
+  SettingsSectionHeader as Header,
+} from "~/renderer/screens/settings/SettingsSection";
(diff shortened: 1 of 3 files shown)


gre (dismissed):

Arnaud97234 (changes requested):
### Bug found
Number of hidden tokens is incorrect, it shows the amount of Currencies concerned.
For instance in the screenshot below the it shows "2 tokens" while it should be 8.
image

Also I hid "HOT" token from 2 distinct Ethereum accounts, in the console there is the following warning:
`Warning: Encountered two children with the same key, `ethereum/erc20/hydro_protocol`. Keys should be unique`

By hiding "HOT" token from one Ethereum account should we by default hide all "HOT" subAccounts present in LL and display in the hidden list "HOT" only once ?

Arnaud97234 (approved):

PR #3310, opened 2020-11-23
live-common 17.1.0

gre:
Upgrade live-common that includes

- some fixes needed for #3291 (nothing to test here)
- all fiats are now supported


Arnaud97234 (changes requested):
- CLP is expressed with the regular dollar sign "$" -> Should be "CLP$"

gre:
@Arnaud97234 fixed

Arnaud97234 (dismissed):

PR #3317, opened 2020-11-24
Show time inside the tooltip on account charts when week range is selected

JunichiSugiura:
Time was missing in the tooltip just on Account page so I added.

### Type
Bug Fix

Account page -> hover on the chart to see tooltip. It should be the same to the one on Portfolio.

### DIFF
--- a/src/renderer/screens/account/BalanceSummary.js
+++ b/src/renderer/screens/account/BalanceSummary.js
@@ -75,10 +75,15 @@ export default function AccountBalanceSummary({
           <Box ff="Inter|Regular" color="palette.text.shade60" fontSize={3} mt={2}>
             {moment(d.date).format("LL")}
           </Box>
+          {range === "week" ? (
+            <Box ff="Inter|Regular" color="palette.text.shade60" fontSize={3}>
+              {moment(d.date).format("LT")}
+            </Box>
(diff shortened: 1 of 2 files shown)


gre (approved):

nabil-brn (changes requested):
It's back 👍 but why `range === "week" ?`, it's not the case on portfolio, the time is on all charts.
cc @cthepot-ledger

---
Windows
4533c7e5b (grafted, HEAD -> hotfix/account-chart, ledger/hotfix/account-chart) show time on account charts

gre:
@JunichiSugiura we should prefer ask live-common the granularity, because in future we'll add more like 1D
https://github.com/LedgerHQ/ledger-live-common/blob/master/src/portfolio/range.js#L46

JunichiSugiura:
It should always show time no mater what granularity/range is selected now. Although it no longer necessary, I'll keep in mind about `granularityId` property for feature tasks.

gre (approved):

nabil-brn (approved):
The time is back on all charts in account section.

---
Windows
157de039 (HEAD -> hotfix/account-chart, ledger/hotfix/account-chart) display time no matter what granualarity/range is selected

PR #3325, opened 2020-11-25
LL-3980 reorder banners

juan-cortes:


gre (approved):

Arnaud97234 (commented):
- Update available -> Phishing banner not displayed
- After updated, when the app restart -> Phishing banner displayed

Only comment: User is not aware of the phishing campaign as long as he don't update LL.
If we are OK with this behaviour then it's a Go.

gre:
we are ok with this behavior because it is a temporary banner and is already planned in next sprint to drop it.

PR #3327, opened 2020-11-25
Fix regression in Price component

juan-cortes:
During the recent rework for countervalues we lost the ability to provide a fixed rate for the `Price` component, a feature used by Swap since we need to provide the rate from Changelly or the appropriate provider.

### Type

Bug fix

### Context

Slack

- Enter swap
- Get a rate
- Confirm we get a conversion under the To currency 

image

### DIFF
--- a/src/renderer/components/Price.js
+++ b/src/renderer/components/Price.js
@@ -46,6 +46,7 @@ export default function Price({
   iconSize,
   showAllDigits = true,
   withIcon = true,
+  rate,
 }: Props) {
   const effectiveUnit = unit || from.units[0];
   const valueNum = 10 ** effectiveUnit.magnitude;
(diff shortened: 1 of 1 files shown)


src/renderer/components/Price.js
```diff
@@ -57,8 +58,12 @@ export default function Price({
     value: valueNum,
     disableRounding: true,
   });
-  const counterValue =
-    typeof rawCounterValue !== "undefined" ? BigNumber(rawCounterValue) : rawCounterValue;
+
+  const counterValue = rate
+    ? rate.times(valueNum) // NB Allow to override the rate for swap
```
gre, line 63: Should it do .integerValue() at the end? Because I'm not sure that the formatter will 🤔

gre (approved):

src/renderer/components/Price.js
juan-cortes: What do you mean? I only added the `rate` part but further down the code still uses `counterValue` as a BigNumber and is expected as such in `CurrencyUnitValue`

JunichiSugiura (approved):

nabil-brn (approved):
Reproduced on a previous version and fixed on this one.

image


---
Windows
a62fe73d (HEAD -> pr/juan-cortes/3327, juan/fix/price-swap) Fix regression in Price component

PR #3331, opened 2020-11-27
LL-3910 fix wording files

machard:
https://ledgerhq.atlassian.net/browse/LL-3910

cf
https://github.com/LedgerHQ/ledger-live-mobile/pull/1484
https://github.com/LedgerHQ/ledger-live-common/pull/967

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -872,7 +872,7 @@
   "DeviceAction": {
     "allowAppPermission": "Open the {{wording}} app on your device",
     "allowAppPermissionSubtitleToken": "to manage your {{token}} tokens",
-    "allowManagerPermission": "Allow Ledger Manager on your device",
+    "allowManagerPermission": "Allow {{wording}} on your device",
     "loading": "Loading...",
     "connectAndUnlockDevice": "Connect and unlock your device",
     "unlockDevice": "Unlock your device",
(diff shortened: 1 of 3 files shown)


static/i18n/en/app.json
```diff
@@ -867,7 +872,7 @@
   "DeviceAction": {
     "allowAppPermission": "Open the {{wording}} app on your device",
     "allowAppPermissionSubtitleToken": "to manage your {{token}} tokens",
-    "allowManagerPermission": "{{wording}} on your device",
+    "allowManagerPermission": "Allow Ledger Manager on your device",
```
gre, line 875: we never should have "Ledger Manager" in the wording, the goal is always to have "static" wording that are on device in live-common which we did here. this is to avoid having "Ledger Manager" in translation as it's not to be translated.

I think chris want to translate just the "Allow" part, so I think we need to introduce a new "wording".

gre (changes requested):

static/i18n/en/app.json
machard: ok. Then this one needs to go as well, https://github.com/LedgerHQ/ledger-live-common/pull/967

gre:
HODL for QA until we bump a new live-common.

gre:
it landed on develop via #3347

PR #3334, opened 2020-11-27
LL-3900 if no device currently use lastDeviceSeen in events context

machard:
https://ledgerhq.atlassian.net/browse/LL-3900

### DIFF
--- a/src/renderer/analytics/segment.js
+++ b/src/renderer/analytics/segment.js
@@ -12,7 +12,6 @@ import {
   shareAnalyticsSelector,
   lastSeenDeviceSelector,
 } from "~/renderer/reducers/settings";
-import { getCurrentDevice } from "~/renderer/reducers/devices";
 import type { State } from "~/renderer/reducers";
 
 // load analytics
@@ -49,6 +48,7 @@ const extraProperties = store => {
         appLength: device.apps.length,
       }
     : {};
+
   const sidebarCollapsed = sidebarCollapsedSelector(state);
 
   return {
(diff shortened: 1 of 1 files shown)


src/renderer/analytics/segment.js
```diff
@@ -40,7 +41,7 @@ const extraProperties = store => {
   const state: State = store.getState();
   const { language, region } = langAndRegionSelector(state);
   const systemLocale = getSystemLocale();
-  const device = getCurrentDevice(state);
+  const device = getCurrentDevice(state) || lastSeenDeviceSelector(state);
   const deviceInfo = device ? { modelId: device.modelId } : {};
```
gre, line 45: I see two issues in that:

- getCurrentDevice returns a Device but lastSeenDeviceSelector returns a DeviceModelInfo
- it doesn't solve LL-3900: we want to add in the "deviceInfo" object that I think comes from DeviceModelInfo

My proposal is to actually drop the "getCurrentDevice" usage. in the lastSeenDeviceSelector, we have enough we need and we can just use that. 

https://github.com/LedgerHQ/ledger-live-common/blob/5fd037de618d245e960046e69aab240c847cc7d2/src/types/manager.js#L31-L35

i would suggest that we track `{ modelId, deviceVersion, appLength }` (deviceVersion is deviceInfo.version, appLength is apps.length )

kinda consistently with https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/screens/manager/Dashboard.js#L73-L76

gre (changes requested):

gre (dismissed):

Arnaud97234 (approved):

PR #3341, opened 2020-11-30
LL-2846 add warning box when utxo > 50

machard:
https://ledgerhq.atlassian.net/browse/LL-2846


src/renderer/modals/Send/steps/StepSummary.js
```diff
@@ -64,13 +65,19 @@ export default class StepSummary extends PureComponent<StepProps> {
     const feesUnit = getAccountUnit(mainAccount);
     const feesCurrency = getAccountCurrency(mainAccount);
     const unit = getAccountUnit(account);
+    const utxoLag = mainAccount.operationsCount >= 50;
```
gre, line 68: we should warn when the transaction uses a lot of utxo. not when the accounts have a lot of utxos.

TransactionStatus have a txInputs array. that you can use.

probably `status.txInputs ? status.txInputs.length >= WARN_FROM_UTXO_COUNT : false`

the tx will be slow where there are a lot of utxo inputs to create the transaction.

(also, it is good if you can make this a variable)

gre (changes requested):

gre (dismissed):

Arnaud97234 (approved):

PR #3350, opened 2020-12-01
use recommendation from github security labs

valpinkman:


gre (approved):

valpinkman:
@gre @machard 

I've changed the `pull_request` to `pull_request_target` event as we do not use secrets (the github token is created and used only during workflow execution so there should be no secrets leaking

https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret
https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

PR #3364, opened 2020-12-06
LL-4003 ProductTour - Initial version

juan-cortes:
Initial version of the product tour, with some pending issues listed below:
- [ ] Navigation should be controller further, preventing the user from accessing flows that are not related.
- [ ] Integrated with the new onboarding, this setup should only be accessible for new users.
- [ ] Persist the product tour context between launches, right now it's reset on a new launch.
- [ ] Add a way for dev/qa to mark a flow as completed or reset the version

image
image
image
image
image

### Type

Feature
### Context

https://ledgerhq.atlassiaan.net/browse/LL-4003


juan-cortes:
Screenshots are not representative of the current state of the PR since we went back to the original design which accommodates more cards than the three shown above.

cthepot-ledger:
@juan-cortes no artefacts

cthepot-ledger:
@juan-cortes no artefacts... rerunning jobs does not product them

juan-cortes:
Closing this because github doesn't want to act

PR #3366, opened 2020-12-08
LL-4018 FullNode / Satstack Bugs

juan-cortes:
Bug fixes and polishes from https://ledgerhq.atlassian.net/browse/LL-4018

- LL-4084 Prevent backdrop click on full node modal
- LL-4022 Wording fixes
- LL-4021 UX improvement of satstack step
- LL-4005 Proper handling of the number of scanned descriptors to look for (cc @gre this addresses what you noticed, it was always going for 10, regardless of the user input)
- LL-4006/LL-4024 Attempt to fix scrolling on the account step

PR #3368, opened 2020-12-08
LL-4016 Move full node to accounts settings screen

juan-cortes:
Pending the work on https://github.com/LedgerHQ/ledger-live-desktop/pull/3366 this pr will move the block back to the accounts screen. Made a separate pr in case we want further releases while still keeping it experimental.


valpinkman (approved):

gre:
Let's reopen this PR when we are ready to enable Full Node in non-experimental.

PR #3373, opened 2020-12-10
live-common 17.4.0

gre:
### Type

upgrade

## Context

mashup of fixes needed in the sprint

## Parts of the app affected / Test plan

- upgrading some small lib as part of this. (patches for openpgp, ws, uuid). It should be safe because they are just patches.
- VES fiat has replaced VEF
  - chosing VES before testing this PR will not break the app. but the fiat might fallback to USD.
- LL-3908 less chance to get negative ROI on compound (still might happen but we lowered the risk – no guarantee i solved it but it’s good enough for now)
- currencies like Cosmos/Tron/Tezos which implement preload still work correctly regarding loading the delegator lists.
- ripple-lib upgrade: check send and receive works ok. It also includes the fix we did regarding websocket.
- new ERC20 update (not much to test, signatures are done so just check one erc20 i guess) https://github.com/LedgerHQ/ledgerjs/commit/281ffde5ae162ff7c58bdc1f0705ecf61abe5417
- fetch countervalues at least for a year to workaround bug when you have accounts that no longer fetch old ops

### Many SatStack polishes

(closes https://github.com/LedgerHQ/ledger-live-desktop/pull/3366)

- [x] LL-4001 SatStack version will be checked and should be recent enough (currently >=0.11.1) – a SatStackVersionTooOld error will appear everywhere if it's not the case
- [ ] LL-3998/LL-4002 adding accounts with a SatStack that wouldn't know about them yet would throw a SatStackDescriptorNotImported
- [x] LL-4084 Prevent backdrop click on full node modal
- [x] LL-4022 Wording fixes
- [ ] LL-4021 UX improvement of satstack step
- [x] LL-4005 Proper handling of the number of scanned descriptors to look for
- [ ] LL-4006/LL-4024 Attempt to fix scrolling on the account step


gre:
@machard we need to make this bot comment more clear. We don't have a instant feedback if it's a ✅  or a ❌ 

In case of a ✅ I also think we should just have a one line comment, without the details

gre:
I need to fix some stuff

machard:
@gre with the check and cross, do you still prefer to have the image diff section always populated ? (the reason I initially chose to have it expandable was to avoid polluting too much the comments)

gre:
I think it's ok 👍

Arnaud97234 (changes requested):
- LL-3998/LL-4002 adding accounts with a SatStack that wouldn't know about them yet would throw a SatStackDescriptorNotImported

See my comment on https://ledgerhq.atlassian.net/browse/LL-3998

- LL-4021 UX improvement of satstack step
See my comment on https://ledgerhq.atlassian.net/browse/LL-4021

gre:
- https://ledgerhq.atlassian.net/browse/LL-3998 is fixed 👍 
- https://ledgerhq.atlassian.net/browse/LL-4021 remaining improvement postponed to later 🙏

Arnaud97234 (approved):
There are still some bugs in satstacks (Add account from a 2nd device...), But the rest is OK.

PR #3381, opened 2020-12-14
LL-3997 hard reset 

machard:
https://ledgerhq.atlassian.net/browse/LL-3997

### DIFF
--- a/src/renderer/screens/settings/sections/Help/ResetButton.js
+++ b/src/renderer/screens/settings/sections/Help/ResetButton.js
@@ -3,19 +3,16 @@
 import React, { useCallback } from "react";
 import styled from "styled-components";
 import { useTranslation } from "react-i18next";
-import { useGlobalSyncState, SyncSkipUnderPriority } from "@ledgerhq/live-common/lib/bridge/react";
-import { useCountervaluesPolling } from "@ledgerhq/live-common/lib/countervalues/react";
 import logger from "~/logger";
 import { useHardReset } from "~/renderer/reset";
 import ConfirmModal from "~/renderer/modals/ConfirmModal";
 import ResetFallbackModal from "~/renderer/modals/ResetFallbackModal";
+import { SyncSkipUnderPriority } from "@ledgerhq/live-common/lib/bridge/react";
 import Box from "~/renderer/components/Box";
 import Button from "~/renderer/components/Button";
 import IconTriangleWarning from "~/renderer/icons/TriangleWarning";
 import type { ThemedComponent } from "~/renderer/styles/StyleProvider";
 import { useActionModal } from "./logic";
-
-let reset;
 
 export default function ResetButton() {
   const { t } = useTranslation();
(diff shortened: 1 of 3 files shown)


src/renderer/screens/settings/sections/Help/ResetButton.js
```diff
@@ -3,17 +3,20 @@
...
 import ConfirmModal from "~/renderer/modals/ConfirmModal";
 import ResetFallbackModal from "~/renderer/modals/ResetFallbackModal";
-import { SyncSkipUnderPriority } from "@ledgerhq/live-common/lib/bridge/react";
 import Box from "~/renderer/components/Box";
 import Button from "~/renderer/components/Button";
 import IconTriangleWarning from "~/renderer/icons/TriangleWarning";
 import type { ThemedComponent } from "~/renderer/styles/StyleProvider";
 import { useActionModal } from "./logic";
 
+let reset;
```
gre, line 18: i'm not really sure the global variable pattern is a good idea. This button could be used at different places too.

src/renderer/screens/settings/sections/Help/ResetButton.js
```diff
@@ -22,17 +25,28 @@ export default function ResetButton() {
...
 
-  const onConfirm = useCallback(async () => {
-    if (pending) return;
+  const globalSyncState = useGlobalSyncState();
+  const cvPolling = useCountervaluesPolling();
+  const isPending = cvPolling.pending || globalSyncState.pending;
+
+  reset = async () => {
+    if (isPending) {
+      return setTimeout(reset, 1000);
```
gre, line 34: setTimeout do not return a promise so i'm not sure what is going on here.

wouldn't it be simpler to have a "isPending()" predicate that you call in a

```
while(isPending()) {
  await delay(1000);
}
```

gre (changes requested):
i'm not fan of the current implementation, but also I don't think it's a good idea to implement it this way.

a sync and a fetch countervalues can possibly takes A LOT of time. to me the reset action should be near instant, but should cancel all ongoing actions that are happening. (like kill the thread, stop from receiving any REDUX events so we don't "race condition" on the "save again" situation)

src/renderer/screens/settings/sections/Help/ResetButton.js
machard: yes indeed i should have put it in a ref

src/renderer/screens/settings/sections/Help/ResetButton.js
machard: i am not retuning a promise, i am just waiting to recall the function.
indeed i did not think to do it this way. (need to be careful with the isPending to be sure it uses an updated version of isPending in the logic though).

machard:
(i removed the wiping of the countervalues, but I assume since we clean everything at the beginning, the countervalue will be initiated with the initialState)

gre (approved):

nabil-brn (approved):
Tested with 70 accounts _app.json_.
`,"accounts":[]}}`

---
Windows
fd8058e23 (HEAD -> pr/machard/3381, machard/LL-3997) go the other way around

PR #3384, opened 2020-12-15
LL-3897 closed loans sorting

machard:
https://ledgerhq.atlassian.net/browse/LL-3897

### DIFF
--- a/src/renderer/screens/lend/Closed/ClosedLoans.js
+++ b/src/renderer/screens/lend/Closed/ClosedLoans.js
@@ -170,9 +170,11 @@ const ClosedLoans = ({ loans }: Props) => {
           {t("lend.headers.closed.date")}
         </Text>
       </Header>
-      {loans.map((l, i) => (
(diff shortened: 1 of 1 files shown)


src/renderer/screens/lend/Closed/ClosedLoans.js
```diff
@@ -170,9 +170,11 @@ const ClosedLoans = ({ loans }: Props) => {
           {t("lend.headers.closed.date")}
         </Text>
       </Header>
-      {loans.map((l, i) => (
-        <Row key={`${l.endDate.toDateString()}${i}`} loan={l} />
-      ))}
+      {loans
+        .sort((l1, l2) => (l1.endDate <= l2.endDate ? 1 : -1))
```
gre, line 174: I think we instead should fix it in live-common because this is both shared by LLM and LLD and this is business logic to me

gre (changes requested):

machard:
https://github.com/LedgerHQ/ledger-live-common/pull/987

PR #3392, opened 2020-12-16
Hide onboarding lang picker

gre:
we didn't get the langs at time so we decided to disable the Onboarding lang selection.

The "stable" langs are back to be just EN and FR in the lang selection of settings.


IAmMorrow (approved):

valpinkman (approved):

gre:
i merged a bit too fast. let's confirm it works on develop @nabil-brn

PR #3398, opened 2020-12-18
LL-3979 Add UI testing for the app update banner states + QA tooling

juan-cortes:
This introduces test coverage of the app update UI. 
There are still some fails for the onboarding spectron tests but I believe @machard is working on them.
image

### Type

UI Testing, QA tooling

This pr introduces a couple of helpful flags for the QA/Devs, you can now debug Theme changes easily (like we did for mobile) by adding the `DEBUG_THEME=1` flag on startup as well as the `DEBUG_UPDATE=1` which is now reworked to match the mocked event emitter ui, and wrapped in the bottom left corner.

Screenshot 2020-12-18 at 14 36 49
Screenshot 2020-12-18 at 14 35 00
Screenshot 2020-12-18 at 14 34 02
Screenshot 2020-12-18 at 14 32 59


src/renderer/Default.js
```diff
@@ -94,6 +92,11 @@ export default function Default() {
 
               <LibcoreBusyIndicator />
               <DeviceBusyIndicator />
+              <DebugWrapper>
```
machard, line 95: I think I put the DebugMock outside of OnboardingElse to ensure that the aimations are disabled on the onboard

machard: it might be the reason why the onboarding screenshots fail

machard: actually when running in spectron, there is global class spectron-run added to the body, so maybe the global styles defined in DebugMock can be but directly in the global.css under .spectron_run { }

juan-cortes: I see, we could move the whole thing out then since it's absolutely positioned

juan-cortes: fixed.

machard:
the computer missing on onboarding_start was discussed in slack. it seems you've updated the screenshot but I am not sure the problem is solved in the code (theme not being set when launching fresh). I think hugo will fix it so maybe you can remove the updated snapshot

juan-cortes:
> the computer missing on onboarding_start was discussed in slack. it seems you've updated the screenshot but I am not sure the problem is solved in the code (theme not being set when launching fresh). I think hugo will fix it so maybe you can remove the updated snapshot. (actually i think he's even going to update the computer to an other illu)

Indeed, I'll wait until I can run spectron without errors on develop again to have this one merged, no rush really

gre:
@juan-cortes  can you check why CI fails?

juan-cortes:
> @juan-cortes can you check why CI fails?

I can, and I did, if all goes well, ci will not fail with the latest commit 🧘

juan-cortes:
wtf...

juan-cortes:
This includes some refactoring of the onboarding use of the Lottie component directly instead of the wrapped Animation one, it was breaking the spectron screen shots testing.

Until we find a better solution Lottie animations are paused on the first frame and don't play. I don't like this because we'll, we miss the UI, but it's something that we've agreed on already as the less bad solution for now.

Anyway, let's get this one out of the way please @gre

machard:
@juan-cortes did you deactivate as well the animation on the filling bars ? 
https://github.com/LedgerHQ/ledger-live-desktop/pull/3398/files?short_path=633883a#diff-633883a6512fae60c82692a98f4f595385a21f8784b22c4255e8cd765d667710

machard (dismissed):

gre:
@nabil-brn if you can do a quick check that onboarding assets are still ok (on one platform)

gre (approved):

gre:
@juan-cortes can you fix conflict?

juan-cortes:
> @juan-cortes can you fix conflict?

I'd like to tell you I can fix it @gre but it's up to CI at this point, and the longer I have to keep rebasing develop here and rebuilding the screenshots the worse my mental health will become.

juan-cortes:
Bullshit

machard:
I remember now, its specific to the wave thing ^^
https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/tests/specs/onboarding.spec.js#L149

gre (approved):

PR #3401, opened 2020-12-21
LL-4091 use spendableBalance in swap origin

machard:
https://ledgerhq.atlassian.net/browse/LL-4091

### DIFF
--- a/src/renderer/components/PerCurrencySelectAccount/Option.js
+++ b/src/renderer/components/PerCurrencySelectAccount/Option.js
@@ -71,6 +71,11 @@ function AccountRow({
 }) {
   const palette = useTheme("colors.palette");
 
+  const balance =
+    account.type !== "ChildAccount" && account.spendableBalance
+      ? account.spendableBalance
+      : account.balance;
+
   return (
     <AccountRowContainer nested={nested} disabled={disabled}>
       <Left>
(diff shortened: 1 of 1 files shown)


juan-cortes (dismissed):
We should always use the spendable balance, like we do for the send/receive flows. This whole concept of the "balanceKey" is not necessary and is prone to errors since accounts can be of different types (`TokenAccount`, `ChildAccount` and `Account`) and perhaps not share the same properties.

```const balance =
    account.type !== "ChildAccount" && account.spendableBalance
      ? account.spendableBalance
      : account.balance;
```

As used on the https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/components/SelectAccount.js#L94-L97 `SelectAccount` component would be more than enough

machard:
@juan-cortes you're saying I should add the snippet you posted in `PerCurrencySelectAccount/Option.js` and that it will be fine everywhere ?
i'll edit that way then, thx

juan-cortes (approved):
Lgtm

Arnaud97234 (approved):

src/renderer/components/PerCurrencySelectAccount/Option.js
```diff
@@ -71,6 +71,11 @@ function AccountRow({
 }) {
   const palette = useTheme("colors.palette");
 
+  const balance =
+    account.type !== "ChildAccount" && account.spendableBalance
+      ? account.spendableBalance
+      : account.balance;
```
gre, line 77: Can use this https://github.com/LedgerHQ/ledger-live-common/blob/master/src/account/helpers.js#L65

gre: But not blocking

PR #3409, opened 2020-12-23
Add token account flow to the Account test suite

Arnaud97234:
add subAccount (ERC2O) flow:
- add sub account while account from the parent Currency is missing -> Should redirect to add ETH account
- add account from the parent currency (Ethereum)
- add sub account while account from the parent Currency is present already  -> Should redirect to receive flow for ETH account

### Type

UI test

### Context

Add account


tests/common.js
```diff
@@ -73,6 +73,8 @@ export default function initialize(name, { userData, env = {}, disableStartSnap
       await app.client.pause(1000);
       expect(await app.client.screenshot()).toMatchImageSnapshot({
         customSnapshotIdentifier: `__start__${name}`,
+        failureThreshold: 0.01,
+        failureThresholdType: "percent",
```
machard, line 77: not a good idea, we need perfect screenshot.
If only 2 pixel are failing its ok to use failureThresholdType pixel

machard:
Looks like you've updated many screenshots.
When you run tests locally on develop, the tests don't pass ?

tests/specs/account.spec.js
```diff
@@ -25,14 +26,50 @@ describe("Account", () => {
...
-        await elemImportAddButton.waitForDisplayed();
-        await elemImportAddButton.waitForEnabled();
-        await elemImportAddButton.click();
-        await modalPage.close();
+        if (currency === "chainlink") {
+          // Add token when parent missing
+          const elemAddParentButton = await $("#modal-token-continue-button");
+          await elemAddParentButton.waitForDisplayed();
+          expect(await app.client.screenshot()).toMatchImageSnapshot({
+            customSnapshotIdentifier: "addTokenWithoutParent",
```
machard, line 34: we should have only one screenshot test per tests and the screenshot test should be at the end of the test so that if it fails it does not break what is following

PR #3413, opened 2020-12-24
Add subAccount flows UI test

Arnaud97234:
Add Receive token flows to the Account test suite

Add subAccount (ERC2O):
- add sub account while account from the parent Currency is missing -> Should redirect to add ETH account
- add account from the parent currency (Ethereum)
- receive token while corresponding subAccount already exist  -> Should lead to subAccount receive flow
- receive new token -> Should lead to Ethereum account selection

### Type

UI test

### Context

Add account


valpinkman (approved):

gre:
reason for CI to fail? shall we merge develop in? 🤔

PR #3424, opened 2020-12-29
LL-4261 Onboarding welcome screen fixes

juan-cortes:
Brings back illustration
Fixes colors for the welcome screen when in dark mode
Offer use a way out of the onboarding only if it's relaunched


src/renderer/components/Onboarding/Screens/Welcome/index.js
```diff
@@ -21,6 +25,7 @@ const WelcomeContainer: ThemedComponent<*> = styled.div`
   align-items: center;
   flex-direction: column;
   position: relative;
+  background: ${p => (p.dark ? "#1E2A3B" : p.theme.colors.palette.background.paper)};
```
valpinkman, line 28: this should not happen. Why is not in our color palette ?

src/renderer/components/Onboarding/Screens/Welcome/index.js
```diff
@@ -48,28 +53,41 @@ type Props = {
...
+        {null /* LL-4236 */ && <LangSwitcher />}
+        {onboardingRelaunched ? (
+          <Button small onClick={handleQuit}>
+            {t("common.cancel")}
+          </Button>
+        ) : null}
+      </TopRightContainer>
       <WaveContainer>
-        <AnimatedWave height={600} color={theme === "dark" ? "#587ED4" : "#4385F016"} />
+        <AnimatedWave height={600} color={useDarkColorPalette ? "#23324A" : "#4385F016"} />
```
valpinkman, line 83: i guess same here ?

src/renderer/components/Onboarding/Screens/Welcome/index.js
juan-cortes: You are right, moved the wave colors to the palette

PR #3435, opened 2021-01-05
confirmationsNb min value should be 1

henri-ly:
### Type

Bug fix

### Context

During the implementation of Polkadot, pending operation have a status as confirmed after the send flow (view details) instead of not confirmed
It appeared that a min value of the confirmationsNb is set to 0 by default instead of 1.

It was set like this by default but you can find the value was set as 0 in the pull request
`      confirmationsNb = { min: 1, def, max: 3 * def };`

Well, it has a huge impact on every coins, everywhere where the status are set as confirmed or not.


henri-ly:
Nvm, we can just change the blockAvgTime. After discussing with @rjarasson-ledger we probably gonna had a condition to check if the operation.blockHeight exist tho in an another PR.

PR #3440, opened 2021-01-05
LL-4278 build prod version for linux on ci

valpinkman:
Create a production build of Ledger Live on the CI on the lowest ubuntu to make sure new iterations will be backward compatible. Also libcore is built on ubuntu 16.04 so we keep it consistent

### Type

Chore

### Context

LL-4278


.github/workflows/build-linux-prod.yml
```diff
@@ -0,0 +1,45 @@
+name: linux production build
+on:
+  push:
+    tags:
+      - "v*"
```
gre, line 5: do you think it is possible to also constraints on master? or it makes no sense? thanks

machard (approved):

machard:
ha there is a logic bug in what i did for the updated screenshot. not sure why

machard (approved):

PR #3445, opened 2021-01-06
LL-4021 Delay first disconnect event for 10 seconds

juan-cortes:
Add an infobox while not connected, the ignore part is done https://github.com/LedgerHQ/ledger-live-common/pull/1005 there

### Type

Not sure

### Context

https://ledgerhq.atlassian.net/browse/LL-4021

I would suggest we merge all other prs before this since develop seems to be quite outdated at this point.

### DIFF
--- a/src/renderer/modals/FullNode/steps/StepSatStack.js
+++ b/src/renderer/modals/FullNode/steps/StepSatStack.js
@@ -76,6 +76,15 @@ const SatStack = ({ satStackDownloaded }: { satStackDownloaded: boolean }) => {
       >
         <Trans i18nKey={`fullNode.modal.steps.satstack.connectionSteps.${type}.description`} />
       </Text>
+      {type !== "ready" ? (
+        <Box mt={3}>
+          <InfoBox type="secondary" onLearnMore={onLearnMore}>
+            <Text ff="Inter|Regular" fontSize={3} color="palette.text.shade50">
+              <Trans i18nKey="fullNode.modal.steps.satstack.connectionSteps.notConnected.disclaimer" />
+            </Text>
+          </InfoBox>
+        </Box>
+      ) : null}
     </Box>
   );
 };


src/renderer/hooks/useSatStackStatus.js
```diff
@@ -3,10 +3,20 @@ import { useMemo } from "react";
...
 const useSatStackStatus = (): ?SatStackStatus => {
-  const observable = useMemo(() => command("getSatStackStatus")(), []);
+  const observable = useMemo(
+    () =>
+      command("getSatStackStatus")().pipe(
+        filter(({ type }) => type !== "satstack-disconnected"),
+        timeoutWith(INITIAL_DISCONNECT_TIMEOUT, of({ type: "satstack-disconnected" })),
+      ),
+    [],
+  );
```
gre, line 19: I'm not sure to understand what this fixes, typically to me it doesn't fix https://ledgerhq.atlassian.net/browse/LL-4021 but the other code does (the code changed in StepSatStack)

To me we really don't want to timeout, specifically not if you watch the Settings tab, it should runs forever, not stops after 10s 🤔

juan-cortes: unless I did it wrong, the command is triggered every second, what this does is to delay showing the disconnect one until after 10 seconds, any other status will work 🤷. Without this it will show the disconnected status directly

gre (dismissed):

gre:
@juan-cortes please update PR with develop.

juan-cortes:
> @juan-cortes please update PR with develop.

You got it

gre (dismissed):

juan-cortes:
This is closing it to 3 months old, no more talks about full node as far as I'm concerned, nobody is worried about this step.
Can we close it for my sanity?

gre:
to me we can merge as soon as one QA confirm the display on this specific case..

Arnaud97234 (changes requested):
@juan-cortes @dasilvarosa Wording and CTAs are OK but the last point is not implemented: 

```
If SatStack is not running, it’s a bit strange to see “Could not teach SatStack” immediately after clicking Continue. 
Option 1: Add an intermediate status showing: “Connecting to SatStack”, display “Could not reach SatStack” after 10 seconds. PREFERRED SOLUTION
Option 2: Reword: “Could not reach SatStack” to “Connecting to SatStack” QUICK FIX
```

After importing the accounts from the device, when clicking Continue, the text in the modal says "Could not reach satstack".

gre:
linter

juan-cortes:
Yeah well, rebasing broke this because the component Infobox no longer exists. 
I don't even know how to test this anymore, I'll try to catch up with @Arnaud97234

juan-cortes:
I can't seem to be able to mock the status for a satstack connection, it just doesn't want to work.
Since there's no way I can have a real satstack instance running with a synced node, I can't test this with a real environment either. I understand that the only remaining part is a small UX improvement to not show the real status disconnected because it's too fast. I thought that the code we introduced in live-common was basically ignoring the first 4 disconnected events. 

https://github.com/LedgerHQ/ledger-live-common/blob/7192225871be459b8c7fb7985d3b9432e53a1a37/src/families/bitcoin/satstack.js#L298-L302

juan-cortes:
@gre assigning you because I can't seem to mock anything for satstack and I don't have the setup to test the real thing, perhaps ask someone else to take a look. To me, it should be already denounced, not to 10 but to 5 seconds due to what I mentioned above

gre (approved):

juan-cortes:
Same, ghost task for two weeks, but pushed over **six months ago** can we either kill it or merge it?
I don't have the means to test this either with mocks or real setup so unless someone wants to take over, I vote for killing it.
cc @gre

PR #3369, opened 2020-12-08
Validators toggle assets display

henri-ly:
### Type

Bug-fix:  
the offset was off when we :
      - do a first search with few results and scroll on the bottom of it
      - then do one more search but with no result
        and redo a research with a result, no items will be display in the scrollContainer

Features : 
- @rjarasson-ledger add a `/maxVotes` to the number of selected voted
- For Polkadot Purpose we do not need the display of "All assets are distributed" because to delegate to a validator in this blockchain you only need to add the address, the blockchain will automatically distributed your locked assets.

### Parts of the app affected / Test plan

Validators list

PR #3450, opened 2021-01-07
reset scroll offset when data change on validator list

henri-ly:
### Type

Bug fix

### Context

https://github.com/LedgerHQ/ledger-live-desktop/pull/3369

We have fix one bug but discovered an another one.
When you scroll down with a few result like 19 and then you search a list with less than 3 (when there's no scrollbar) then going back to the previous result or any new result 
It will only be lock with 3 results until you reset the modal.

Validator list, and any element that use ScrollLoadingList with search


nabil-brn (approved):
Reproduced and validated 👍, it's better now.

---
0cac18c73 (HEAD -> validator-list-scroll-search-fix, Ledger-Coin-Integration-team/validator-list-scroll-search-fix) reset scroll offset when data change on validator list

PR #3469, opened 2021-01-13
LL-3255 Global coverage of features from the portfolio screen

juan-cortes:
Global coverage of features from the portfolio screen. It's good enough for a first version (to me).
Scroll top for example was left out for now.

image

### Type

Testing

### Context

https://ledgerhq.atlassian.net/browse/LL-3255


gre (approved):

juan-cortes:
Seems like it no longer wants to resize in a predictable way to accomodate all content @machard 🤣 (i cry inside)

PR #3473, opened 2021-01-14
Ll 3257 bis

machard:
Capture d’écran 2021-01-14 à 18 16 51

unfortunately since animation are deactivated and first frame empty we don't see the difference in the illustration :/


gre:
Which pr is to be merged?

machard:
@gre i made a mistake yesterday, i did not see that i created an other PR

gre (approved):

PR #3493, opened 2021-01-20
benchmark tests

machard:


machard:
Next 10 runs with 2 workers

PR #3503, opened 2021-01-21
proposal to split accounts specs

valpinkman:
closes #3538


machard:
looks like the bookmarking button is randomly failing. I think there is a transition to deactivate in the Star component.

Also, the transactions that are sometimes pending, sometimes confirmed, its stable between runs ?

valpinkman:
@machard you are right for the star, didn't even realised it. 
Also I have no idea why some operations are pending since they are all mocked /shrug
Will try to disable animation of the star and we'll run them again

machard:
I think the white version of the button is the good one (ethereum passed on the last run with the white one)

valpinkman:
yes. i updated the star animation for SPECTRON_RUN

valpinkman:
Rework with Page Object Model incoming

tests/specs/global.spec.js
```diff
@@ -1,87 +1,86 @@
...
 
   it("can open send modal", async () => {
-    const sendButton = await $("#drawer-send-button");
+    const sendButton = await portfolioPage.drawerSendButton;
     await sendButton.click();
     await modalPage.isDisplayed();
     expect(await app.client.screenshot(2000)).toMatchImageSnapshot({
       customSnapshotIdentifier: "global-send-modal",
     });
+    await modalPage.close();
```
machard, line 20: expect needs to be the last instruction of an it.
otherwise it breaks what comes next and it can be annoying (next tests and screenshot will be more wrong than they should)

tests/specs/global.spec.js
```diff
@@ -1,87 +1,86 @@
...
-    await modalPage.close();
-    const sendButton = await $("#drawer-receive-button");
-    await sendButton.click();
+    const receiveButton = await portfolioPage.drawerReceiveButton;
+    await receiveButton.click();
     await modalPage.isDisplayed();
     expect(await app.client.screenshot()).toMatchImageSnapshot({
       customSnapshotIdentifier: "global-receive-modal",
     });
+    await modalPage.close();
```
machard, line 30: expect needs to be the last instruction of an it.
otherwise it breaks what comes next and it can be annoying (next tests and screenshot will be more wrong than they should)

machard (changes requested):

machard (dismissed):

machard (approved):

nabil-brn (approved):

PR #3504, opened 2021-01-22
put back channel notif on develop failure + fix ci comment

machard:


gre (dismissed):
wow i didn't know we could use the outputs of something back into the next step of a github actions

juan-cortes (approved):

PR #3509, opened 2021-01-25
use custom version of live-common to test proxy

### LINKED ISSUE #1073: network detection does not work through a proxy (squid)

- Ledger Live 1.0.1
- Platform: linux

When the manager is opened, ledger live checks Internet connectivity.

If I have a proxy (configured via environment variables), the program does not detect that connectivity, it shows a red message “Oops, internet seems down” and offers to retry; which never detects it (see attached screenshot ledger-live.proxy-netdown.png).

On my squid3 logs I have messages like this one: “TAG_NONE/400 4062 NONE error:invalid-request - HIER_NONE/- text/html”

If I unset proxy environment variables, then network detection, and the program, works fine.

I could attach application.log if you need it, I’m guessing you won't since I've been able to reproduce it on two different computers (one with debian sid, the other with debian stretch, both updated to date).

Thanks
ledger-live proxy-netdown

gre:
attempt to fix #1073 
using https://github.com/LedgerHQ/ledger-live-common/pull/1026


gre:
status: current solution is not yet working. need to work more on #1073 to see what is still not working.

gre:
to be continued

PR #3510, opened 2021-01-26
Make Ethereum sync on renderer instead of internal

gre:
thanks to brainstorm on slack with @rjarasson-ledger , we identified a problem in LLD.

The problem is in this part

```js
const cmdAccountSync = (o: {
  account: AccountRaw,
  syncConfig: SyncConfig,
}): Observable<AccountRaw> => {
  const account = fromAccountRaw(o.account);
  const bridge = bridgeImpl.getAccountBridge(account, null);
  return bridge.sync(account, o.syncConfig).pipe(map(f => toAccountRaw(f(account))));
};
```

and the fact we will use `account` in `f => toAccountRaw(f(account))`

that `account` is potentially **outdated** because on the renderer side, it could have been updated (e.g. pendingOperations)

---

There are 2 solutions for this I think:
- this current PR: is fixing it just for Ethereum, by running sync on render. it's probably the future target solution btw because internal<>renderer have a cost.
- another solution, that would work for all coins, is to keep accounts up to date on **internal** side, maybe by sending accounts update when they change and maybe only when a sync on such account is pending.

### DIFF
--- a/src/renderer/bridge/proxy.js
+++ b/src/renderer/bridge/proxy.js
@@ -60,15 +60,22 @@ export const getCurrencyBridge = (currency: CryptoCurrency): CurrencyBridge => {
   return b;
 };
 
+// the proxied sync have problems because of patchAccount limitations:
+// - perf overhead
+// - race-condition with pendingOperations
+const canSyncOnRenderer = currency => currency.family === "ethereum";
+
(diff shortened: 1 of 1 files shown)


gre:
I think we'll reopen this idea next sprint.

PR #3512, opened 2021-01-26
LL-4416 fix swap errors display

machard:
https://ledgerhq.atlassian.net/browse/LL-4416

needs https://github.com/LedgerHQ/ledger-live-common/pull/1031


gre (changes requested):
we actually must fix that in live-common's side.

PR #3513, opened 2021-01-26
LL-4391 Analytics and stylings for starred sidebar

juan-cortes:
In order to determine if the feature is highly used or not to be able to make a decision on an incoming sidebar reword of what to do with the starred accounts section, this pr introduces a few metrics and in the meantime fixes the styles in the placeholder so that it's not cut-off when in the smallest window size.

### Type

UI Polish + analytics

### Context

https://ledgerhq.atlassian.net/browse/LL-4391

Metrics can be verified by enabling the `ANALYTICS_CONSOLE` flag when launching the app, there should be a new metric in the accounts page listing the number of accounts with stars, and then the actions for accessing an account from the sidebar, starring and unstarring.


juan-cortes:
This will likely break the screenshots, so either we wait until they are generated or I generate them manually 🤔

gre (approved):

nabil-brn (approved):

PR #3515, opened 2021-01-26
webpack: add support for mjs files

henri-ly:
### Type

Bug fix

### Context

We are needed to use Polkadot.js new lib, but it's seems that they are using some mjs that wasn't compatible with LLD.
(cf. https://ledger.slack.com/archives/C011AEKR21Z/p1611656602028500)

We tried once to update the lib on LLC but didn't test it on LLD. It was 

refer to the screen it shouldn't be able to launch the app

### DIFF
--- a/preloader.webpack.config.js
+++ b/preloader.webpack.config.js
@@ -67,6 +67,11 @@ module.exports = {
           },
         },
       },
+      {
+        type: 'javascript/auto',
+        test: /\.mjs$/,
+        use: []
+      }
     ],
   },
   resolve: {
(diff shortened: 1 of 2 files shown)


gre (approved):

gre:
If this PR builds and pass tests, it can be merged imo.

juan-cortes (approved):

gre:
@henri-ly your commit is not signed. can't merged at the moment. (maybe I need to remove the restriction but still)

henri-ly:
The sign has been done @gre and @juan-cortes it should be alright now

rjarasson-ledger:
I tested and it solves the issue with @polkadot/types v3.6.4 imports.

rjarasson-ledger (approved):

PR #3516, opened 2021-01-26
LL-4448 Disable beginAtZero for the chart

### LINKED ISSUE #3507: In Ledger Live the historical weekly view looks flat. It should be scaled so that min is the closest even number less than the min for the period
See title

juan-cortes:
image

### Type

UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-4448
closes #3507 

For the chart of a given account / portfolio, it should only start on zero if the selected period contains a data point with a value of zero. So if I select Week and I only fluctuate between 0.5 and 0.6, the scale of the Y axis should be constrained to these values and any fluctuations will be a lot more visible (second screenshot)

### DIFF
--- a/src/renderer/components/Chart/index.js
+++ b/src/renderer/components/Chart/index.js
@@ -159,7 +159,7 @@ export default function Chart({
               zeroLineColor: theme.text.shade10,
             },
             ticks: {
-              beginAtZero: true,
+              beginAtZero: false,
               suggestedMax: 10 ** Math.max(magnitude - 4, 1),
               maxTicksLimit: 4,
               fontColor: theme.text.shade60,


juan-cortes:
Pending screenshots

src/renderer/components/Chart/index.js
```diff
@@ -159,7 +159,7 @@ export default function Chart({
               zeroLineColor: theme.text.shade10,
             },
             ticks: {
-              beginAtZero: true,
+              beginAtZero: false,
```
gre, line 162: :smooth:

gre (approved):

src/renderer/components/Chart/index.js
juan-cortes: it's probably false by default but this seemed nicer :trollface:

nabil-brn (changes requested):
It crashes when I displayed the 1W chart the first time and now it works but I can still reproduce the crash with another userdata set.



For me it's a _nogo_ until we find the root cause because of the random nature of the bug.

---
2.20.0-sha.30adb6e06fefef3395bee0a15067be7d9db93f1e-win.exe.zip

juan-cortes:
I see no reason for a correlation between that crash and this change. If you can provide me with a way to reproduce the bug I'd be happy to look into it though

nabil-brn (approved):
✓ QA-ed

---
2.20.0-sha.30adb6e06fefef3395bee0a15067be7d9db93f1e-win.exe.zip

PR #3517, opened 2021-01-26
Fixes LL-4244

gre:
### Type

bug fix

### Context

see also https://ledgerhq.atlassian.net/browse/LL-4458 (it's not fixing https://ledgerhq.atlassian.net/browse/LL-4458 but it's a workaround before we do)

firmware update


gre:
automating the test of this would be interesting btw. but not doing this now. cc @machard

valpinkman (approved):

Arnaud97234 (approved):

PR #3519, opened 2021-01-27
LL-4458 (Analytics): change method of tracking source of events

LFBarreto:
(Analytics): change method of tracking source of events
### Type

Bug fix
### Context

LL-4458

Every navigation trigerring some tracking should track the source on the target page. like before without potential rerender bugs introduced by previous method.


gre (dismissed):

gre:
@LFBarreto  conflicts

Arnaud97234 (approved):

gre:
let's prettier it and i think if it pass we are good. arnaud approved so essentially if we keep same code it's still good.

PR #3520, opened 2021-01-27
cmd AccountSyncSet: internal reconcile with latest account during sync

gre:
This deliver the 2nd suggested solution for the problem discussed in #3510. The benefit of this solution is it's a bit safer to take and it prevent likelihood of race condition on all currencies we have today, in case an account get modified during a AccountSync.

Sans titre

Essentially, we add a new renderer->internal command called "AccountSyncSet" that is called when a sync is ongoing and each time the local state of renderer have an Account reference change. that gives opportunity to let internal being up to date with latest account state. NB: race condition still possible, but very unlikely (cmd runs very fast)

Capture d’écran 2021-01-27 à 12 18 57

### Type

Bug fix

### Context

discussion with @rjarasson-ledger 

account synchronisation for all accounts.
the race condition were not really causing problem in existing coins (afaik) but it will be important for the future.

### DIFF
--- a/src/renderer/bridge/BridgeSyncContext.js
+++ b/src/renderer/bridge/BridgeSyncContext.js
@@ -1,7 +1,8 @@
 // @flow
 
-import React, { useCallback } from "react";
+import React, { useCallback, useEffect, useRef } from "react";
 import { BridgeSync } from "@ledgerhq/live-common/lib/bridge/react";
+import { toAccountRaw } from "@ledgerhq/live-common/lib/account";
 import { useSelector, useDispatch } from "react-redux";
 import logger from "~/logger";
 import { updateAccountWithUpdater } from "~/renderer/actions/accounts";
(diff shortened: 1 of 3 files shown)


valpinkman (approved):

valpinkman:
Sorry I wrongly clicked the `ready for review`.

However I do like this approach to have an updater with the latest account. LGTM

rjarasson-ledger:
@gre I was unable to reproduce our bug with this fix.
I tried 8 transaction with various timings, and haven't seen pending operations disappear.

Hope this is definitely fixed.

gre:
Ok thanks!

gre:
note for future:
- if a AccountSync finishes when a AccountSyncSet of the same account was launched and has not finished (with a slight delay, actually not finished "recently"), we could discard the effect of the sync (aka we don't apply the account update at all). it's a very rare case tho, but it could be a way to fix it.

gre:
I've merged that PR inside https://github.com/LedgerHQ/ledger-live-desktop/pull/3525 so we focus on just that other wider PR regarding testing.

PR #3528, opened 2021-01-29
LL-4514 - Wording for ledger status

juan-cortes:
image

### Type

Wording

### Context
https://ledgerhq.atlassian.net/browse/LL-4514


juan-cortes:
Added support for the help drawer on the spectron specs, since the wording here affects the feature and it wasn't covered. We should probably improve our tests like this whenever we touch a part of the app that's not covered 🤔 
cc @gre

gre:
@juan-cortes test CI fails.


for the test code, i have a feeling that everytime we do this:

```
await app.client.pause(400)
```

we should replace by something like

```
await waitFor_some_UI_To_Refresh()
```

with a specific condition to wait for

juan-cortes:
If you look at the screenshot diff it seems like there's a row that has a hover/active effect, not related to the pr's code at all. 
I bet running it again would work 🤔

juan-cortes:
> @juan-cortes test CI fails.
> 
> for the test code, i have a feeling that everytime we do this:
> 
> ```
> await app.client.pause(400)
> ```
> 
> we should replace by something like
> 
> ```
> await waitFor_some_UI_To_Refresh()
> ```
> 
> with a specific condition to wait for

You are probably right

juan-cortes:
🤷🏼‍♂️ @machard any clue what can be causing the hover effect?

machard:
I don't understand why it works on develop but not in here :/

machard:
I think @valpinkman will stabilize accounts tests

valpinkman:
@machard @juan-cortes the PR is ready, just awaiting validation

valpinkman (dismissed):

Arnaud97234 (changes requested):
@juan-cortes In French, desc is still "Learn crypto". Translation needed ?

Arnaud97234 (dismissed):

gre (approved):

gre:
For translation we should not update them. We should not test a feature but in English @Arnaud97234 
Translation get invalidated when a feature is merged.

gre:
@juan-cortes conflicts

juan-cortes:
> @juan-cortes conflicts

Should be fine now, I didn't update the test though since @valpinkman was working on removing the getters,
it can addressed on a separate pr

gre:
tests are failing 🌊

juan-cortes:
Yeah things tend to fail went left unmerged for weeks at a time :trollface: hopefully it will work now

gre:
shipped through mashup PR

PR #3532, opened 2021-02-01
CI - add docker caching + remove nev7n/wait_for_response

machard:
try docker caching.
unfortunately it will not cache the part related to the node_modules but i think it should speeds things up a bit already.


machard:
before
Capture d’écran 2021-02-01 à 16 46 01
after
Capture d’écran 2021-02-01 à 16 46 13

machard:
i'm not sure what is the behavior if package.json file change. i'll try changing live common to see what happens

juan-cortes (approved):

juan-cortes (approved):

PR #3541, opened 2021-02-02
LL-4423 bump electron + libcore bindings

valpinkman:
- it includes a rework of Cosmos with a new implementation that can be tested only with TESTNET with EXPERIMENTAL_CURRENCIES=cosmos_testnet
- for the actual Cosmos (mainnet) it is likely to be broken until stargate happens.
- however, this PR is not mergeable until we have a new version of libcore with latest stargate fix

---

This updates electron and libcore to the latest versions. 

⚠️ the libcore is only in beta for now, we need this PR tested to be sure it's safe

Also, coming with electron 12.0, `contextIsolation` will be set to true by default in `webPreferences` when instanciating a new window. 
From what I've tested, we are not ready for contextIsolation, as we use `require` on the renderer side.
https://www.electronjs.org/docs/breaking-changes#planned-breaking-api-changes-120
https://github.com/electron/electron/issues/23506
https://github.com/electron/electron/blob/master/docs/tutorial/security.md#3-enable-context-isolation-for-remote-content

This needs to be discussed as it improves the security BUT will require us to change a bit the way we do stuff on the render side. Either that or we just keep it to `false`

### Type

Chore

### Context

LL-4423


Arnaud97234 (changes requested):
Can't access to Manager or add account.
When I want to reach the Manager I see a looping loader but it never ask to plug the device.

On Linux LL crash at launch with `Internal killed by signal (SIGSEGV)`
[ledgerlive-logs-2021.02.02-19.31.22-9f5bd463.json.txt](https://github.com/LedgerHQ/ledger-live-desktop/files/5913611/ledgerlive-logs-2021.02.02-19.31.22-9f5bd463.json.txt)

PR #3544, opened 2021-02-04
CI - use vercel to cancel previous flows

machard:


valpinkman (approved):

machard:
actually i think that instead of canceling previous flow, we should just avoid posting a comment if a new run has been started AFTER.
===> we get the check/crosses on intermediate commit AND only the last comment

valpinkman:
I think a rebase is needed

machard:
looks like starting today the pull_request token has permission to cancel previous workflow :/

valpinkman:
as seen with @machard, we have transfered the repo to ledgerhq, awaiting validation to integrate it in Vercel.

when it's done, we'll need to update the url to reflect the change.

valpinkman (changes requested):
update url when ready

valpinkman (approved):

PR #3546, opened 2021-02-04
CI - try 4 workers

machard:
Beside putting more pauses to ensure stability with more worker I had to fix the reordering hook that was nesting useEffects and it was causing unpredictable sorting behaviors.
(The swap screenshot is still randomly failing, like 1 out of 20 runs, the swap transaction disappear)


machard:
usual random swap error

src/renderer/actions/general.js
```diff
@@ -100,19 +100,20 @@ export function useRefreshAccountsOrderingEffect({
...
     }
 
     return () => {
       if (onUnmount) {
-        refreshAccountsOrdering();
+        dispatch(reorderAccounts(comparator));
       }
     };
-  }, [onMount, onUnmount, refreshAccountsOrdering]);
+  }, [comparator, dispatch, onMount, onUnmount]);
```
gre, line 116: 🤔  we are no longer using the `useRefreshAccountsOrdering` ?

gre:
```js
  const comparator = useSortAccountsComparator();
```
  
comparator will often change references. for instance, when countervalues reload

```js
  }, [comparator, dispatch, onMount, onUnmount]);
```
  
it’s injected in the useEffect, making it re-run on each countervalues change

```js
  if (onMount) {
      refreshAccountsOrdering();
      dispatch(reorderAccounts(comparator));
    }
```
    
in the effect, if the useRefreshAccountsOrderingEffect was used with onMount, it actually no longer just run “onMount” but on each countervalues update
and we absolutely not want account to reorder “on the fly” because it create click jacking. this need to happen at precise lifecycle, like loading Accounts screen, etc..

gre (changes requested):
we should not change the countervalues hook logic. rest is ok.

machard:
still flaky but tolerable:
swap history bug
onboarding

valpinkman (approved):

gre (approved):

PR #3552, opened 2021-02-08
LL-4524 add firmware update step for the new "updating" phase

gre:
### Type

feature

### Context

LL-4524

existing firmwares are not much impacted. I actually tested 1.3.1->1.6.0->1.6.1 when doing this PR.

The real thing to test is the 1.6.1->2.0.0* which should have an extra step.

current UI proposal looke like this

Capture d’écran 2021-02-08 à 18 57 39


gre:
i thought this issue was over

cthepot-ledger:
the success description message should be 
“You can now install apps on your device.” rather than "Please reinstall the apps on your device"

gre:
Ok this is super confusing to me. we'll wait on monday when valentin is back. because i don't understand why we added these

```
  "successTextApps": "Please re-install the apps on your device",
      "successTextNoApps": "You can now install apps on your device",
      "sucessCTAApps": "Re-install apps",
      "SuccessCTANoApps": "Install apps",
```

to me we already had the "success apps" and it was just a "Done" button. my understanding is that we wanted to only change two wordings, not add new ones.

valpinkman:
@gre context here https://ledgerhq.atlassian.net/browse/LL-4571

gre:
shipped through mashup pr

PR #3557, opened 2021-02-09
Coin 1145 display fees desktop

henri-ly:
### Type

UI add info

### Context

https://ledgerhq.atlassian.net/browse/COIN-1145

Change on modal of all confirmation, it should not be affected by any other coin + refacto of Info box

### DIFF
--- a/src/renderer/components/InfoBox.js
+++ b/src/renderer/components/InfoBox.js
@@ -15,7 +15,37 @@ type Props = {
   onLearnMore?: () => void,
   learnMoreLabel?: React$Node,
   horizontal?: boolean,
-  type?: "primary" | "secondary" | "warning" | "security" | "danger" | "success", // TODO implement the styles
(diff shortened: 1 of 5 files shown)


rjarasson-ledger (dismissed):

gre (approved):

Arnaud97234 (changes requested):
The fees amount is still missing for the Nominate operation
Screenshot 2021-02-09 at 17 37 31

henri-ly:
Weird, it's probably because your live-common is not up to date ? there's no dump yet with live-common and the fees field, you will need to yalc it I guess

rjarasson-ledger:
@Arnaud97234 This modal is scrollable (as you can see in your screenshot) - the field should be below.

The fees are not part of the transaction to be validated on device, but set by the network on broadcast.
So we added an infobox to inform that the user will not see it on device even though it's indicated.

Arnaud97234 (approved):

gre:
shipped through mashup pr

PR #3560, opened 2021-02-09
fix(Polkadot): remove transferAmount in OperationsList - LL-4435

rjarasson-ledger:
### Type

UI/UX Fix

### Context

For Polkadot, OUT operations (send) used to have a transfer amount column with a positive value in the Operations List.

It's been acted that this is not consistent with other coins.

This removes the amounts, and only let the effective transaction amount.

BEFORE
image

AFTER
Transfer Amount should not be present in OperationsList.

image

### DIFF
--- a/src/renderer/families/polkadot/operationDetails.js
+++ b/src/renderer/families/polkadot/operationDetails.js
@@ -278,32 +278,6 @@ type Props = {
   unit: Unit,
 };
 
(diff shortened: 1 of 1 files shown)


gre (dismissed):

Arnaud97234 (approved):

gre:
shipped through mashup pr

PR #3574, opened 2021-02-15
change getters to functions in page model

valpinkman:
To unify the way we test, i changed all getter calls to function calls

very readable
much wow


gre (approved):

Arnaud97234 (approved):

juan-cortes (approved):
lgtm

PR #3575, opened 2021-02-15
Test/remove delays

nabil-brn:
Remove some delays and fix false negative results.

### Type

Test robustness

### Context

Reduce test flakiness by replacing static delays by dynamic ones.

Some components & `tests\`


machard:
@nabil-brn on linux files need to have the correct case.
Rename SendModal.page.js to sendmodal.page.js

PR #3590, opened 2021-02-19
Prevent console error when parsin the svg path

juan-cortes:
image

A path can't finish with a semicolon, which caused this initial rendering of the wave to trigger that error, the subsequently render the animated path. It didn't seem to cause any rendering problems from what I see, but it does log that error any time the wave is present, so a bit of code cleanup.

### Type

Code cleanup


gre (approved):
:svgmaster:

juan-cortes:
Can we merge, testing this is literally making sure there's no svg error warning on the console, functionality is not affected

PR #3628, opened 2021-03-03
ci - LL-4593 - generate and upload cpuprofiles

machard:
https://kwonoj.github.io/en/post/electron-content-trace/
one profile per spec file attached as a ci artifact


gre (approved):
that's nice

PR #3630, opened 2021-03-04
feat(ReadOnlyAddressField): ensure address last line has MINLENGTH (5 chars) - LL-4656

rjarasson-ledger:
### Type

UI Polish

### Context

When displaying the device address in the Receive Flow (also used in Exchange Device Confirm), long address can break line arbitrarily.

We want to avoid adresses to break line with too fews characters on a second line (can be misleading or look like a bug).

This fix ensures at least 5 chars are display if address breaks line. It won't break at all if there is enough room.

On Receive Flow:
* address with enough room
image

* address breaking line
image

* address breaking line (simulate very long address):
image

### DIFF
--- a/src/renderer/components/ReadOnlyAddressField.js
+++ b/src/renderer/components/ReadOnlyAddressField.js
@@ -6,6 +6,8 @@ import styled from "styled-components";
 import Box from "~/renderer/components/Box";
 import IconCopy from "~/renderer/icons/Copy";
 
+const LINE_MINLENGTH = 5; // Minimum of chars for the last line (if multiline)
+
 const Address = styled(Box).attrs(() => ({
   bg: "palette.background.default",
   borderRadius: 1,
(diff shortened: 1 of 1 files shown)


gre (approved):

nabil-brn (approved):
Tested receive flow with many crypto assets

---
2.23.0-sha.14875786458e347324dcc6354ec1be23b142b547-win.exe.zip

valpinkman:
@rjarasson-ledger ready to merge, can you rerun the CI to make sure nothing is broken ?

gre:
imo it likely can't be broken if at any given of time CI has passed. otherwise there would be "screenshot conflicts". imo we can process to merge and worse case is a later fix. otherwise we can't scale this.

gre:
oh it's not passing CI. mmh

gre:
ok go

PR #3674, opened 2021-03-16
live-common 18.8.x

gre:
- live-common release https://github.com/LedgerHQ/ledger-live-common/releases
  - LL-3936 Preferred new account scheme (Bitcoin family). Included with @LFBarreto's work. closes #1573 
  - Notification center work => *this is not testable, it's not used by this PR, no impact*
  - ~~Stellar rework in JavaScript => **Stellar to test**~~
  - Portfolio V2 rework => **This is not used yet! so no impact**
  - Polkadot fixes.  COIN-1401 COIN-1398 COIN-1401
  - ~~Ethereum. LL-4834 – GasLessThanEstimate is now an error (#1088) that will prevent users to try to go lower than the estimation which will mostly always fails at broadcast.~~
  - Ethereum Ropsten accounts are no longer derived on 44'/1'/ path but on 44'/60'/ which means they will have same address as Ethereum accounts. It's the accepted convention in Ethereum ecosystem. If you want to keep your previous ropsten, just use an older Live to transfer them. (no need to scan these as "legacy" even tho we could make it so)
  - Some tokens on Ethereum Ropsten are now available. You can get some ropsten USDC here: https://usdcfaucet.com/
  - Some fixes in cosmos

---

18.8.1 updates:

- fixes Network Down issue on iOS
- rollback some of the points before
- ripple lib minor bump
- wallet connect minor bump

18.8.2 updates:

- COIN-1423 [Polkadot] Keep a safety buffer of (3 * fees) for bond max


Arnaud97234 (commented):
@LFBarreto Derivation type: see subtasks in https://ledgerhq.atlassian.net/browse/LL-3936

Arnaud97234 (commented):
@gre Those are not blockers, if we are ok to fix them later we can merge the PR:
- Bitcoin Derivation type: https://ledgerhq.atlassian.net/browse/LL-4904
- ~~Error message when sending max on Polkadot: https://ledgerhq.atlassian.net/browse/LL-4907~~ -> Should be fixed by  https://github.com/Ledger-Coin-Integration-team/ledger-live-common/tree/fix/COIN-1398
- USDC is not in the ropsten token list
- gasLimit error message: text is very close to the input field
gasLimitError

Arnaud97234 (changes requested):
When sending funds with a memo type Hash, the following error is displayed when broadcasting the tx:
image

[ledgerlive-logs-2021.03.18-11.48.21-5d08930ca.json.txt](https://github.com/LedgerHQ/ledger-live-desktop/files/6163397/ledgerlive-logs-2021.03.18-11.48.21-5d08930ca.json.txt)

gre:
@machard do you know what's going on here? thx

Arnaud97234 (changes requested):
**Bitcoin family - Derivation type**
1) User will have to reset Ledger Live and import his account to see the account name properly. Even after clearing the cache or reimport the accounts from the device the derivation is still in the account name.

2) In many screen the derivation ID is missing:
https://ledgerhq.atlassian.net/browse/LL-4904

gre:
> User will have to reset Ledger Live and import his account to see the account name properly. Even after clearing the cache or reimport the accounts from the device the derivation is still in the account name.

Yes. nothing we can do here 🤔

Arnaud97234 (dismissed):

nabil-brn (approved):

PR #3684, opened 2021-03-20
fix: Add missing comma in en i18n json file

chabroA:
Add a missing comma in the `static/i18n/en/app.json` file.

### Type

Bug Fix

### Context

A comma was missing in the english i18n json file, resulting in an error when starting the app.

PS: I couldn't find contributor guidelines so apologies in advance if I am not following recommended procedure.


gre:
wow, good catch! thanks.

PR #3693, opened 2021-03-23
LL-4839 (Manager): some apps should only be uninstallable

LFBarreto:
(Manager): some apps should only be uninstallable
disable install for a given set of apps

### Type

Feature
### Context

[Ll-4839]

Manager app list should have a given set of apps install button disabled for.


Arnaud97234 (approved):
Tested with a nanoS 1.3.1: Ripple app is only available for uninstallation while user can install it with the current prod version.

gre (approved):

PR #3695, opened 2021-03-23
LL-2886 fix behaviour on firmware update when user cancels

valpinkman:
Slightly change the rendering behaviour on error during firmware update when user cancels

### Type

Fix

### Context

LL-2886

Modal Firmware Update

image

### DIFF
--- a/src/renderer/Default.js
+++ b/src/renderer/Default.js
@@ -1,6 +1,6 @@
 // @flow
 import React, { useEffect, useRef } from "react";
-import { Route, Switch, useLocation } from "react-router-dom";
+import { Redirect, Route, Switch, useLocation } from "react-router-dom";
 import TrackAppStart from "~/renderer/components/TrackAppStart";
 import { BridgeSyncProvider } from "~/renderer/bridge/BridgeSyncContext";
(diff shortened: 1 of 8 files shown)


juan-cortes (approved):
lgtm

MortalKastor (dismissed):

Arnaud97234 (changes requested):
Still not refreshing when clicking the Close button

Arnaud97234 (approved):

juan-cortes (approved):

src/renderer/Default.js
```diff
@@ -98,6 +98,7 @@ export default function Default() {
                     <Route path="/" exact render={props => <Dashboard {...props} />} />
                     <Route path="/settings" render={props => <Settings {...props} />} />
                     <Route path="/accounts" render={props => <Accounts {...props} />} />
+                    <Redirect from="/manager/reload" to="manager" />
```
machard, line 101: i've seen other suggest history.go(0).
I don't have opinion

machard (approved):

src/renderer/Default.js
valpinkman: history.go(0) reloads the whole app. Not possible to use

PR #3700, opened 2021-03-25
LL-4446 Swap floating rates

juan-cortes:
> It depends on https://github.com/LedgerHQ/ledger-live-common/pull/1117 being merged and bumped, currently not testable until that is the case.

Due to the postponed release, reopening a pr with fixes for when this is a go again.

### QA checks

- [x] Switching between two erc20 tokens should no longer keep the amount of the first selected one.
- [x] On device confirmation step, when doing a floating rate swap to an ERC20 with less decimal precission than ETH the amount was formatted as ETH instead of the token. Payout fees are charged on the receiving swap currency, not the parent one.


nabil-brn (commented):
Bad provider link color when theme is dark

image
image

gre (dismissed):

gre:
wtf

gre (approved):

Arnaud97234 (approved):

PR #3709, opened 2021-03-29
LL-4949 Stellar JavaScript rework

gre:
**This is a QA only PR, we won't merge it, it will be part of a future live-common bump**

Reimplementation of Stellar with stellar-sdk

### Type

Refactoring

### Context

LL-4949

Stellar


gre:
:wtf:

Arnaud97234:
@gre When coming from an old app.json file, the account setting have the following. Both "xpub" and "id" are present in the data while for new account imported only "id".
Clear cache or re-importing accounts doesn't solve the issue. 
image

gre:
to me it's not really a problem unless you see some bad consequence of this? the xpub is simply ignored. we could clean it up but maybe it's not even needed 🤔  the .xpub field will not be used by anything but libcore specifics need. in future .xpub would only be used by Bitcoin family.

Arnaud97234 (approved):

Arnaud97234 (approved):

PR #3726, opened 2021-04-09
LL-5190 Operations list: do not apply ellipsis for short recipients

ghost:
(low priority)

Context: https://ledgerhq.atlassian.net/browse/LL-5190

Proposed fix: do not split the displayed recipient if it's shorter than an arbitrary 24 characters.


gre (changes requested):
flow issues

PR #3737, opened 2021-04-13
fix carousel glitch

machard:
there was a glitch when moving the mouse outside of the banner because the component was unmounted/remounted


juan-cortes (approved):

gre:
@machard is there a LL ticket?

machard:
@gre not that i am aware

nabil-brn (approved):

PR #3743, opened 2021-04-14
LL-2985 (AccountPage): add bitcoin address tabs

LFBarreto:

### DIFF
--- a/src/renderer/families/bitcoin/AccountTabs.js
+++ b/src/renderer/families/bitcoin/AccountTabs.js
@@ -73,7 +73,7 @@ function AddressesTab({
           </Cell>
           <Button
             onClick={() => {
-              receiveAddress(address, derivationPath);
+              receiveAddress(i);
             }}
           >
             {t("accounts.contextMenu.receive")}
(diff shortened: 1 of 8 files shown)


src/renderer/families/bitcoin/AccountTabs.js
```diff
@@ -0,0 +1,129 @@
...
+  const { t } = useTranslation();
+  const dispatch = useDispatch();
+
+  const freshAddresses = account.freshAddresses || [];
+
+  const receiveAddress = useCallback(
+    (address, path) => {
+      dispatch(
+        openModal("MODAL_RECEIVE", {
+          account: { ...account, freshAddress: address, freshAddressPath: path },
```
LFBarreto, line 97: @gre :wave: hi we need your input on this part of the code.
What we need to do is be able to trigger a receive flow from a specific bitcoin address. 
The common logic as of now doesn't allow that directly so either we do this little trick here in "emulating" an account with the address we want to receive on, or we update the receive logic so we can force these params maybe.
WDYT?

gre: no we shouldn't do this. we need instead to pass a second parameter that tells the index to use.
that index needs to be passed all along. typically to the bridge.receive.
we must do this properly 🙏 

in scope of this, we must make an evolution on CLI to also debug it from there, etc.. i'm sorry to not have anticipated on that. let's try to write it down for next sprint.

src/renderer/families/bitcoin/AccountTabs.js
```diff
@@ -0,0 +1,129 @@
...
+          parentAccount,
+          skipAccountSelect: true,
+        }),
+      );
+    },
+    [openModal, dispatch, account, parentAccount],
+  );
+
+  console.log(freshAddresses);
```
gre, line 106: console.log leaked

src/renderer/families/bitcoin/AccountTabs.js
gre: ok looks like it's an index in freshAddresses? so index? but what about old addresses tho. Didn't we wanted to list used addresses too?

gre (changes requested):
let's discuss

gre:
it seems this work is postponed so we'll close. work can always be continued from this history.

PR #3744, opened 2021-04-14
Fix swapicon color

hzheng-ledger:
### Type

Bug fix

### Context

https://ledgerhq.atlassian.net/browse/LL-5192

Screenshots for different themes :
Screenshot 2021-04-14 at 16 05 49
Screenshot 2021-04-14 at 16 05 57
Screenshot 2021-04-14 at 16 05 43

The above colours are given by Brian


juan-cortes (dismissed):
looks good to me

juan-cortes:
I think this is caused by the countdown animation in the background behind the modal @hzheng-ledger 
There was a fix from @nabil-brn somewhere that would need to be merged before the screenshots will work, as you can see the only difference is the little clock in the background.
image

IAmMorrow:
So this Jira ticket was lacking a lot of context and only mentioned "getting rid of transparency" for no apparent reasons. I digged a bit and now have a full explanation. Transparency is how we achieve shading through the theme. This is made to avoid specializing the theme for any use case like it's being made in this PR. Turns out the issue here is only that the Icon was built in an odd way and doesn't handle transparency properly. 

We should not merge this PR but rather simply replace the icon.

PR #3745, opened 2021-04-16
LL-5234 - switch operation details to a drawer

machard:
https://ledgerhq.atlassian.net/browse/LL-5234

Test plan;
probably have to check all flow where this modal appears...


src/renderer/drawers/Provider.js
```diff
@@ -0,0 +1,51 @@
...
+
+type State = {
+  Component: *,
+  props: *,
+  open: boolean,
+};
+
+// actions
+// it makes them available and current from connector events handlers
+export let setDrawer: (Component?: *, props?: *) => void = () => {};
```
LFBarreto, line 13: i dont get the use of this here :thinking: you want to be able to control the content outside of the Provider context?
i mean why use a context if you're not using it :sweat_smile:

src/renderer/drawers/Provider.js
```diff
@@ -0,0 +1,51 @@
...
+    ...update,
+  };
+};
+const initialState = {
+  Component: null,
+  props: null,
+  open: false,
+};
+
+export const context = React.createContext<State>(initialState);
```
LFBarreto, line 28: You export the context here but dont seem to use it anywhere :shrug:

LFBarreto (changes requested):
a small feedback on the context use. either update to use the context hooks inside the components with something like that 
```javascript
 const {setDrawer} = useContext(DrawerContext) 
```
or just don't use a context at all :shrug: and dispatch the event like you already did but its not really "react" :goat:

src/renderer/drawers/Provider.js
machard: its used in the root of the app, in Default

src/renderer/drawers/Provider.js
machard: i am using it. at the root of the app. instead of having a map key => component like we do for modal, i just passed directly the component

machard:
ha right lol, i know what you mean, indeed i don't read the content of the context anywhere.

src/renderer/drawers/Provider.js
machard: updated i ve extracted the drawer component

src/renderer/drawers/Provider.js
machard: i've updated, extracted the drawer component, its now using the context

machard: more generally, i've discovered that recently, i prefer to import the action than passing it in the context

src/renderer/drawers/Provider.js
```diff
@@ -0,0 +1,37 @@
...
+
+const DrawerProvider = ({ children }: { children: React$Node }) => {
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  setDrawer = (Component, props) => dispatch({ Component, props, open: !!Component });
+
+  return <context.Provider value={state}>{children}</context.Provider>;
```
LFBarreto, line 34: you could expose setDrawer directly in the context something like that
```js 
value={{ ...state, setDrawer }}
```
Its just the fact of exposing mutable functions works in this case but can also be prone to runtime errors.
Thats what my previous comments were about sorry if that wasn't clear enough :sweat_smile: 
Also setDrawer is set each time the provider updates :shrug: if you really wish to preserve this way of doing it try and put the setDrawer = function inside a useEffect at mount maybe

LFBarreto (changes requested):

src/renderer/drawers/Provider.js
gre: mmh i would not recommend to do `value={{ ...state, setDrawer }}` because it makes triggering a redraw each time DrawProvider itself get re-renderer. However the pattern at stake here is a bit weird to me too. There are many possibility indeed. many we should discuss in a quick brainstorm of what we try to achieve and what could be the best.

- your solution is one
- redux is one
- using a event emitter OR a rxjs Subject would be one (to get the data in, then you .subscribe to it)

LFBarreto (dismissed):

machard:
state: can be merged as is. existing design has been adjusted.

left to do : design change to the drawer (amount in an other view + activity log)

LFBarreto:
closing this for #3844

PR #3750, opened 2021-04-19
Make countdown animation disabled in spectron run

juan-cortes:
Brings the fix from @nabil-brn so that we no longer break spectron screenshots caused by the small animated countdown clock on the swap form.


rjarasson-ledger (dismissed):

juan-cortes:
I mean, swap tests are going to keep failing until we have the work from nabil for the new form anyway

gre (approved):

PR #3751, opened 2021-04-19
LL-4560 (Buy/Sell): add provider logo

LFBarreto:
- (Buy/Sell): add coinify logo

### Type

UI Polish
### Context

LL-4560

Buy Sell first page


src/renderer/screens/exchange/hooks.js
```diff
@@ -44,3 +46,23 @@ export const getAccountsForCurrency = (
...
 };
+
+const PROVIDERS = {
+  COINIFY: {
+    id: "Coinify",
+    iconResource: coinifyIcon,
+  },
+};
+
+const exchangeProviderRef = createRef<{ id: string, iconResource: any }>();
```
gre, line 57: to me, using a ref as a global variable is not necessary and overkill

- either put it in place of the hook. (imo it's better to not "leak" global state, unless there is a reason?)
- or if it's needed global, i would just use a `let`

gre (changes requested):

src/renderer/screens/exchange/hooks.js
LFBarreto: i updated it to a `let` for now since the info as well as the update needs to be shared on buy and sell screens and i was a bit lazy to create a context just for this :thinking: 
This will need to be reworked in any case with the wyre integration for the switch between providers, its just a very simple boilerplate and maybe needs to be reworked in the redux settings in that case as my @TODO states

machard:
yes i think it needs to be reworked when doing the wyre integration.
And probably we should use navigation. since it will be on different pages

valpinkman (approved):

nabil-brn (approved):
Tested with all themes.

PR #3753, opened 2021-04-19
LL-5272 Remove the exchange wall when accessing swap

juan-cortes:
When accessing the swap feature we currently require users to have the exchange app installed and up to date and when inside the form we limit the selectors depending on whether they have the applications installed or not. This means we are capping the feature before users can access it. It has been agreed that having the "go to manager" ending _after_ the summary step would be better than before starting at all.

From this pr, we only check the IP when accessing from the sidebar, and the rest of the checks are moved to the modal flow.

Screenshot 2021-04-19 at 16 27 20
Screenshot 2021-04-19 at 16 25 24

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-5272

 - [ ] No apps installed whatosever, should break the flow for exchange, for currency 1 and 2.
 - [ ] Exchange app outdated, should break the flow with an outdated Exchange app
 - [ ] Missing/Outdated apps should also  break the flow with appropriate errors :trollface:


juan-cortes:
This will no longer be needed (as in an extra step in the modal flow) when https://github.com/LedgerHQ/ledger-live-common/pull/1156 is merged, since we can do the app requirement detection in any flow directly.

PR #3757, opened 2021-04-20
LL-4613 Inline installation on connectApp flows as experimental

juan-cortes:
https://ledgerhq.atlassian.net/browse/LL-4613

This functionality allows the user to install an application when, during a device flow, we detect that they don't have such application installed. This affects the Send, Receive, and Add account flows and enables the user to complete the flow without having to leave the modal/screen, go to the manager, install the application, and return.

This means that we can detach a bit more the user's awareness of what an app is, and what apps are installed on their device, minimise the friction on the flows too.

It's controlled by a new ENV variable `EXPERIMENTAL_INLINE_INSTALL` since we should thoroughly test this before releasing. It can also be enabled in the experimental section of the settings.

### How to test this
- All flows of the app should work with and without the experimental setting.
- When the experimental setting is ON, it should solve app dependencies in-flow
- When the experimental setting is OFF, it should refer the user to the manager

also includes

- https://github.com/LedgerHQ/ledger-live-common/releases/tag/v19.5.2
- https://github.com/LedgerHQ/ledger-live-common/releases/tag/v19.6.0


gre:
@juan-cortes you can now update the PR with bumping latest live-common

juan-cortes:
Sure, so we make it available on the experimental page then

gre:
@juan-cortes please update screenshot and we'll use this PR in prio to get live-common in

juan-cortes:
> @juan-cortes please update screenshot and we'll use this PR in prio to get live-common in

Should be good now

machard (approved):

Arnaud97234 (changes requested):
@juan-cortes https://ledgerhq.atlassian.net/browse/LL-5363

juan-cortes:
@Arnaud97234 this version should have your issue solved

Arnaud97234 (approved):
FYI: If there is no room enough on the device to install the required app, user lands on the following modal:
Screenshot 2021-04-23 at 18 37 06

We could imagine to have a better design / wording in the future.

juan-cortes:
That modal result when not enough space is expected @Arnaud97234 and we all agree it could have better UI for it.
If you have the flag disabled, this would also be the result, correct?

valpinkman (approved):

PR #3772, opened 2021-04-26
LL-5305 - add currencyName in Send and Receive flow

valpinkman:
As for addAccounts, adds currencyName into tracking event in Send and Receive flow

### Type

Improvement

### Context

LL-5305

Analytics

### DIFF
--- a/src/renderer/modals/Receive/Body.js
+++ b/src/renderer/modals/Receive/Body.js
@@ -9,6 +9,7 @@ import { SyncSkipUnderPriority } from "@ledgerhq/live-common/lib/bridge/react";
 import Track from "~/renderer/analytics/Track";
 import type { Account, TokenCurrency, AccountLike } from "@ledgerhq/live-common/lib/types";
 import type { Device } from "@ledgerhq/live-common/lib/hw/actions/types";
+import { getAccountCurrency } from "@ledgerhq/live-common/lib/account/helpers";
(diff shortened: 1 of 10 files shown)


src/renderer/modals/Receive/Body.js
```diff
@@ -175,6 +178,20 @@ const Body = ({
...
+
+      const currencyName = currency
+        ? currency.type === "TokenCurrency"
+          ? currency.parentCurrency.name
+          : currency.name
+        : undefined;
+
+      setCurrencyName(currencyName);
+    }
+  }, [account]);
```
gre, line 193: this shouldn't be an effect to me.

gre (changes requested):

Arnaud97234 (dismissed):
Event present on both Receive and Send flows.
No event on staking operations.
✅

src/renderer/modals/Receive/Body.js
valpinkman: fixed

juan-cortes (dismissed):
lgtm

Arnaud97234 (changes requested):
@valpinkman image

In Send / Receive flows currencyName should be the token name instead of the parent Currency name.

### How to reproduce
- Send / Receive ERC20

Arnaud97234 (approved):
Ok for Currency as well as tokens.
✅

JunichiSugiura (approved):

PR #3781, opened 2021-04-28
LL-5282 Provider integration

juan-cortes:
Pushing this early for @rjarasson-ledger to see if the 5c6a26b commit could be useful. Apparently we ended up doing the same work for the provider entry point. Keep in mind that due to 3e56184 there might be some changes that we don't want to be merged _yet_ for the swap modal flow, the relevant changes are all in /swap/providers and /swap/index


juan-cortes:
This is no longer needed, it will be part of a more up-to-date PR for the new wyre integration.

PR #3786, opened 2021-04-30
(Logger): fix electron import issue on renderer thread

LFBarreto:
(Logger): fix electron import issue on renderer thread
### Type

Bug fix

A dist build should now run properly

### DIFF
--- a/src/logger/logger.js
+++ b/src/logger/logger.js
@@ -1,11 +1,8 @@
 // @flow
-import { app } from "electron";
-import path from "path";
 import winston from "winston";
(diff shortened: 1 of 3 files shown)


src/logger/logger.js
```diff
@@ -25,25 +22,10 @@ const logger = winston.createLogger({
   transports,
 });
 
-const add = (transport: *) => {
+export const add = (transport: *) => {
```
gre, line 25: ( technically it's already exported on the default. but it's ok )

gre:
( part of live-common PR )

PR #3787, opened 2021-04-30
LL-5465 - try to check that builds are launched and running

machard:
https://ledgerhq.atlassian.net/browse/LL-5465


machard:
It does not want to launch in the ci :/

gre:
w'll have to come back to this.

PR #3802, opened 2021-05-04
issue-3740 (Terms): fix default en language issue

### LINKED ISSUE #3740: 404 from incorrect URL on Ledger Live Linux, prevents setting up new device

- tested on Ledger Live **2.25.1**
- Platform and version: **Arch Linux**

#### Expected behavior

App should contain correct URL and work.

#### Actual behavior

URL that is being requested is:

https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.en.md 

This doesn't exist, returns a 404, which should indicate that the URL is wrong. However, the application, oddly, interprets the 404 as meaning the user's internet connection is down, which leads to a bizarre error message for the user.

Try and set up a new Nano S on Linux. The above URL is obviously not the right one, this should be changed and care should be taken to not move files in a way that will break the app.

LFBarreto:
(Terms): fix default en language issue
### Type

Bug fix
### Context

Fixes #3740

https://ledgerhq.atlassian.net/browse/LL-6001

Terms loading should work properly now

### DIFF
--- a/src/renderer/terms.js
+++ b/src/renderer/terms.js
@@ -6,11 +6,17 @@ import { useSelector } from "react-redux";
 import { languageSelector } from "~/renderer/reducers/settings";
 import { urls } from "~/config/urls";
 
-const getRawLanguageURL = (language: string) =>
-  `https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.${language}.md`;
 const rawDefaultURL =
   "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERMS.md";
+
 export const url = "https://github.com/LedgerHQ/ledger-live-desktop/blob/master/TERMS.md";
+
+const termsUrlLocalized = {
+  en: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERMS.md",
+  fr: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.fr.md",
+  es: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.es.md",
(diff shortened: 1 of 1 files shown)


valpinkman (approved):

src/renderer/terms.js
```diff
@@ -6,12 +6,18 @@ import { useSelector } from "react-redux";
...
 const rawDefaultURL =
   "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERMS.md";
+
 export const url = "https://github.com/LedgerHQ/ledger-live-desktop/blob/master/TERMS.md";
 
+const termsUrlLocalized = {
+  en: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERMS.md",
+  fr: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.fr.md",
+  es: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.es.md",
+  ru: "https://raw.githubusercontent.com/LedgerHQ/ledger-live-desktop/master/TERM.ru.md",
```
gre, line 18: I'm not super fan of this idea that it won't scale each time we just add a TERM file. I actually wonder if we shouldn't have a "try load and fallback" approach instead.

it would be more future proof that any live version can be backward compatible to new TERM that would be added in future.

gre (changes requested):

src/renderer/terms.js
LFBarreto: the try load approach was the issue reported at first if i understood correctly :shrug: 
its still in place though i haven't changed it just made sure if we have the url ready we target it directly instead of 404 then going back to english.

gre (approved):

lkrust (approved):

gre:
unfortunately not mergeable.

PR #3805, opened 2021-05-05
LL-5438 - Fix select use case onboarding link

machard:
https://ledgerhq.atlassian.net/browse/LL-5438


LFBarreto (changes requested):
I dont think thats what needs to be done. The learn more actually opens a drawer where you should find a link, update it here
https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/components/Onboarding/Help/RecoveryWarning.js#L22

juan-cortes (dismissed):
If that's the right url, lgtm

LFBarreto (dismissed):

stual-ledger (changes requested):
Unfortunately, clicking this link still brings up a drawer on the side instead of the expected url (see screenshots below)

Reference # for the commit it was tested on : `6a784b4e982a2cfd1fd06c90e0b05ea3bde4fd1b`

Screenshot 2021-08-09 at 15 57 03

valpinkman (approved):

ychen-ledger (approved):
Validated with:

- MacOs big Sur 11.4
- Nano X / Nano S

PR #3814, opened 2021-05-07
Optimize usage of getCurrencyColor

gre:
after a performance audit, I see that the library "color" takes non negligible time on CPU. This adds some memo.


LFBarreto (dismissed):

gre:
included in #3815

PR #3836, opened 2021-05-15
fix animation glitch on slides  illustration

hayyaun:
slides were re-rendering on paused state changes in Carousel, which caused animations to start over by each mouse event in CarouselWrapper hoc, fixed by memorizing calculated slide components

### Type

UI Polish

Carousel Component


hayyaun:
Hi, It seems like there was a PR already fixed this issue.

PR #3840, opened 2021-05-18
LL-5505 (DiscreetMode): show fee values on send flow even on discreet mode

LFBarreto:
(DiscreetMode): show fee values on send flow even on discreet mode
### Type

Bug
### Context

[LL-5505]

Send flow with fees selection should now show the amount even with discreet mode on

[LL-5505]: https://ledgerhq.atlassian.net/browse/LL-5505

### DIFF
--- a/src/renderer/components/FormattedVal.js
+++ b/src/renderer/components/FormattedVal.js
@@ -60,6 +60,7 @@ type OwnProps = {
   ellipsis?: boolean,
   suffix?: string,
   showAllDigits?: boolean,
+  alwaysShowValue?: boolean, // overrides discreet mode
 };
 
 const mapStateToProps = createStructuredSelector({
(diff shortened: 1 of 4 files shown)


valpinkman (approved):

juan-cortes (approved):

Arnaud97234 (approved):
@LFBarreto Do we want to also permit the user to see the fees amount in the tx summary (step 3) ? if not just update the screenshots 🙏  then we can merge cc: @valpinkman

LFBarreto:
> @LFBarreto Do we want to also permit the user to see the fees amount in the tx summary (step 3) ? if not just update the screenshots  then we can merge cc: @valpinkman

if we want to update this it has to be done on live-common i think :thinking: not showing the amounts on the device validation step is a bit less critical since you can see the amounts on the device itself, whereas in the form you have no idea what you're selecting. We can merge this for now and loop back with product if they want to update for the summary as well afterwards

PR #3845, opened 2021-05-24
select fixed rate as default in swap screen

hzheng-ledger:
Jira ticket: https://ledgerhq.atlassian.net/browse/LL-5258


ghost (dismissed):

ghost (dismissed):

juan-cortes (changes requested):
I think you misunderstood the task. This is not about selecting the default trading method and certainly not about setting "fixed" as the default. We purposely moved to floating as a default a few months ago.

The task was merely UI, we want the disabled state of the trading methods to not show one of the methods as selected. Currently it looks like this
image

Note how Fixed is gray and float is white, we want both to be either white or gray (sync with @bscholving) 

This task is **not about choosing the default** unless I'm missing something.

Looping in @cthepot-ledger

cthepot-ledger:
@juan-cortes agreed with Juan, spec was poorly described my bad. the work is only UI and the goal is solely to change the inactive state of the toggle and to have Floating as active (in the inactive state only)

Arnaud97234:
@hzheng-ledger see juan last comment.

hzheng-ledger:
Fixed now
New UI:
Screenshot 2021-06-07 at 09 57 50
Screenshot 2021-06-07 at 09 55 43

hzheng-ledger:
@juan-cortes  @cthepot-ledger  Thanks for the feedback. Please review again :)

juan-cortes (dismissed):
lgtm

Arnaud97234 (dismissed):
QA OK.
✅ 

There is a small glitch just after selecting the currencies: It switch from fixed to floating.

Its not a blocker anyway If there is nothing we can do to improve this then we can merge.
ezgif com-gif-maker

gre (changes requested):
screenshots to regenerate and we should be good.

hzheng-ledger:
> QA OK.
> ✅
> 
> There is a small glitch just after selecting the currencies: It switch from fixed to floating.
> 
> Its not a blocker anyway If there is nothing we can do to improve this then we can merge.
> ezgif com-gif-maker

@Arnaud97234 Thanks for the feedback. I've corrected this small issue.

hzheng-ledger:
> screenshots to regenerate and we should be good.

@gre Screenshots updated. Thanks for the feedback

juan-cortes (approved):

PR #3853, opened 2021-05-25
LL-4831 improve logs parsing

valpinkman:
Take inspiration from live-common/cli/src/stream.js to parse in a safer way the log message we have incoming into LLD

### Type

Feature improvement

### Context

LL-4831

Logs

### DIFF
--- a/src/main/InternalProcess.js
+++ b/src/main/InternalProcess.js
@@ -107,13 +107,17 @@ class InternalProcess {
             if (count === 0) {
               acc += str.slice(lastIndex, i + 1);
               if (process.env.INTERNAL_LOGS) console.log(acc);
-              const obj = JSON.parse(acc);
-              if (obj && obj.type === "log") {
-                logger.onLog(obj.log);
-                return;
+              try {
+                const obj = JSON.parse(acc);
+                if (obj && obj.type === "log") {
+                  logger.onLog(obj.log);
+                  return;
+                }
+                logger.debug("I: " + acc);
+                acc = "";
+              } catch (error) {
+                if (process.env.INTERNAL_LOGS) console.error(error);
               }
-              logger.debug("I: " + acc);
-              acc = "";
             }
             break;
           default:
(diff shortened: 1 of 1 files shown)


LFBarreto (dismissed):
:exploding_head:

src/main/InternalProcess.js
```diff
@@ -87,22 +87,40 @@ class InternalProcess {
...
+              if (process.env.INTERNAL_LOGS) console.log(acc);
+              const obj = JSON.parse(acc);
+              if (obj && obj.type === "log") {
+                logger.onLog(obj.log);
+                return;
+              }
+              logger.debug("I: " + acc);
+              acc = "";
             }
-          } catch (e) {}
```
LFBarreto, line 102: I would keep the catch just in case. maybe the inner parsing can break at some point we never know :shrug:

LFBarreto (approved):

src/main/InternalProcess.js
```diff
@@ -87,22 +87,44 @@ class InternalProcess {
...
-          logger.debug("I: " + msg);
-        }),
-    );
+            break;
+          default:
+        }
+      }
+      acc += str.slice(lastIndex);
+    });
```
gre, line 127: I think we will have to challenge a lot this implementation regarding its performance as well as its reliability.
as it's meant to recreate json strings that are logged, we need for instance to see if an accidental `console.log` on internal side could be a breaking case of this. I feel we could have bad issue but i can't prove it much.

I think if we do a `console.log("{{{")` on internal side, it could be bad for instance, but need to see how bad it is.

gre:
@valpinkman if you don't mind i think we will HODL a bit this PR. i'm a bit scared on two parts:
- the performance
- the comment i mentioned regarding possible console.log issues that internal might have (or any lib of internal might use)

gre:
i judge this too risky. we'll have to "live with the problem" and possibly simplify the logs to avoid the problem of buffering.

PR #3854, opened 2021-05-25
LL-5280  Swap with Wyre

juan-cortes:
### Description
This pr also includes a rework on the app state regarding swap making it more manageable as well as general sync between LLD and LLM naming and component logic, reusing as much as possible unlike the previous versions (my bad).

It is rebased with develop as of 28th of Jun, which includes the drop of the providers' selector screen. Artifacts from this pr should be testable, but require a few env to work. First of all, we need to launch the app pointing to staging v3 `SWAP_API_BASE=https://swap-stg.ledger.com/v3` and in case we want to disable a provider (it will fallback to changelly if both are available) we have a new env `SWAP_DISABLED_PROVIDERS` which can be used to disable one or more providers.

For instance, if we launch with `SWAP_DISABLED_PROVIDERS="wyre,changelly"` we can test the no providers case, which should trigger the "service unavailable" screen.

### How to test invalid KYC id
- After completing a valid KYC flow edit your app.json data and alter the user id for Wyre, it starts with "US_.." change it to "_US.." or whatever
- Open the swap feature and try to request a rate
- A modal should appear, offering the option of resetting KYC
   - If accepted you should be kicked to the KYC form
   - If rejected, the swap form should be disabled and there should be a CTA to reopen the modal

### How to test latest firmware requirement
- Launch the app and start a swap request
- Connect a LNX/LNS that is **not** on the latest version
- Check if you get an error at the beginning of the device action, redirecting you to the manager
### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-5280


cthepot-ledger:
- [ ] @juan-cortes to add LL-5954

src/renderer/modals/Swap/steps/StepDevice.js
```diff
@@ -150,6 +151,7 @@ const StepDevice = ({
         account,
         transaction: swapData.transaction,
         appName: "Exchange",
+        requireLatestFirmware: true,
```
juan-cortes, line 154: we already know it at this point, no?

LFBarreto (approved):

LFBarreto (dismissed):

LFBarreto (approved):
:rocket:

gre:
good to go. i'll merge the live-common activation on a separate PR .

PR #3860, opened 2021-05-27
LL-4998 [Polkadot] Add a flow to change controller

ghost:
In the specific situation where the current account is a stash with a different controller, provide a button to the user to set the account as its own controller.

**Requires:** https://github.com/LedgerHQ/ledger-live-common/pull/1217
**Requires:** minimum Nano app v7.9050.0

**How to test:** from a new account, do a first bond that uses a different account as controller (can only be done using https://polkadot.js.org/apps)

Screenshot from 2021-05-27 16-20-53


Arnaud97234 (changes requested):
Operation type needs to be mapped:
image

**How to reproduce**
- Operation history

gre (approved):

gre (changes requested):
unfortunately there are changed in the **de** language which is forbidden as only English is supposed to be changed.

gre (approved):

gre:
We'll take this required work in to be needed for LL-6872

Issue #3794, opened 2021-05-01
English Grammar Issue Discovered in Ledger Live Lend


Maxosphere:
#### Ledger Live Version and Operating System
Latest Ledger Live Version 2:35PM CST 5/1/2021 (USA Central Time)
Windows Operating System

- tested on Ledger Live **2.26.1**
- Platform and version: **Windows 10**

#### Expected behavior

The feature should read, "You don't have **a** DAI Stablecoin v2.0 account"

#### Actual behavior

The feature reads, "You don't have **any** DAI Stablecoin v2.0 account"

#### Steps to reproduce the behavior

Open Ledger Live > Lend > Click Lend Button next to a lending asset.
This issue exists in all lending assets available.
![Screenshot 2021-05-01 144735](https://user-images.githubusercontent.com/83515398/116793437-5d285600-aa8c-11eb-8665-9a96d014e842.png)

gre:
Fixed in [2.29.0](https://github.com/LedgerHQ/ledger-live-desktop/releases/tag/v2.29.0)

gre:
oh sorry actually not yet on LLD.

PR #3862, opened 2021-05-28
LL-5497 Grammar fix

MortalKastor:
Grammar fix

### Context

#3794 
https://ledgerhq.atlassian.net/browse/LL-5497

### Parts of the app affected

Wording in _Lend_ feature, when there's no matching account

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -700,7 +700,7 @@
       "cta": "Add account"
     },
     "emptyAccountDeposit": {
-      "title": "You don't have any {{ asset }} account.",
+      "title": "You don't have a {{ asset }} account.",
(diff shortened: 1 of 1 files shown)


haammar-ledger:
Nitpicking here, but this won't fit some currencies starting with a vowel... eg. "_you don't have a ALGO account_"

valpinkman (approved):

juan-cortes (changes requested):
I'm going to be a pita but what happens when you don't have ETH, or ATOM, or Algorand?
That's also wrong grammar. We can't just say "You don't have a Algorand account." 
Maybe "You don't have any Algorand account**s**" would work best

gre:
for now we merge to be 1:1 with mobile. but plz raise to product initial task owner @juan-cortes

PR #3870, opened 2021-05-31
skip the currency selection step when it not not necessary

hzheng-ledger:
Jira ticket: https://ledgerhq.atlassian.net/browse/LL-5711

### DIFF
--- a/src/renderer/modals/AddAccounts/index.js
+++ b/src/renderer/modals/AddAccounts/index.js
@@ -61,6 +61,7 @@ export type StepProps = {
 type St = Step<StepId, StepProps>;
 
 const createSteps = (skipChooseCurrencyStep): St[] => {
+  // the back button is not needed when we skip "chooseCurrency" step because the back button brings user to "chooseCurrency" step
   const onBack = skipChooseCurrencyStep
     ? null
     : ({ transitionTo, resetScanState }: StepProps) => {
--- a/src/renderer/modals/AddAccounts/index.js
+++ b/src/renderer/modals/AddAccounts/index.js
@@ -30,6 +30,7 @@ type Props = {
   closeModal: string => void,
   replaceAccounts: (Account[]) => void,
   blacklistedTokenIds?: string[],
+  skipChooseCurrencyStep?: ?boolean,
 };
 
 type StepId = "chooseCurrency" | "connectDevice" | "import" | "finish";
--- a/src/renderer/modals/AddAccounts/index.js
+++ b/src/renderer/modals/AddAccounts/index.js
@@ -59,12 +59,14 @@ export type StepProps = {
 
 type St = Step<StepId, StepProps>;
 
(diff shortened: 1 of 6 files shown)


hzheng-ledger:
The jira ticket only requires that we skip the first step in the "add account" modal from manager screen.
I find that it is also useful to do that in other situations when the first step (currency selection step) is not necessary

src/renderer/modals/AddAccounts/index.js
```diff
@@ -243,6 +253,10 @@ class AddAccounts extends PureComponent<Props, State> {
 
     const errorSteps = err ? [2] : [];
 
+    if (stepId === "chooseCurrency" && this.props.skipChooseCurrencyStep) {
```
juan-cortes, line 256: Why would we reach this case? Aren't we shifting the steps when we have the `skipChooseCurrencyStep` flag in the first place?

src/renderer/modals/AddAccounts/steps/StepConnectDevice.js
```diff
@@ -30,6 +27,9 @@ const StepConnectDevice = ({ currency, device, transitionTo }: StepProps) => {
       ? currency.parentCurrency.name
       : currency.name
     : undefined;
+  if (!currency) {
```
juan-cortes, line 30: Why do we need to remove the invariant? I feel we should never reach this step without a currency at all 🤔

juan-cortes (changes requested):

src/renderer/modals/AddAccounts/index.js
hzheng-ledger: yes, we reach this case. 
In fact, we have this: https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/modals/AddAccounts/index.js#L126
So the stepId is set to "chooseCurrency" initially. 
I can either set the initial stepid = connectDevice (if skipChooseCurrencyStep == true)
or do the way what I did.

src/renderer/modals/AddAccounts/steps/StepConnectDevice.js
hzheng-ledger: In theory, we never reach this step.
but in fact,
I found that the code in this component is triggered even if skipChooseCurrencyStep== true
because of we have this component in the following line:
https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/modals/AddAccounts/index.js#L69
I don't know how react handles this and why react reach inside the StepChooseCurrency component even if I put "steps.shift();" when skipChooseCurrencyStep== true
tell me if you have a better idea. Thanks

gre (changes requested):
please merge develop in so CI can pass.

hzheng-ledger:
replace by https://github.com/LedgerHQ/ledger-live-desktop/pull/4117

PR #3875, opened 2021-06-01
LL-3084 Prevent Crypto assets without any settings to be selected in account settings page

JunichiSugiura:
### Type

UI Polish

### Context

Currently, the first account of user is automatically selected as default value in crypto asset selectbox unless user has bitcoin account. It could be a bit annoying when asset wtihout any setting is selected or when user select specific asset and realized it doesn't have any settings. This PR removes default values and uses placeholders instead as a default. It also prevent users from selecting assets without settings by disabling option. (User can still see these asset options.)

Screen Shot 2021-06-01 at 12 35 15
IMG_1694


juan-cortes (dismissed):

valpinkman (approved):

gre (changes requested):
this should not have ✅  because CI fails and lint too.

src/renderer/screens/settings/sections/Accounts/Currencies.js
```diff
@@ -2,34 +2,40 @@
...
 
-  return !currency ? null : (
-    <Box key={currency.id}>
-      <TrackPage category="Settings" name="Currencies" currencyId={currency.id} />
-      <Track onUpdate event="Crypto asset settings dropdown" currencyName={currency.name} />
+  const currencyId = currency?.id ?? "placeholder";
+  const currencyName = currency?.name ?? "placeholder";
+
+  const isCurrencyDisabled = useCallback(
+    (currency: Currency) => !currencySettingsDefaults(currency).confirmationsNb,
```
LFBarreto, line 31: maybe check for at least a key in the object maybe there are currencies with a single setting that isn't confirmationNb :thinking:

src/renderer/screens/settings/sections/Accounts/Currencies.js
```diff
@@ -2,34 +2,40 @@
...
     },
     [setCurrency],
   );
 
-  return !currency ? null : (
-    <Box key={currency.id}>
-      <TrackPage category="Settings" name="Currencies" currencyId={currency.id} />
-      <Track onUpdate event="Crypto asset settings dropdown" currencyName={currency.name} />
+  const currencyId = currency?.id ?? "placeholder";
+  const currencyName = currency?.name ?? "placeholder";
```
LFBarreto, line 28: why even bother with a placeholder in this case? 
maybe dont add the TrackPage and Track in the template if currency isn't set :shrug:

LFBarreto (changes requested):

src/renderer/screens/settings/sections/Accounts/Currencies.js
JunichiSugiura: I've used the current logic here.

https://github.com/LedgerHQ/ledger-live-desktop/blob/7639b7ed0684d1883953ea931114dc19cead72a1/src/renderer/screens/settings/sections/Accounts/CurrencyRows.js#L69

juan-cortes: That's because we currently only show the settings for number of confirmations, we used to have other settings such as the endpoint for ripple but we don't anymore. Quoting the comment in the same linked file from @JunichiSugiura 
> // NB ideally we would have a dynamic list of settings
But we don't, so for now I think it's ok

juan-cortes (approved):

LFBarreto (approved):

valpinkman (approved):

PR #3876, opened 2021-06-01
Update live-common with latest optimisations

gre:
Impact:
- the logic to calculate if an app is installable was optimized. (the button install that goes grey)
- the progress % is restored for installs.


ghost (approved):

Arnaud97234 (approved):
- Install button becomes grey for applications that can't be installed
- Progress bar when installing apps is back

valpinkman (approved):

PR #3877, opened 2021-06-01
LL-5325 - add logs

machard:
needs LedgerHQ/ledger-live-common#1226

https://ledgerhq.atlassian.net/browse/LL-5325

test plan;
2 differents kind of transaction. One that use the bridge proxy broadcast and one that use useBroadcast

### DIFF
--- a/src/renderer/hooks/useBroadcast.js
+++ b/src/renderer/hooks/useBroadcast.js
@@ -1,7 +1,8 @@
 // @flow
 import invariant from "invariant";
 import { useCallback } from "react";
-import { getMainAccount } from "@ledgerhq/live-common/lib/account";
+import { log } from "@ledgerhq/logs";
+import { getMainAccount, formatOperation } from "@ledgerhq/live-common/lib/account";
 import type {
   SignedOperation,
   Operation,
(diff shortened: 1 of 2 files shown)


valpinkman (dismissed):

gre (approved):

valpinkman (dismissed):

src/renderer/hooks/useBroadcast.js
```diff
@@ -28,12 +29,17 @@ export const useBroadcast = ({ account, parentAccount }: SignTransactionArgs) =>
...
           account: mainAccount,
           signedOperation,
-        }),
-      );
+        });
+        log(
+          "transaction-summary",
+          `✔️ broadcasted! optimistic operation: ${formatOperation(mainAccount)(operation)}`,
+        );
+        return operation;
```
gre, line 41: I think it's already covered inside `bridge.broadcast` because it in fact use the "proxy commands" implicitly

could you check @machard ?

i suspect we have a double log happening.

valpinkman (approved):

ychen-ledger (approved):
Discussed with @machard CC @gre 
Validated with: 

- Swap and Send TX section as Send/ MacOs Big Sur  version 11.4/ledger live 2.32.2
- For missing part like how long the TX keep track by logs could be implement in a new ticket

juan-cortes (approved):

PR #3880, opened 2021-06-03
LL-5492 - dont display app install success if the user already has accounts

machard:
https://ledgerhq.atlassian.net/browse/LL-5492

### DIFF
--- a/src/renderer/screens/manager/AppsList/AppsList.js
+++ b/src/renderer/screens/manager/AppsList/AppsList.js
@@ -1,7 +1,7 @@
 // @flow
 import React, { useState, memo, useCallback, useEffect, useRef } from "react";
 import { useLocation, useHistory } from "react-router-dom";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import styled from "styled-components";
 import { Trans } from "react-i18next";
 
@@ -10,6 +10,7 @@ import { useAppsSections } from "@ledgerhq/live-common/lib/apps/react";
 import type { TFunction } from "react-i18next";
 import type { DeviceInfo } from "@ledgerhq/live-common/lib/types/manager";
 import type { State, Action, AppsDistribution } from "@ledgerhq/live-common/lib/apps/types";
+import { hasAccountsSelector } from "~/renderer/reducers/accounts";
 import UpdateAllApps from "./UpdateAllApps";
 import Placeholder from "./Placeholder";
 import Card from "~/renderer/components/Box/Card";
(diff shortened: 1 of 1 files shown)


src/renderer/screens/manager/AppsList/AppsList.js
```diff
@@ -154,7 +156,7 @@ const AppsList = ({
         dispatch={dispatch}
         isIncomplete={isIncomplete}
         addAccount={addAccount}
-        disabled={update.length >= 1}
+        disabled={update.length >= 1 || hasAccounts}
```
juan-cortes, line 159: This also has the side effect of not showing the install success banner after I install ETH if I already have BTC accounts for instance, I don't know if this is something we want from product's perspective. To me the story reads as _"don't show the success banner _if coming from a fw update reinstall_"_ which is different. 🤔 

To me it should still show (once per manager session) after a successful install of a supported app.
cc @cthepot-ledger

juan-cortes (changes requested):

src/renderer/screens/manager/AppsList/AppsList.js
LFBarreto: yep you need to create a `hasCurrencyAccounts` of use the `currenciesSelector` to compare the list you get with what you installed :thinking:

LFBarreto (changes requested):

src/renderer/screens/manager/AppsList/AppsList.js
```diff
@@ -83,6 +85,10 @@ const AppsList = ({
   const { push } = useHistory();
   const { search } = useLocation();
   const reduxDispatch = useDispatch();
+  const currenciesAccountsSetup = useSelector(currenciesSelector);
+  const disableInstallBanner = every(state.installed, app =>
+    some(currenciesAccountsSetup, currency => currency.id === state.appByName[app.name].currencyId),
```
LFBarreto, line 90: should we disable if you already added one account from the installed list? :thinking:

LFBarreto (changes requested):

LFBarreto (approved):
:+1:

Arnaud97234 (changes requested):
@machard see my comment in https://ledgerhq.atlassian.net/browse/LL-5492

LFBarreto (approved):

Arnaud97234 (changes requested):
@machard check my comment on https://ledgerhq.atlassian.net/browse/LL-5492?focusedCommentId=115287
The banner is still present after reinstalling the nano apps at the end of the firmware update

LFBarreto (approved):

stual-ledger (approved):
🚀 lgtm

PR #3881, opened 2021-06-03
Make CI build config even with release one (minus signing)

MortalKastor:
Fix Windows CI build missing graphical assets.

### Type

Bug fix

### Context

Switching to CI for Windows release builds (minus the signing part)

### Parts of the app affected

CI build

### Test plan

- Make sure the CI doesn't break
- Verify that the installer now show Ledger images rather than generic ones

### DIFF
--- a/electron-builder-nightly.yml
+++ b/electron-builder-nightly.yml
@@ -21,7 +21,7 @@ linux:
 
 win:
   artifactName: ${name}-${version}-${os}-${arch}.${ext}
-  # icon: build/windows/app.ico
+  icon: build/windows/app.ico
   verifyUpdateCodeSignature: false
   signAndEditExecutable: false
   target:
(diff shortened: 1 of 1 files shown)


electron-builder-nightly.yml
```diff
@@ -33,11 +33,13 @@ nsis:
...
+  installerIcon: build/windows/installer.ico
+  installerSidebar: build/windows/installerSidebar.bmp
+  uninstallerIcon: build/windows/uninstaller.ico
+  uninstallerSidebar: build/windows/uninstallerSidebar.bmp
 
 files:
   # Include files
   - .webpack/**/*
-  - node_modules/**/*
-  - package.json
```
MortalKastor, line 41: While unrelated to the Windows CI build fix, these two lines are not in the release config, so I removed them too.
It shouldn't change anything as `electron-builder` includes both by default so they don't need to be listed AFAIK.

PR #3888, opened 2021-06-04
LL-5833 - reconcile user ids

machard:
https://ledgerhq.atlassian.net/browse/LL-5833

### DIFF
--- a/src/renderer/init.js
+++ b/src/renderer/init.js
@@ -70,12 +70,12 @@ async function init() {
     global.localStorage.removeItem("userId");
   }
 
-  let userId = global.localStorage.getItem("userId");
-  if (!userId) {
-    userId = uuid();
-    setKey("app", "user", userId);
+  let user = await getKey("app", "user");
+  if (!user) {
+    user = { id: uuid() };
+    setKey("app", "user", user);
   }
-  setEnvOnAllThreads("USER_ID", userId);
+  setEnvOnAllThreads("USER_ID", user.id);
 
   setBridgeProxy({ getAccountBridge, getCurrencyBridge });
 
--- a/src/renderer/init.js
+++ b/src/renderer/init.js
@@ -12,19 +12,23 @@ import { remote, webFrame, ipcRenderer } from "electron";
 import { render } from "react-dom";
 import moment from "moment";
 import _ from "lodash";
-import { reload, getKey, loadLSS } from "~/renderer/storage";
+import { reload, getKey, setKey, loadLSS } from "~/renderer/storage";
 import { hardReset } from "~/renderer/reset";
 
+import analytics from "~/renderer/middlewares/analytics";
 import "~/renderer/styles/global";
 import "~/renderer/live-common-setup";
 import { getLocalStorageEnvs } from "~/renderer/experimental";
 import "~/renderer/i18n/init";
+import { setBridgeProxy } from "@ledgerhq/live-common/lib/bridge";
+import { getAccountBridge, getCurrencyBridge } from "./bridge/proxy";
+import { setEnvOnAllThreads } from "./../helpers/env";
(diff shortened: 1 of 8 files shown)


src/renderer/init.js
```diff
@@ -61,6 +65,20 @@ async function init() {
     connect,
   });
 
+  const oldUserId = global.localStorage.getItem("userId");
+  if (oldUserId) {
+    global.localStorage.removeItem("userId");
```
gre, line 70: imo this needs to be moved only after the get + set so make sure it's "atomic". otherwise if user kill the live before it ends being set, it would fail.

src/renderer/init.js
```diff
@@ -61,6 +65,20 @@ async function init() {
...
+  }
+
+  let user = await getKey("app", "user");
+  if (!user) {
+    user = { id: uuid() };
+    setKey("app", "user", user);
```
gre, line 76: i think an await is needed on setKey

src/renderer/init.js
```diff
@@ -84,7 +102,14 @@ async function init() {
     await hardReset();
   }
 
-  const store = createStore({ dbMiddleware });
+  const store = createStore({
+    dbMiddleware,
+    analyticsMiddleware: analytics({
+      migration: {
+        ...(oldUserId ? { oldUserId } : {}),
+      },
```
gre, line 110: probably can just be `migration: oldUserId ? { oldUserId } : null`

src/renderer/middlewares/analytics.js
```diff
@@ -1,18 +1,22 @@
...
+export default ({ migration }: { migration?: any }) => (store: *) => (next: *) => (action: *) => {
   next(action);
   const state: State = store.getState();
   const hasCompletedOnboarding = hasCompletedOnboardingSelector(state);
 
   if (hasCompletedOnboarding && !isAnalyticsStarted) {
     isAnalyticsStarted = true;
     start(store);
+
+    if (Object.keys(migration || {}).length) {
```
gre, line 18: if (migration)

gre (commented):
approach looks viable. some small notes.

src/renderer/init.js
machard: https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/storage.js#L59 looks ok

src/renderer/init.js
machard: https://github.com/LedgerHQ/ledger-live-desktop/pull/3888#discussion_r645625073 problem is we don't know when the setKey is actually done, there a debounce. or i change the setkey to be async or add additional callback ?
However i think its a very slight possibility that the user has time to close just in between

src/renderer/init.js
machard: yes, i was thinking if we use this migration thing with other stuff in the futur

src/renderer/middlewares/analytics.js
machard: https://github.com/LedgerHQ/ledger-live-desktop/pull/3888#discussion_r645643524

src/renderer/init.js
gre: but we don't so i would not anticipate something we don't need now.

David-0077 (approved):

valpinkman (approved):

src/renderer/init.js
gre: so clearly to me this is a risk that worth a fix. we can't afford possibly losing the userId.

gre (changes requested):
actuzlly we have comments that raised a problem of possible loss of userId so it's not a safe merge until we fix it.

cthepot-ledger:
status on this?

PR #3895, opened 2021-06-08
LL-5455 (Swap): lottie animation updated

LFBarreto:
(Swap): lottie animation updated
### Type

UI Polish
### Context

[LL-5455]

Swap double arrow icon updated

[LL-5455]: https://ledgerhq.atlassian.net/browse/LL-5455


juan-cortes:
I don't think it's a good idea to replace the current animation with this one as it stands, it's too slow, too small, and too janky. By the time that animation ends the user can go grab a coffee from the kitchen. I understand the idea of not using an icon that we didn't design ourselves, although we do that pretty much everywhere in the app, if the result is not as good as the original one I don't see the point in moving in that direction.

https://user-images.githubusercontent.com/4631227/121226623-10098200-c88b-11eb-8d20-f2522e06a075.mov
https://user-images.githubusercontent.com/4631227/121226461-e0f31080-c88a-11eb-87fc-45aa3cbd624c.mov

cthepot-ledger:
i agree with @juan-cortes

PR #3897, opened 2021-06-09
Bump beta version of ledgerjs in order to validate TypeScript works

gre:
https://ledgerhq.atlassian.net/browse/LL-3928


gre:
for this PR. we will not merge it as is, we just need a quick check that USB works ok on it as well as one coin of each family:
- Bitcoin
- Ethereum (e.g. an erc20)
- xrp


that gives a confidence that we did good in our typescript migration. we can then proceed to include it in a future live-common bump.

gre:
thx for QA we noticed issues that i logged in https://github.com/LedgerHQ/ledgerjs/pull/619#pullrequestreview-680657118

PR #3898, opened 2021-06-10
Update config.json

gre:
enable release path to 2.29.0 for 10% users.


IAmMorrow (approved):
Reviewed mechanics tied to this as well, looks like it's still valid.

PR #3906, opened 2021-06-14
LL-5904 (Swap) Remove the bifurcation screen

juan-cortes:
Should be a good starting point for both the catalog feat and the rework for wyre swap.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-5904

- Access swap, there should be no questions, just form (loading/ip/etc) for changelly


rjarasson-ledger:
Included and superseded by https://github.com/LedgerHQ/ledger-live-desktop/pull/3922

PR #3911, opened 2021-06-15
LL-5450 upgrading to node 14

valpinkman:
This PR bumps libcore to use the new node 14 bindings.
Linked to https://github.com/LedgerHQ/ledger-live-common/pull/1261

### Type

Chore

### Context

LL-5450

The whole app. We need to see if upgrading to node 14 doesn't break anything. 
As of now I have tested to add accounts and sync and I did not see any issues, but extensive QA is required.

To all devs, whenever this PR is merge, bump to node 14

### DIFF
--- a/.github/workflows/build-linux-prod.yml
+++ b/.github/workflows/build-linux-prod.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'Tag?'
+        description: "Tag?"
         required: true
 
 jobs:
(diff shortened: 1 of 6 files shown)


gre:
algo i think it's wise to test this by also using Node 14 on the yarn etc.. so we test the whole node 14 env package (not mandatory but good)

gre:
i'll include this in live-common bump to converge the QA effort.

Hbutlercapone (commented):
lets try it

PR #3923, opened 2021-06-21
[hackathon-magic-accounts] just here for ci builds

juan-cortes:


gre:
@juan-cortes please do this on your fork directly. you should be able to produce bundles directly in your fork i think no?

PR #3698, opened 2021-03-24
Ll 4555 - Wallet connect

machard:
need https://github.com/LedgerHQ/ledger-live-common/pull/1127

PR #3924, opened 2021-06-21
LL-5978

valpinkman:
- setup to fix flaky tests
- block sync

### Type

TEST: features update

### Context

LL-5978

requires https://github.com/LedgerHQ/ledger-live-common/pull/1272

⚠️  It might break all the current screens if they use to have a notification update ⚠️ 

#### SYNC_ALL_INTERVAL
As of now, we use the env `SYNC_ALL_INTERVAL` with a crazy high number to basically simulate a sync.stop()
We will use the method `synchronize` available on each PageObject

#### Service Statuses and Notifications
I have added a couple of method on the base `Page` object (which means it's available on all PageObject)

- **Page.addAnnouncement()** => look into a pool of mock announcements push the first one or generate if no more mocks are left, then triggers a sync
- **Page.generateAnnouncement(params?)** => generate a new announcement (possible to pass params to it), then triggers a sync
- **Page.resetAnnouncements()** => resets announcements and announcements pool lists (for next tests for example)
- **Page.toggleIncident(bool)** => toggles on or off a mock list of incidents, then triggers a sync
- **Page.resetIncidents()** => resets incidents (for next tests for example)


machard:
I think its better to wait for llc to be merged and finish in this PR.
there is some part that are missing to make it work.
the mock lives in docker. jest lives in the host. see https://github.com/LedgerHQ/ledger-live-desktop/pull/3698/files#diff-91c95d7d5c7dd9da02dec2de6d64882760a5b353345b6e5d19e55dfcae074104R16 and https://github.com/LedgerHQ/ledger-live-desktop/pull/3698/files#diff-969401d244c0d718e2b6952b07ac4133aaa2c2ed60d83441d41db8ea79133837R55 for comms

PR #3932, opened 2021-06-22
Hackathon nft

gre:


gre:
[3932.diff.txt](https://github.com/LedgerHQ/ledger-live-desktop/files/6704022/3932.diff.txt)

diff for record. closing now.

PR #3943, opened 2021-06-28
LL-5453 Electron 12 upgrade

LFBarreto:
Electron 12 upgrade.
- in order to maintain node integration in renderer we need to disable contextIsolation
- migrated deprecated electron.remote usage in favor of @electron/remote as per upgrade docs
### Type

Feature
### Context

[LL-5453]

All of the app, try and run it, lock the json data with password, do some usb actions with a device. A general pass on the app should be enough.

### DIFF
--- a/src/main/window-lifecycle.js
+++ b/src/main/window-lifecycle.js
@@ -55,6 +55,7 @@ const defaultWindowOptions = {
     devTools: __DEV__ || DEV_TOOLS,
     experimentalFeatures: true,
     nodeIntegration: true,
+    contextIsolation: false,
   },
 };
 
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "generate-screenshots": "PWD=$(pwd); cd docker; docker run -v \"$PWD:/github/workspace\" -w \"/github/workspace\" --rm -it $(docker build -q .)"
   },
   "dependencies": {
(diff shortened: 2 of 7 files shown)


src/main/window-lifecycle.js
```diff
@@ -55,6 +55,7 @@ const defaultWindowOptions = {
     devTools: __DEV__ || DEV_TOOLS,
     experimentalFeatures: true,
     nodeIntegration: true,
+    contextIsolation: false,
```
valpinkman, line 58: unfortunately

valpinkman (approved):

src/main/window-lifecycle.js
juan-cortes: Making `contextIsolation` false breaks the security model if we import external scripts from what I understand. Gab pointed out in the context of the chatbot poc https://ledgerhq.atlassian.net/browse/LL-2349 that without it scripts could go rogue on our app, I'm scared of this change 😕

LFBarreto: Its either that or we wait for upgrade to deactivate nodeIntegration that requires this :disappointed:

juan-cortes: I lack the knowledge of whether this is dangerous or not so I'd rather not approve

PR #3944, opened 2021-06-28
LL-5533 update sentry url?

valpinkman:
### Type

FIX

### Context

LL-5533

Let's be honest, I don't know how we can test this. I think I fixed the url by going into the sentry settings and copying the url but still, I couldn't find to trigger an event for Sentry

Recommendation => move to the new sdk and not use raven-js anymore. Pain point: anonymizing the data


gre (approved):
I've been able to confirm this works. We don't need to QA this it's purely a maintainer topic. merging.

PR #3948, opened 2021-06-30
LL-6086 add tracking to view in explorer

valpinkman:
I readded the ability to track events using `openURL` (the API was there in v1 but it seems like we forgot to resume it on v2)
Then I added the tracking event to the View in Explorer

### Type

tracking improvement

### Context

LL-6086

View Operation / View Swap Operation


LFBarreto (approved):
:ok_hand:

lruggiero-ledger (approved):

JunichiSugiura (approved):

PR #3954, opened 2021-07-02
Live desktop libcore 4.2

hzheng-ledger:


gre:
closing for now as the focus is more on the rework at this moment. good to know at least it's going to be a simple bump if we go to that plan in future.

PR #3955, opened 2021-07-02
test focus instead of click

valpinkman:


juan-cortes (approved):
Seems to work, no? 🚀

valpinkman:
> Seems to work, no? 🚀

thanks but still working on it

PR #3969, opened 2021-07-08
[Ll-6102] Add twitter banner

achabrolin-ledger:
Add a dismissible Twitter banner with clickable hashtag in the Live Catalog page.

Ll-6102-twitter-banner-preview

Ll-6102-twitter-message

### Type

Feature / UI Polish

### Context

https://ledgerhq.atlassian.net/browse/LL-6102

The part relating to the "static UI element with the TWITTER hashtag"

#### Which part of the app is affected?

The Live Apps Catalog

#### What to do to test it, any special thing to do?

You should be able to see the banner, dismiss it with the cross and click on the hashtag to open twitter and (if you are logged in), start to compose a pre-filled tweet.

Once the banner is dismissed, it can be enabled again by removing the `platform-twitter-notices` from the `dismissedBanners` array in the `app.js` file located in the user data folder (cf. https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/1392869669/Location+of+user+data+folder)


src/renderer/screens/platform/TwitterBanner.js
```diff
@@ -0,0 +1,43 @@
...
+    text: t("platform.catalog.twitterBanner.tweetText"),
+    hashtags: "LedgerLiveApp",
+  };
+
+  const paramsURI = Object.entries(params).reduce(
+    (acc, [key, val]) => (acc += `&${key}=${encodeURI(val)}`),
+    "",
+  );
+
+  const url = `https://twitter.com/intent/tweet?${paramsURI}`;
```
IAmMorrow, line 26: That works ! I see two easy improvements ! :) 

You could use the URL api to handle the query params ina mor
You could wrap the whole thing in a useMemo to prevent recalculation at every render !

Here's an extract of a similar use case from the same codebase:

`  const url = useMemo(() => {
    const urlObj = new URL(manifest.url.toString());

    urlObj.searchParams.set("backgroundColor", theme.background.paper);
    urlObj.searchParams.set("textColor", theme.text.shade100);

    return urlObj;
  }, [manifest.url, theme]);
`

IAmMorrow (changes requested):
Extending the Alert with a "twitter" variant is a good idea, that can be of good use later.
Added a small comment but otherwise good first contribution ! 🎉

IAmMorrow:
Looks like Flow is sad ! Try running `yarn flow` to find out why !

achabrolin-ledger:
> Looks like Flow is sad ! Try running `yarn flow` to find out why !

Indeed, it looks like I didn't run `yarn flow` before committing.

We could use git hooks and husky to run (at least) these lint and format commands automatically before committing. This could prevent this kind of errors in the future.
cf.
- https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
- https://typicode.github.io/husky/#/

IAmMorrow (approved):

src/renderer/screens/platform/TwitterBanner.js
```diff
@@ -0,0 +1,44 @@
...
+  mb: 24,
+}))``;
+
+const twitterHashtag = "LedgerLiveApp";
+
+export default function TwitterBanner() {
+  const { t } = useTranslation();
+
+  const url = useMemo(() => {
+    const urlObj = new URL("https://twitter.com/intent/tweet");
```
juan-cortes, line 21: Move this to `config/urls.js` so that it's not hardcoded here

juan-cortes (changes requested):

src/renderer/screens/platform/TwitterBanner.js
achabrolin-ledger: Thanks for the feedback, it has been updated here d95d571

gre (approved):

ychen-ledger (approved):
Validated with: 

- MacOS BigSur version 11.4 
- FW 1.3.0
- Ledger Live 2.30.0

PR #3978, opened 2021-07-12
[LL-6363] add platform app info panel

achabrolin-ledger:
Add a button that open a side drawer featuring generic platform app information
cf. https://ledgerhq.atlassian.net/browse/LL-6363

<details>
  <summary>Desktop 📸 </summary>

Screenshot

</details>

### Type

Feature

_Platform catalog_

Once in a "service" or "app" in the "discover" page, open an app and click the information button on the top right corner. A side drawer with informations relative to the selected app are displayed, including title, description and a link to the service website. Clicking on this link should open the service webpage.

### DIFF
--- a/src/renderer/components/SideDrawer.js
+++ b/src/renderer/components/SideDrawer.js
@@ -203,7 +203,7 @@ export function SideDrawer({
       {state => (
         <DrawerContainer className="sidedrawer" state={state} ref={focusTrapElem} tabIndex="-1">
           <DrawerContent {...props} isOpened={isOpen} state={state} direction={direction}>
-            {onRequestClose || onRequestBack ? (
+            {onRequestClose || onRequestBack || title ? (
               <Box
                 horizontal
                 justifyContent="space-between"
@@ -230,10 +230,12 @@ export function SideDrawer({
                   </Text>
                 )}
 
-                {onRequestClose && (
+                {onRequestClose ? (
                   <TouchButton onClick={onRequestClose} className="sidedrawer-close">
                     <IconCross size={16} />
                   </TouchButton>
+                ) : (
+                  <Box />
                 )}
               </Box>
             ) : null}
--- a/src/renderer/components/SideDrawer.js
+++ b/src/renderer/components/SideDrawer.js
@@ -116,6 +116,7 @@ type DrawerProps = {
   onRequestBack?: (*) => void,
   direction?: "right" | "left",
   paper?: boolean,
+  title?: string,
 };
 
 export function SideDrawer({
@@ -124,6 +125,7 @@ export function SideDrawer({
   onRequestClose,
   onRequestBack,
   direction = "right",
+  title,
   ...props
 }: DrawerProps) {
   const [isMounted, setMounted] = useState(false);
@@ -211,15 +213,23 @@ export function SideDrawer({
                 p="24px"
                 style={{ zIndex: 200 }}
               >
-                {onRequestBack && (
+                {onRequestBack ? (
                   <Button onClick={onRequestBack} className="sidedrawer-close">
                     <IconAngleLeft size={12} />
                     <Text ff="Inter|Medium" fontSize={4} color="palette.text.shade40">
                       <Trans i18nKey="common.back" />
                     </Text>
                   </Button>
+                ) : (
+                  <Box />
                 )}
-                <Box flex="1" />
+
+                {title && (
+                  <Text ff="Inter|SemiBold" fontWeight="600" fontSize="18px">
+                    {title}
+                  </Text>
+                )}
+
                 {onRequestClose && (
                   <TouchButton onClick={onRequestClose} className="sidedrawer-close">
                     <IconCross size={16} />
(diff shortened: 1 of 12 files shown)


juan-cortes:
Why does the backdrop not cover the sidebar? that looks like a glitch to me,
Could you check if it's the case for other side drawers too @achabrolin-ledger ?

achabrolin-ledger:
> Why does the backdrop not cover the sidebar? that looks like a glitch to me,
> Could you check if it's the case for other side drawers too @achabrolin-ledger ?

Indeed, this behaviour does not happen on other drawers.
This is because I displayed (rendered) this info drawer component inside of the `WebPlatformPlayer/TopBar` component (`src/renderer/components/WebPlatformPlayer/TopBar.js`).

It looks like I should render it inside `src/renderer/components/TopBar/index.js`, alongside the other drawers, and trigger it's `open` state by dispatching a redux action.

I'll take a look at it and update it accordingly

juan-cortes:
@achabrolin-ledger mind requesting a review from @LFBarreto ,I'm off til monday

klambert-ledger (approved):
LGTM 👍

achabrolin-ledger:
> @achabrolin-ledger mind requesting a review from @LFBarreto ,I'm off til monday

Sure will do, sorry about that

src/renderer/components/SideDrawer.js
```diff
@@ -211,15 +213,23 @@ export function SideDrawer({
...
-                {onRequestBack && (
+                {onRequestBack ? (
                   <Button onClick={onRequestBack} className="sidedrawer-close">
                     <IconAngleLeft size={12} />
                     <Text ff="Inter|Medium" fontSize={4} color="palette.text.shade40">
                       <Trans i18nKey="common.back" />
                     </Text>
                   </Button>
+                ) : (
+                  <Box />
```
LFBarreto, line 224: do you need an empty box here?

src/renderer/components/WebPlatformPlayer/LiveAppInformationDrawer.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { closePlatformAppInfo } from "~/renderer/actions/UI";
+import { platformAppInfoStateSelector } from "~/renderer/reducers/UI";
+
+import Text from "../Text";
+import ExternalLink from "../ExternalLink/index";
+import AppDetails from "../Platform/AppDetails";
+
+const Divider = styled(Box)`
+  width: 420px;
```
LFBarreto, line 20: avoid using hard values like these prefer either % or defined theme values.
test this with different app sizes to make sure it works :+1:

src/renderer/components/WebPlatformPlayer/LiveAppInformationDrawer.js
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <SideDrawer
+      title={t(`platform.app.informations.title`)}
+      isOpen={isOpen}
+      onRequestClose={() => {
+        dispatch(closePlatformAppInfo());
+      }}
+      direction="left"
+    >
+      <Box pt="60px" height="100%" px="40px">
```
LFBarreto, line 44: try using the theme values for pt and px 
something like `pt={7} px={6}`
https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/styles/theme.js#L6

src/renderer/components/WebPlatformPlayer/TopBar.js
```diff
@@ -123,9 +128,14 @@ const WebPlatformTopBar = ({ manifest, onReload, onHelp, onClose }: Props) => {
         </ItemContent>
       </ItemContainer>
       <RightContainer>
-        {/* <ItemContainer isInteractive onClick={onHelp}>
+        <ItemContainer
+          isInteractive
+          onClick={() => {
```
LFBarreto, line 133: try using as much as possible `useCallback` when you can for inline callbacks to minimize reassigning functions on each render

LFBarreto (changes requested):
some minor changes needed :+1: 
good work otherwise

src/renderer/components/SideDrawer.js
achabrolin-ledger: Indeed, It's used in the same way as [this former Box](https://github.com/LedgerHQ/ledger-live-desktop/pull/3978/files#diff-4174bb1ccf94b6f908bbf947982f1b401f1f904e6e8401acb57accc155d4b663L222) `<Box flex="1" />` to maintain components position in this "header". 

But now, since we have 3 different components in this header (which can all be independently enable or not), we need to ensure we always have a component on each side of the header to maintain components position.
For example, we always want the cross to be on the right, even if we don't have back button or title (or only one of the two).

src/renderer/components/WebPlatformPlayer/LiveAppInformationDrawer.js
achabrolin-ledger: Thanks for the feedback 👍 
I still need to improve my `styled-components` / `styled-system` knowledge 💅

LFBarreto (approved):

valpinkman (approved):

PR #4018, opened 2021-07-15
LL-7199 Deliver the crypto_org UI components

crypto-facs:
### Type

Add Crypto Org to supported currencies.

- Right now, only tesnet indexer is available

NOTE: In the documentation some need for translations was mentioned. I am not sure of the require translations that are needed 

### Context

Crypto org wallet integration

Tested it my self checking balances, history transactions and generating new transfers. (Only basic functionality covered)


gre (commented):
code lgtm

gre (approved):

PR #4039, opened 2021-07-21
LL-6334 - Change Buy flow to add account if user has not one

klambert-ledger:
In Buy flow, if there is no account from the selected Currency we should only show the add account button instead of displaying an empty list.

### Type

UX Polish

### Context

[LL-6334](https://ledgerhq.atlassian.net/browse/LL-6334)

Buy flow with Coinlify
Buy flow with Wyre DApp


src/renderer/screens/exchange/Buy/SelectAccountAndCurrency.js
```diff
@@ -96,6 +96,65 @@ const SelectAccountAndCurrency = ({ selectAccount, defaultCurrency, defaultAccou
...
+          </FormContent>
+        </>
+      ) : (
+        <FormContent>
+          <ConfirmButton primary onClick={openAddAccounts}>
+            {t("exchange.buy.addAccount")}
+          </ConfirmButton>
+        </FormContent>
+      );
+    }
```
achabrolin-ledger, line 146: As a rule of thumb and in order to improve readability, it's a good practice to favor early returns

```suggestion
    if (!currency) {
      return;
    }

    if (!availableAccounts.length) {
      return (
        <FormContent>
          <ConfirmButton primary onClick={openAddAccounts}>
            {t("exchange.buy.addAccount")}
          </ConfirmButton>
        </FormContent>
      );
    }

    return (
      <>
        <FormContent>
          <AccountSelectorLabel>
            <span>{t("exchange.buy.selectAccount")}</span>
            <FakeLink fontSize={3} ff="Inter|SemiBold" onClick={openAddAccounts}>
              <PlusIcon size={10} />
              <Text style={{ marginLeft: 4 }}>{t("exchange.buy.addAccount")}</Text>
            </FakeLink>
          </AccountSelectorLabel>
          <SelectAccount
            accounts={availableAccounts}
            value={{ account, subAccount }}
            onChange={setAccount}
          />
        </FormContent>
        <FormContent>
          <ConfirmButton
            primary
            onClick={() => {
              if (!account) {
                return;
              }

              track("Buy Crypto Continue Button", {
                currencyName: getAccountCurrency(account).name,
                isEmpty: isAccountEmpty(account),
              });

              subAccount ? selectAccount(subAccount, account) : selectAccount(account);
            }}
            disabled={!account}
          >
            {t("exchange.buy.continue")}
          </ConfirmButton>
        </FormContent>
      </>
    );
```

src/renderer/screens/exchange/Buy/SelectAccountAndCurrency.js
```diff
@@ -96,6 +96,65 @@ const SelectAccountAndCurrency = ({ selectAccount, defaultCurrency, defaultAccou
     dispatch(openModal("MODAL_ADD_ACCOUNTS", { currency }));
   }, [dispatch, currency]);
 
+  const addOrSelectAccount = useMemo(() => {
```
achabrolin-ledger, line 99: Is there a specific reason for a `useMemo`? It seams like this is always re-rendered anyway considering the dependencies of this memoized component.

Alternatively, it could make sense to create a new `AddOrSelectAccount` or even 2 separated `AddAccount` and `SelectAccount` components to further separate concerns, leverage component composition and reduce components re render

achabrolin-ledger (changes requested):

src/renderer/screens/exchange/Buy/SelectAccountAndCurrency.js
klambert-ledger: 👍

src/renderer/screens/exchange/Buy/SelectAccountAndCurrency.js
klambert-ledger: Nope you're right, it's a reminiscence of a feature I tried to add but that didn't work. Removed. 👍 

Keeping it in 1 variable was on purpose, mainly to avoid creating another variable that would be named "selectAccountSomething" that would be conflicting with already declared selectors with similar names, making it way more confusing.

(Concerning the separate concerns, as discussed in my other PR, there will be no useless derender/rerender because the React renderer will diff everything)

achabrolin-ledger (approved):
Looks good to me 👌

stual-ledger:
Screenshot 2021-08-09 at 15 18 40
Screenshot 2021-08-09 at 15 16 37
Hey there :) unfortunately the fix doesn't seem to work, see screenshot attached. A quick `git log` show last commit `04f1a722c46f2fb111ccada610048053f89a769c` when I tested it.

stual-ledger (dismissed):
see comment above :)

klambert-ledger:
> see comment above :)

You're right, we had 2 times the same flow at 2 differents places in the code... 
It should be fixed now, sorry for this miss ! 👍

achabrolin-ledger (changes requested):
These 2 components are (almost) identical, why not refacto?

klambert-ledger:
> These 2 components are (almost) identical, why not refacto?

I agree, but I might not have the time to correctly refactoring this as of right now and test it properly to assure that there is no regression because of it. Maybe we should ask for this refacto in another ticket ?

achabrolin-ledger (dismissed):

ychen-ledger (changes requested):
hey @klambert-ledger , please find below, 

- The UI design is a bit different compare to expected behaviour: 
- The Select account window is not center align 
Expected behaviour: 
image-2021-07-08-15-21-47-868
Actual Result: 
Screenshot 2021-08-19 at 16 59 02

achabrolin-ledger:
> hey @klambert-ledger , please find below,
> 
> * The UI design is a bit different compare to expected behaviour:
> * The Select account window with Wyre Dapp is not center align

Hello @ychen-ledger, reply to your feedbacks in https://ledgerhq.atlassian.net/browse/LL-6334?focusedCommentId=125525

_**tldr**_;
everything is fine, we should be able to move on with this PR as is

ychen-ledger (approved):
Validate with this PR only for functionality but not UI part

achabrolin-ledger (approved):

valpinkman (approved):

PR #4043, opened 2021-07-22
live-common 20.8.0

gre:
activate swap wyre.
will regen snapshots.


gre:
@juan-cortes please try to provide a fix for this flaky test in another pr

PR #4050, opened 2021-07-27
LL-6512 (Manager): add fallback for manager app icons

LFBarreto:
(Manager): add fallback for manager app icons
when network is down or manager app icon can't load, the app will now show a placeholder icon instead
### Type

UI Polish
### Context

[LL-6512]

A bit hard to test unless you can block request from our cdn that provides the icons

[LL-6512]: https://ledgerhq.atlassian.net/browse/LL-6512


achabrolin-ledger:
You should be able to test this with apps called `Liquid Hless` and `Liquid Test Hless` (I receive 403 error from the CDN for these apps icons)

src/renderer/screens/manager/AppsList/AppIcon.js
```diff
@@ -26,13 +27,38 @@ const Container = styled.div`
...
+
 type Props = {
   app: App,
   theme: any,
 };
 
 function AppIcon({ app, theme }: Props) {
   const { currencyId, icon } = app;
+  const [loading, setLoading] = useState(true);
+  const onLoad = useCallback(() => setLoading(false), []);
```
achabrolin-ledger, line 61: What's the reason for using `useCallback` in this case?

achabrolin-ledger (changes requested):

src/renderer/screens/manager/AppsList/AppIcon.js
LFBarreto: i kept it to not re-render the function in case of props changes

valpinkman (approved):

PR #4058, opened 2021-07-29
[LL-6363] Fix typo in uncountable "information" wording

achabrolin-ledger:
_Information_ is uncountable in english, i.e can't be plural

### Type

Bug Fix

### Context

cf. [this comment](https://ledgerhq.atlassian.net/browse/LL-6363?focusedCommentId=120716) on Jira

cf. [Cambridge dictionary](https://dictionary.cambridge.org/grammar/british-grammar/information)

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -19,7 +19,7 @@
     },
     "app": {
       "informations": {
-        "title": "Informations",
+        "title": "Information",
         "website": "Website"
       }
     },


static/i18n/en/app.json
```diff
@@ -19,7 +19,7 @@
     },
     "app": {
       "informations": {
```
achabrolin-ledger, line 21: The key should be renamed as well, is it possible or would it wreak havoc with Crowdin translating platform?

ggilchrist-ledger: Good point. Ideally it would be consistent but probably not worth spending hours on if it does break things. Will you get errors before trying to run the app or would they only be caught by using the app?

achabrolin-ledger: If the key is updated alongside all of the places it is used in the app (today only one place) everything will work fine. But since these i18n files are generated and managed by a third party service (Crowdin), I don't know how these kind of key changes are managed (if they are)

ggilchrist-ledger (approved):

ggilchrist-ledger:
Looks good. I can see there are some test failures but a quick look at other PRs of LLD shows these are flaky tests. I'll raise this in slack to see what happens with these.

I have approved but I think my account isn't a high enough level to be merged @gre

PR #4060, opened 2021-07-29
Migrate to chartjs v3

juan-cortes:
Major upgrade on the charting library used in the app. Claims to bring performance improvements and allows us to stay up to date. The changes in our codebase are minor and if you look at the changelog they claim large performance improvements that can only benefit us in the long run.

https://www.chartjs.org/docs/latest/getting-started/v3-migration.html

image

### Type

Maintenance

### Context

n/a

Ideally, test the performance compared to the current version.


gre:
i'm including this in https://github.com/LedgerHQ/ledger-live-desktop/pull/4072

juan-cortes:
It's included in https://github.com/LedgerHQ/ledger-live-desktop/pull/4072

gre:
it is no longer included. it crashes at runtime. we'll have to check later.

gre:
@juan-cortes can you create a JIRA task and ping me on it? thanks.

PR #4063, opened 2021-07-30
HACKATHON - feat(transaction): display ens name

jgiamporcaro-ledger:


ririen:
@jgiamporcaro-ledger  I wonder about the title - what hackathon you are talking about?

gre:
it's an internal hack day event we do R&D in Ledger Live team. there is no ETA for this work but this is exploration we want to dig and consider for the future

gre:
This PR is a record of the work done in the internal hackday. we can close it for now as it's not mergeable and scoped yet internally, but the work is still on git :)

PR #4067, opened 2021-08-03
Swap drop the landing screen for spectron

juan-cortes:
With the merger of the swap Wyre, we broke (even more) the swap Spectron test. I'm not going to uncomment the commented tests because we are moving away from this flow completely and it should be part of that pr.

### Type

QA


juan-cortes:
This will not work 🤔

gre (changes requested):
still not passing?

juan-cortes:
Nope, doesn't seem to want to pass

juan-cortes:
Spectron, We haven't touched ETH at all, why are you doing this to me

juan-cortes:
Closes https://github.com/LedgerHQ/ledger-live-desktop/pull/4066 since it includes the https://github.com/LedgerHQ/ledger-live-desktop/pull/4067/commits/db34128823d127990cced5e5a714324ba3ba982f commit from @LFBarreto

gre (approved):

Issue #4016, opened 2021-07-14
Ubuntu 18.04 LTS - version 'GLIBCXX_3.4.26' not found


cryptonik:
#### Ledger Live Version and Operating System
- tested on Ledger Live 2.30.0 (ledger-live-desktop-2.30.0-linux-x86_64.AppImage)
- Platform and version: Ubuntu 18.04 LTS

#### Expected behavior
- Successful application startup

#### Actual behavior
- Application failed on startup with the error:
```
Internal process error (1)
Error
    at createCustomErrorClass (file:///tmp/.mount_ledgerqrA8Hp/resources/app.asar/.webpack/renderer.bundle.js:11608:19)
    at deserializeError (file:///tmp/.mount_ledgerqrA8Hp/resources/app.asar/.webpack/renderer.bundle.js:11636:35)
    at EventEmitter.handleCommandEvent (file:///tmp/.mount_ledgerqrA8Hp/resources/app.asar/.webpack/renderer.bundle.js:461212:61)
    at EventEmitter.emit (events.js:315:20)
    at Object.onMessage (electron/js2c/renderer_init.js:91:925)
```

#### Steps to reproduce the behavior
- Open `ledger-live-desktop-2.30.0-linux-x86_64.AppImage` on Ubuntu 18.04 LTS (or similar)
- Error found in logs:
```
/tmp/.mount_ledgerovAV0H/resources/app.asar/node_modules/bindings/bindings.js:121
throw e;
^
    Error: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /tmp/.mount_ledgerovAV0H/resources/app.asar.unpacked/node_modules/@ledgerhq/ledger-core/build/Release/ledger-core-node.node)
at process.func [as dlopen] (electron/js2c/asar_bundle.js:5:1812)
at Object.Module._extensions..node (internal/modules/cjs/loader.js:1203:18)
at Object.func [as .node] (electron/js2c/asar_bundle.js:5:2039)
at Module.load (internal/modules/cjs/loader.js:992:32)
at Module._load (internal/modules/cjs/loader.js:885:14)
at Function.f._load (electron/js2c/asar_bundle.js:5:12738)
at Module.require (internal/modules/cjs/loader.js:1032:19)
at require (internal/modules/cjs/helpers.js:72:18)
[...]

PR #4073, opened 2021-08-05
new libcore fix to fix linux crash

hzheng-ledger:
Fix linux crash:
https://github.com/LedgerHQ/ledger-live-desktop/issues/4016

For the libcore part, 
when we build the nodejs binding for libcore, I use the 
our own docker image ghcr.io/ledgerhq/lib-ledger-core-build-env/lib-ledger-core-build-env:latest
instead of the ubuntu16.04 github action image because our docker image has a lower stable GLIBCXX version. So that we don't have the GLIBCXX version problem now.

More detail can be found here for the nodejs binding build:  
https://github.com/LedgerHQ/lib-ledger-core-node-bindings/compare/v6.12.3_fix#diff-4e8bf7a54f3e0839dd1725cf4053a605ac30839b8c5b3202c17872960596fd60

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@ledgerhq/hw-transport": "6.3.0",
     "@ledgerhq/hw-transport-http": "6.3.0",
     "@ledgerhq/hw-transport-node-hid-singleton": "6.3.0",
-    "@ledgerhq/ledger-core": "6.12.3-beta1",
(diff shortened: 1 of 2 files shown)


gre:
we have internal confirmation that this fixes the problem. i'll do a merge to include it in release 2.32.2 and we'll do more assessment on it.

PR #4100, opened 2021-08-16
Ll 6641 Divider component

juan-cortes:
It's just a line with some margin.

### Type

Rebranding - UI

### Context

https://ledgerhq.atlassian.net/browse/LL-6641

### DIFF
--- a/ui-lib/src/components/Divider/Divider.stories.tsx
+++ b/ui-lib/src/components/Divider/Divider.stories.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+import Divider from "./index";
+import Text from "@ui/components/Text";
+
+export default {
+  title: "Divider",
+  component: Divider,
+  argTypes: {
+    type: {
+      options: ["light", "dark"],
+      control: {
+        type: "select",
+      },
+    },
+    margin: {
+      type: "number",
+    },
+  },
+};
+
+const Template = args => (
+  <div>
+    <Text>
+      {
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer ante mauris, tempor faucibus urna quis, gravida vulputate nisl. Nam nec commodo dui. Quisque semper neque non risus elementum maximus. Nam molestie porttitor nibh, quis tempus quam maximus quis. Phasellus ultrices, leo nec pulvinar volutpat, ipsum purus lobortis metus, eu pharetra eros libero sed dui. Pellentesque in ligula lorem. In id vestibulum ligula. Aliquam quam tellus, ornare non molestie non, ornare vitae ex. Aliquam vel urna nec leo mattis iaculis. Maecenas velit metus, bibendum eu tincidunt et, rhoncus ac elit. Cras tristique quam vel mauris scelerisque aliquet. Donec dapibus lectus at nisi blandit, nec tristique ligula porttitor. Aliquam nec iaculis ex."
+      }
+    </Text>
+    <Divider {...args} />
+    <Text>
+      {
+        "Cras maximus orci eget sem elementum accumsan. Donec at mollis odio, a varius metus. Donec dapibus id elit vitae dignissim. Suspendisse consequat ex ut scelerisque imperdiet. Donec interdum lorem risus. Praesent consectetur lectus vel libero volutpat, id dictum sapien vulputate. Donec vulputate eros at nulla aliquet rutrum sit amet lacinia tortor. Ut pretium, purus eget sollicitudin finibus, eros dolor commodo libero, ut pellentesque mauris odio vitae sem. Nam ornare arcu sit amet posuere vestibulum. Donec eget felis sed erat placerat suscipit."
+      }
+    </Text>
+  </div>
+);
+
+export const Default = Template.bind({});
--- a/ui-lib/src/components/Divider/Divider.stories.tsx
+++ b/ui-lib/src/components/Divider/Divider.stories.tsx
@@ -3,7 +3,7 @@ import Divider from "./index";
 import Text from "@ui/components/Text";
 
 export default {
-  title: "Divider",
+  title: "Interface/Divider",
   component: Divider,
   argTypes: {
     type: {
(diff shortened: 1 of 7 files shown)


valpinkman (dismissed):

ui-lib/src/styles/palettes/dusk.json
```diff
@@ -38,7 +38,8 @@
       "contrast": "hsla(0, 0%, 0%, 1)"
     },
     "grey": {
-      "border": "hsla(0, 1%, 23%, 1)"
+      "border": " hsla(0, 1%, 23%, 1)",
+      "borderDark": "hsla(0, 0%, 90%, 1)"
```
jgiamporcaro-ledger, line 42: Is the Dusk theme not outdated? I thought I heard that

ui-lib/src/styles/palettes/light.json
```diff
@@ -38,7 +38,8 @@
       "contrast": "hsla(0, 0%, 100%, 1)"
     },
     "grey": {
-      "border": "hsla(0, 0%, 90%, 1)"
+      "border": "hsla(0, 0%, 90%, 1)",
+      "borderDark": " hsla(0, 1%, 23%, 1)"
```
jgiamporcaro-ledger, line 42:
```suggestion
```

I think that we must only have one border key with different values depending on the theme.

_light.json_
```json
"grey": {
      "border": "hsla(0, 0%, 90%, 1)",
}
```

_dark.json_
```json
"grey": {
      "border": "hsla(0, 1%, 23%, 1)"
}
```

Thanks to that styles will be reactive.

ui-lib/src/components/Divider/index.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from "styled-components";
+
+type DividerTypes = "light" | "dark";
+
+interface Props {
+  margin?: number;
+  type?: DividerTypes; // Nb default to light grey
```
jgiamporcaro-ledger, line 7: Type shouldn't be a prop here, we want to design components in a way that the style is reactive.
In this implementation, we'll have to listen to the theme switch at the top of this component and return new props value depending on the user choice. 

By using a single key `theme.colors.palette.v2.grey.border` to reference two different values from the light and dark theme file, style is reactive by default.

jgiamporcaro-ledger (changes requested):

ui-lib/src/styles/palettes/dusk.json
juan-cortes: I have no idea how to maintain the palettes, if you are more up to date with what @LFBarreto had in mind, let me know

ui-lib/src/components/Divider/index.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from "styled-components";
+
+type DividerTypes = "light" | "dark";
+
+interface Props {
+  margin?: number;
```
jgiamporcaro-ledger, line 6: *I would like to add one more thing, I know that was the scope of the ticket, that's why I separate this comment from the PR review.*

---

For me, components must not use margin propriety, this is not their role. We should imagine components like a capsule and only define **internal style**.
That isn't the role of atoms, molecules or organisms to have an impact on the layout of a template or the pages. 

image


By setting the margin here, we go beyond the design definition because designers don't define external spaces on Figma. In fact, they will use grid/flexbox systems to apply some spacing rules, depending on the context (e.g define a larger spacing when the Divider is between two sections than between two paragraphs) and that's exactly what we will. When we'll start to integrate templates and pages we'll use Flexbox and Grid components to define layouts and use gaps to separate components in a harmony way. By allowing the possibility to set margin here, we mix atoms/molecules roles with layout/pages roles, the components receive two specific responsibilities which is one too many.

I am deeply convinced that is not a good way to go and I think that we should adopt the atomic [design methodology created by Brad Frost](https://bradfrost.com/blog/post/atomic-web-design/). (the book is a must-read btw). I will be happy to discuss with the team about that.

ui-lib/src/styles/palettes/light.json
juan-cortes: I read the figma has two different colors _per theme_, you are reading it as one per theme, which makes more sense.
I'll sync with design on this

ui-lib/src/components/Divider/Divider.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import Divider from "./index";
+import Text from "@ui/components/Text";
+
+export default {
+  title: "Divider",
```
jgiamporcaro-ledger, line 6:
```suggestion
  title: "Interface/Divider",
```

To respect Figma structure as asked by Gregor.

_(we should find a new category name tbh)_

ui-lib/src/components/Divider/index.tsx
juan-cortes: Once again, type makes sense if there are two types of divider per theme, if it's one per theme then indeed this is not needed

jgiamporcaro-ledger (changes requested):

ui-lib/src/styles/palettes/dusk.json
jgiamporcaro-ledger: Added note from my side to ask Luiz about that

ui-lib/src/styles/palettes/light.json
juan-cortes: Confirmed by design, there's one design per theme, changed it to match

ui-lib/src/components/Divider/index.tsx
juan-cortes: I trust you on this, up to you to convince the rest of the team though

ui-lib/src/components/Divider/index.tsx
juan-cortes: Confirmed by design, you were correct.

jgiamporcaro-ledger (approved):

valpinkman (approved):

PR #4103, opened 2021-08-17
LL 6640 - Rebrand Progress Bar

jelbaz-ledger:
Implement a fluid progress bar (aka in Figma: `SidePanel ProgressBar`).

See [here](https://github.com/LedgerHQ/ledger-live-desktop/pull/4103#issuecomment-902714682) for the latest update.

~https://user-images.githubusercontent.com/86958797/129705881-1da4525b-4378-4001-950b-0edffe574531.mp4~

### Type

Feature

### Context

[LL-6640]

[Updated Figma version](https://www.figma.com/file/WXCEmRMwW47ELrFonzPTGg/Library---LLD?version-id=1067897426&node-id=114%3A241)
[Original Figma version](https://www.figma.com/file/WXCEmRMwW47ELrFonzPTGg/Library---LLD?version-id=1066968409&node-id=114%3A241) 

[LL-6640]: https://ledgerhq.atlassian.net/browse/LL-6640


ui-lib/src/components/ProgressBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * The next step state, or undefined if the current step is the last one.
+   */
+  nextState?: StepState;
+};
+
+export const Item = {
+  Container: styled.div.attrs(_ => ({
+    mx: 2,
+  }))<ColorProps & BorderProps & SpaceProps>`
```
jgiamporcaro-ledger, line 44:
```suggestion
  Container: styled.div.attrs({ mx: 2 })<ColorProps & BorderProps & SpaceProps>`
```

If I remember correctly, the .attrs method accepts an object.

ui-lib/src/components/ProgressBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    display: flex;
+    align-self: stretch;
+    ${space}
+  `,
+  Current: styled.div.attrs(_ => ({
+    backgroundColor: "palette.v2.primary.dark",
+  }))<ColorProps>`
+    width: 6px;
+    height: 6px;
+    border-radius: 9999px;
```
jgiamporcaro-ledger, line 64: 🤔 ?

ui-lib/src/components/ProgressBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+};
+
+export const Item = {
+  Container: styled.div.attrs(_ => ({
+    mx: 2,
+  }))<ColorProps & BorderProps & SpaceProps>`
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 24px;
```
jgiamporcaro-ledger, line 48: -- **out of scope** --

We should talk together about the relative/absolute units and write some rules between us. For the moment, nothing is harmonised.

ui-lib/src/components/ProgressBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+          </Item.Container>
+        ) : state === "current" ? (
+          <Item.Container backgroundColor="palette.v2.primary.backgroundLight" borderRadius="8px">
+            <Item.Current />
+          </Item.Container>
+        ) : (
+          <Item.Container color="palette.v2.text.default">
+            <Item.Completed />
+          </Item.Container>
+        )}
```
jgiamporcaro-ledger, line 126: Maybe it's only me but I'm not a big fan of using nested ternary from the JSX tbh.
I generally prefer using a "switch" component that returns different JSX based on the state value.

Just my two cents, we don't have rules about that so it's not a change request.

jgiamporcaro-ledger (approved):

ui-lib/src/components/ProgressBar/index.tsx
jelbaz-ledger: Haha yeah it might look weird, it's a habit I took when I want to round something safely [after being bitten](https://github.com/tailwindlabs/tailwindcss/issues/1949) by `border-radius: 50%`. In this case it's overkill, I'll change to something more sensible 👍.

ui-lib/src/components/ProgressBar/index.tsx
jelbaz-ledger: > We should talk together about the relative/absolute units and write some rules between us

Absolutely!! 👍

ui-lib/src/components/ProgressBar/index.tsx
jelbaz-ledger: Good catch 🎣

ui-lib/src/components/ProgressBar/index.tsx
jelbaz-ledger: > I generally prefer using a "switch" component that returns different JSX based on the state value.

No strong opinion on this, I use both depending on the context. In this case I find it readable enough but I'll happily follow the guidelines if we define some project-wide rules.

juan-cortes:
I haven't yet looked at the code @jelbaz-ledger but I'm 100% sure @LFBarreto wants these to animate the fill of the bar when we transition from one step to another. A simple fill effect should suffice, can you take it?

jelbaz-ledger:
> I haven't yet looked at the code @jelbaz-ledger but I'm 100% sure @LFBarreto wants these to animate the fill of the bar when we transition from one step to another. A simple fill effect should suffice, can you take it?

Sure, I'm on it 👍

jelbaz-ledger:
@juan-cortes Allright so I added a transition, unfortunately not 100% perfect since there are 3 bar segments to animate in sequence. It still look good when the steps labels are not huge.

Capture d’écran 2021-08-19 à 14 25 06

ui-lib/src/components/ProgressBar/index.tsx
juan-cortes: We use ternary galore throughout all the projects, I think this is a decision that should be made by live-core and not us, so feel free to raise the point to the team and in the meantime, do whatever makes you happy

juan-cortes:
> @juan-cortes Allright so I added a transition, unfortunately not 100% perfect since there are 3 bar segments to animate in sequence. It still looks good when the steps labels are not unreasonably huge.
> 
>  Kapture.2021-08-19.at.17.12.59.mp4

Looks good to me, does it work with an arbitrary number of steps? Like... 5 steps
And if so, can I move from step 1 to step 4 with animation or would that break it all?

jelbaz-ledger:
> Looks good to me, does it work with an arbitrary number of steps? Like... 5 steps
And if so, can I move from step 1 to step 4 with animation or would that break it all?

Well the number of steps should not matter, but moving from more than 1 step makes all the separator animations trigger at the same time which looks a bit weird:


https://user-images.githubusercontent.com/86958797/130201582-ed92fb09-118c-4411-a0f5-b6ea83b73128.mp4

juan-cortes:
> > Looks good to me, does it work with an arbitrary number of steps? Like... 5 steps
> > And if so, can I move from step 1 to step 4 with animation or would that break it all?
> 
> Well the number of steps should not matter, but moving from more than 1 step makes all the separator animations trigger at the same time which looks a bit weird:
> 
>  Kapture.2021-08-20.at.10.05.38.mp4

I was just curious, it's fine for me, 90% of the time they will be sequential.
Does it handle error steps and disabled steps? If not, please ping design on it because on the current design we do have error indicators at the progress step level and I'm sure they will eventually want this too. Something like replacing the number with an X and showing the bar and bullet in red or whatever.

jelbaz-ledger:
> Does it handle error steps and disabled steps? If not, please ping design on it because on the current design we do have error indicators at the progress step level and I'm sure they will eventually want this too.

Pinged design on Figma about this, but got no reply yet 📞.

jelbaz-ledger:
Ok so here are the changes made to comply with the new design in Figma:

- No more animation or separator color changes
- Extra separators on the side of the 
- Error state added
- Color changes
- Renamed the component to ProgressSteps (+ figma category to Stepper)


https://user-images.githubusercontent.com/86958797/130236945-f41a1cc4-c348-4948-8244-21b197eedec8.mp4

jelbaz-ledger:
Updated again to comply with design changes, restored the separator transition on step change and removed extra separators on the side.


https://user-images.githubusercontent.com/86958797/130245150-d01acc41-de9d-42c5-be9d-6850d981bf75.mp4

ui-lib/src/components/ProgressSteps/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const inactive = state === "pending";
+  const nextInactive = nextState === "pending";
+  const errored = state === "errored";
+  return (
+    <Flex flexDirection="column" alignItems="center">
+      <Item.Spacer mb="12px">
+        {(!hideLeftSeparator && <Separator.Item inactive={inactive} position="left" />) || (
+          <Flex flex="1" />
+        )}
+        {state === "pending" ? (
```
LFBarreto, line 145: prefer the use of a switch or better use something like a registry of step components by state like 
```javascript
const stepComponentsByState = {
 pending: <Item.Container>
            <Item.Pending />
          </Item.Container>,
  current: <Item.Container backgroundColor="palette.v2.primary.backgroundLight" borderRadius="8px">
            <Item.Current />
          </Item.Container>
}


...
return <>
{ stepComponentsByState[state] }
</>;
```

LFBarreto (changes requested):

achabrolin-ledger (changes requested):

ui-lib/src/components/ProgressSteps/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+function getState(activeIndex: number, index: number, errored?: boolean) {
+  return activeIndex < index
+    ? "pending"
+    : activeIndex === index
+    ? errored
+      ? "errored"
+      : "current"
+    : "completed";
+}
```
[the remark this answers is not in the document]
jelbaz-ledger: IMO this is purely subjective. For instance, the rule you linked is not included in the eslint `recommended` preset.

jelbaz-ledger:
@LFBarreto @jgiamporcaro-ledger @achabrolin-ledger 

Guys, I understand that many of you dislike ternaries. I know there is an eslint rule (which we don't seem to enforce, and which is not included in the recommended eslint preset). 

I can change the code in this PR to remove ternaries, but as some of you noticed ternaries are used all over the place already. 

**So my point is that maybe we should enforce a project-wide rule that we all follow by asking all the team members about this?**

*Personally, I have no problem following whatever is decided. But my two cents is that this is purely related to code style and 100% subjective.*

jelbaz-ledger:
@LFBarreto @achabrolin-ledger @jgiamporcaro-ledger 

Hmm seems like I did not check the right eslint config file actually, `ui-lib` has the `"no-nested-ternary": 0,` rule, so my bad! 🤭

LFBarreto (approved):

valpinkman (approved):

PR #4115, opened 2021-08-18
LL-6642 Rebrand - Carousel component

juan-cortes:
Screenshot 2021-08-19 at 12 35 35

There are a few shortcomings in the current PR, in particular, we are missing translatable strings, which I'm not sure whether it should be on the `ui-lib` side or somewhere else 🤔 and also the color rework that will need to be addressed on a separate PR at a later date.

### Type

Rebrand - Component Carousel

### Context

https://ledgerhq.atlassian.net/browse/LL-6642


juan-cortes:
Don't review yet, it's a draft, pending work from design

juan-cortes:
Merged develop in since I don't think the CI failure is related to anything I did (considering it's scoped to the ui-lib folder)

valpinkman (approved):

PR #4117, opened 2021-08-19
skip choose currency step if needed

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LL-5711

### DIFF
--- a/src/renderer/screens/exchange/Sell/SelectAccountAndCurrency.js
+++ b/src/renderer/screens/exchange/Sell/SelectAccountAndCurrency.js
@@ -108,7 +108,7 @@ const SelectAccountAndCurrency = ({
   const dispatch = useDispatch();
 
   const openAddAccounts = useCallback(() => {
-    dispatch(openModal("MODAL_ADD_ACCOUNTS", { currency, skipChooseCurrencyStep: true }));
+    dispatch(openModal("MODAL_ADD_ACCOUNTS", { currency }));
   }, [dispatch, currency]);
 
   const renderOptionOverride = ({ data: currency }: Option) => {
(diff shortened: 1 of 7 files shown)


src/renderer/screens/exchange/Sell/SelectAccountAndCurrency.js
```diff
@@ -108,7 +108,7 @@ const SelectAccountAndCurrency = ({
   const dispatch = useDispatch();
 
   const openAddAccounts = useCallback(() => {
-    dispatch(openModal("MODAL_ADD_ACCOUNTS", { currency }));
+    dispatch(openModal("MODAL_ADD_ACCOUNTS", { currency, skipChooseCurrencyStep: true }));
```
gre, line 111: correct me if i'm wrong but to me we don't need to introduce a new flag, the simple fact there is a currency in prop is enough to know we should "skip"

gre (changes requested):

src/renderer/screens/exchange/Sell/SelectAccountAndCurrency.js
juan-cortes: @gre As far as I can see ( I checked the code and tested it) there is no provision for skipping a step in the current code, what we do have is a means to provide the currency to prefill, as we do on the manager, it merely opens the add account modal with the currency pre-filled. What am I missing 🤓 ?


https://user-images.githubusercontent.com/4631227/130210300-6a3e3726-e708-43bb-89d5-37942e5c6fd2.mov

hzheng-ledger: @gre @juan-cortes In terms of the code. Thanks for your feedback. I take into account all of them and they should be all fixed now. You can review again.
@juan-cortes For the feature, what I understand is that we provide the currency to prefill in the "choosecurrency" step if we know the currency, but to simplify the flow to add an account, we skip this "choosecurrency" step if we already know the currency. this can save one click in terms of the user experience.
Correct me if I make a mistake. Thanks

gre (approved):

Arnaud97234 (approved):
QA OK, After installing an app on the device then the add account banner is displayed in the Manager. The add account flow start at step 2, the currency is selected automatically.
✅

PR #4122, opened 2021-08-20
Add AccountSubHeader per family section

gre:
This allows a coin integration family to add a section between the header and the graph in the Account screen.

For instance, by adding this file

**src/renderer/families/ethereum/AccountSubHeader.js**

```
function AccountSubHeader() {
  return "IT WORKS.";
}
export default AccountSubHeader;

```

it will renders this:

Screenshot 2021-08-20 at 09 40 00

### Type

feature

### Context

coin integration
https://ledgerhq.atlassian.net/browse/LL-6944

**if CI pass, it can gets in, there are no impact on existing app as no currencies uses this yet**

### DIFF
--- a/scripts/sync-families-dispatch.sh
+++ b/scripts/sync-families-dispatch.sh
@@ -9,6 +9,7 @@ accountActions.js \
 TransactionConfirmFields.js \
 AccountHeaderActions.js \
 AccountBodyHeader.js \
+AccountSubHeader.js \
 SendAmountFields.js \
 SendRecipientFields.js \
 SendWarning.js \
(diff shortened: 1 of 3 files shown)


juan-cortes (dismissed):
lgtm

achabrolin-ledger (approved):
Looks good to me but CI fails, it looks like spectron snapshots should be updated (maybe because UI changed 🤔 )

haammar-ledger (dismissed):

gre:
i think it's an actual issue. i changed the padding.

gre:
i just don't get why it's changing. i moved numbers (of margin). and their total didn't change.

PR #4156, opened 2021-09-01
bump live-common

valpinkman:


juan-cortes (dismissed):

achabrolin-ledger (commented):
I couldn't find the 21.3.1 [release on LLC repo](https://github.com/LedgerHQ/ledger-live-common/releases) to see the changes.

But other that that looks good to me I guess.

For info, what is the `autofocus` update you snuck into this PR as well? It seems related to test because it relates to spectron from what I see, but don't know what it's related to 🤷

valpinkman:
@achabrolin-ledger so i don't think we push release in github for LLC, we need to review that with @gre 
The autofocus is indeed linked to spectron tests, it was the one causing the wallet connect modal to sometimes be focused.

I think I need to update my screenshots as well, as the changes seem legit

achabrolin-ledger:
> @achabrolin-ledger so i don't think we push release in github for LLC, we need to review that with @gre
> The autofocus is indeed linked to spectron tests, it was the one causing the wallet connect modal to sometimes be focused.
> 
> I think I need to update my screenshots as well, as the changes seem legit

Ok thanks, then looks good to me 👌

achabrolin-ledger (dismissed):

PR #4158, opened 2021-09-01
LL-6359 - SWAP - Improve the from selector

jgiamporcaro-ledger:
- Flag as disable unswappable accounts
- Render the check mark icon close to the current value
- Render a tooltip when an account is disable
- Remove the search bar

https://user-images.githubusercontent.com/33158502/131709027-22d16864-eada-46c5-af34-f007fbb075cf.mov

## 🖤  Expectations to reach

### DIFF
--- a/src/renderer/components/Select/index.js
+++ b/src/renderer/components/Select/index.js
@@ -18,7 +18,7 @@ export type Option = {
   data: any,
 };
 
-type Props = {
+export type Props = {
   // required
   value: ?Option,
   options: Option[],
@@ -46,7 +46,7 @@ type Props = {
   rowHeight: number,
   error: ?Error, // NB at least a different rendering for now
   stylesMap: CreateStylesReturnType => CreateStylesReturnType,
-  extraRenderers?: { [string]: (props: *) => React$ElementType }, // Allows overriding react-select components. See: https://react-select.com/components
+  extraRenderers?: { disableTooltipText?: string, [string]: (props: *) => React$ElementType }, // Allows overriding react-select components. See: https://react-select.com/components
 };
 
 const Row = styled.div`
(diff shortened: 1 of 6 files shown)


src/renderer/components/Select/index.js
```diff
@@ -46,7 +46,7 @@ type Props = {
   rowHeight: number,
   error: ?Error, // NB at least a different rendering for now
   stylesMap: CreateStylesReturnType => CreateStylesReturnType,
-  extraRenderers?: { [string]: (props: *) => React$ElementType }, // Allows overriding react-select components. See: https://react-select.com/components
+  extraRenderers?: { disableTooltipText?: string, [string]: (props: *) => React$ElementType }, // Allows overriding react-select components. See: https://react-select.com/components
```
jelbaz-ledger, line 49: IMHO `disableTooltipText`:
- should not be a property of `extraRenderers`, because `extraRenderers` has been made specifically to have keys corresponding to [react-select component names](https://react-select.com/components) - plus the object is spread and passed to react-select through the `components` prop and the additional keys will leak.
- have a different name since my understanding of `disableTooltipText` would be "disables the tooltip text" but I think that it actually "shows the tooltip text when the item is disabled"

So maybe we could pass it as a regular prop?

jelbaz-ledger: @jgiamporcaro-ledger If you're ok with the change I can make a follow-up PR since this one has been merged already

jgiamporcaro-ledger: Yeah, I have no problem with the change.
In fact, the disableTooltipText is the string to pass to the tooltip component.

jelbaz-ledger: > In fact, the disableTooltipText is the string to pass to the tooltip component.

Woops my bad then, name is fine!

PR #4163, opened 2021-09-02
LL-6359 - SWAP - Multiline selectors

jelbaz-ledger:
## 🦒 Context (issues, jira)

To suit the design, changes the account/currency selectors values to display 2 lines of text for the selected item.
Also fixes (kinda) the overlap issue with the BTC derivation mode tag.

[LL-6359]
[LL-7010]

Capture d’écran 2021-09-02 à 17 29 22
Capture d’écran 2021-09-02 à 17 29 29
Capture d’écran 2021-09-02 à 17 29 33

## 🖤  Expectations to reach

### DIFF
--- a/src/renderer/components/AccountTagDerivationMode.js
+++ b/src/renderer/components/AccountTagDerivationMode.js
@@ -23,15 +23,16 @@ const CurrencyLabel: ThemedComponent<*> = styled(Text).attrs(() => ({
   flex: 0 0 auto !important;
   box-sizing: content-box;
   text-transform: uppercase;
-  margin: 0 8px;
+  margin: ${p => p.margin || "0 8px"};
   flex: unset;
 `;
 
 type Props = {
   account: AccountLike,
+  margin?: string,
 };
 
-export default function AccountTagDerivationMode({ account }: Props) {
+export default function AccountTagDerivationMode({ account, margin }: Props) {
   if (account.type !== "Account") return null;
 
   const tag =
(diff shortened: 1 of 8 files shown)


src/renderer/components/AccountTagDerivationMode.js
```diff
@@ -23,21 +23,22 @@ const CurrencyLabel: ThemedComponent<*> = styled(Text).attrs(() => ({
...
   box-sizing: content-box;
   text-transform: uppercase;
-  margin: 0 8px;
+  margin: ${p => p.margin || "0 8px"};
   flex: unset;
 `;
 
 type Props = {
   account: AccountLike,
+  margin?: string,
```
juan-cortes, line 32: Text already inherits the box attrs that allow you to specify the margin with `mt mb m mx and my` so you probably don't need this. You could simply have it as the snippet below, and override any value in the usage itself `<MyText mx={123} />` or whatever.
```
styled(Text).attrs(p => ({
  mx: p.mx || 2,
  my: p.my || 0,
}))`
  // Other styles
`;
```
Which allows you to use theme sizes instead of pixels too

juan-cortes (approved):

src/renderer/components/AccountTagDerivationMode.js
jelbaz-ledger: > You could simply have it as the snippet below, and override any value in the usage itself

Hmm I'm not sure about this 🤔

`margin` is a prop of `AccountTagDerivationMode`, which gets passed down to `CurrencyLabel`. Without it, I think that there is no way of controlling the `CurrencyLabel` margin from above. 

Then, `CurrencyLabel` is a `styled(Text)` with an hardcoded css `margin: 0 8px;` which I guess will always override the styled-system margin prop from `Text` since it extends it. Also This code is legacy, so I tried not to touch it.

What I could do is rewrite it to remove the hardcoded margin and use a margin prop instead which would be better for the reasons you highlighted above.

PR #4167, opened 2021-09-06
electron-docker-webdriver container fix

nabil-brn:
Lastest image doesn't work because x11vnc is missing a library.

## 🦒 Context (issues, jira)
Quick fix for QA tools

## 🖤  Expectations to reach
Docker container builds should not fail with latest docker image.


valpinkman (approved):

juan-cortes:
Thank you @nabil-brn you magnificent specimen

PR #4200, opened 2021-09-13
LL-6773 - SWAP - Rename "swap" to "exchange"

qd-qd:
## 🦒 Context (issues, jira)

[LL-6773]

## 🖤  Expectations to reach

[LL-6773]: https://ledgerhq.atlassian.net/browse/LL-6773

### DIFF
--- a/static/i18n/en/app.json
+++ b/static/i18n/en/app.json
@@ -193,7 +193,7 @@
     "accounts": "Accounts",
     "manager": "Manager",
     "exchange": "Buy / Sell",
-    "swap": "Swap",
+    "swap": "Exchange",
     "lend": "Lend",
     "catalog": "Discover"
   },
@@ -213,7 +213,7 @@
     "disconnectButton": "Disconnect"
   },
   "swap2": {
-    "title": "Swap",
+    "title": "Exchange",
     "tabs": {
       "exchange": "Exchange",
       "history": "History"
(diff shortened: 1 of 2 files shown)


jgiamporcaro-ledger:
Ready to see the e2e tests explode 🔥

static/i18n/en/app.json
```diff
@@ -252,19 +252,19 @@
       "notAvailable": "Service temporarily unavailable, or not available in your country"
     },
     "history": {
-      "disclaimer": "Your swap desktop transactions are not synchronized with the Ledger Live mobile application",
+      "disclaimer": "Your exchange desktop transactions are not synchronized with the Ledger Live mobile application",
```
juan-cortes, line 255: This doesn't seem right @cthepot-ledger can we get a wording check?

static/i18n/en/app.json
```diff
@@ -438,25 +438,25 @@
         },
         "finished": {
           "title": "Pending",
-          "subtitle": "Swap broadcast successfully ",
-          "swap": "Your Swap ID:",
+          "subtitle": "Exchange broadcast successfully ",
```
juan-cortes, line 441: Exchange _broadcasted_ ?

juan-cortes (changes requested):

static/i18n/en/app.json
jgiamporcaro-ledger: I agree with you, this doesn't seem right at all. I asked the product team to double-check that all sentences that include the "swap" word are still correct after the "exchange" renaming because of that. The ticket only ask to find and replace "swap" with "exchange" but I think this is not enough, we should take time to double-check all sentences updated and decide if they need to be rewritten or not.

jgiamporcaro-ledger:
The ticket linked to this PR has been moved back to the spec column 🤷  
Transforming this PR as a draft one.

cthepot-ledger:
closing as swap is primary word

PR #4218, opened 2021-09-15
LL-6359 - SWAP - Displays error when the amount is lower than the fees

jgiamporcaro-ledger:
That covers the case when the amount received from the backend is lower
than the fees itself. In that case, we're returning an dedicated error
and we filter the provider.

## 🦒 Context (issues, jira)

[LL-6359]

## 🖤  Expectations to reach

[LL-6359]: https://ledgerhq.atlassian.net/browse/LL-6359

### DIFF
--- a/src/renderer/screens/exchange/Swap2/utils/shared/useSwapTransaction.js
+++ b/src/renderer/screens/exchange/Swap2/utils/shared/useSwapTransaction.js
@@ -17,8 +17,8 @@ import type {
 } from "@ledgerhq/live-common/lib/types";
 import { AmountRequired } from "@ledgerhq/errors";
 import { flattenAccounts } from "@ledgerhq/live-common/lib/account/helpers";
-
 import { shallowAccountsSelector } from "~/renderer/reducers/accounts";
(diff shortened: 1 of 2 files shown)


juan-cortes (commented):
It makes no sense for us to handle this UI side to me, we should sync on this topic

LFBarreto (changes requested):
this will need to be handled by the bridge on live common side :grin: 
error handling is done there we can indeed explain to you a bit more in detail the bridge / transaction error flow :+1:

juan-cortes:
> this will need to be handled by the bridge on live common side 😁
> error handling is done there we can indeed explain to you a bit more in detail the bridge / transaction error flow 👍

It's not amount at the transaction level but at the rate level I think, so it's not at the bridge at all (correct me if I'm wrong @jgiamporcaro-ledger) it's either at the getRates if we handle it ourselves or by backend if they do it themselves. 🤔

jgiamporcaro-ledger:
I'm closing this PR. After talking with Juan/Luiz/Julien, we decided that the logic to handle this error should be located to `live-common`.

PR #4240, opened 2021-09-20
added stake button to eth account; fixed design bug on dropdown item …

IAmMorrow:
…width

## 🦒 Context (issues, jira)

[LL-7343]

image

## 🖤  Expectations to reach

Added a "Stake" button to eth accounts. Clicking on it takes you to the Lido Platform app page on the correct account.

[LL-7343]: https://ledgerhq.atlassian.net/browse/LL-7343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/renderer/components/Button.js
```diff
@@ -347,6 +349,8 @@ class Button extends PureComponent<
     const { isFocused } = this.state;
     const { disabled } = this.props;
     const { onClick, children, isLoading, event, eventProperties, ...rest } = this.props;
+
+    console.log(this.props);
```
achabrolin-ledger, line 353:
```suggestion
```

src/renderer/screens/account/AccountHeaderActions.js
```diff
@@ -234,6 +241,16 @@ const AccountHeaderActions = ({ account, parentAccount, openModal, t }: Props) =
...
+      {currency.id === "ethereum" ? (
+        <Button small primary onClick={onPlatformStake} event={"Eth Stake Account Button"} icon>
+          <Box horizontal flow={1} alignItems="center">
+            <IconCoins size={14} />
+            <Box>
+              <Trans i18nKey="account.stake" values={{ currency: currency.name }} />
+            </Box>
+          </Box>
+        </Button>
+      ) : null}
```
achabrolin-ledger, line 253: From what I've seen it look likes this whole logic could be implemented in an Ethereum specific `AccountHeaderActions` file and leverage code generation, as done today with algorand, cosmos, polkadot and tron (cf. [here](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/scripts/sync-families-dispatch.sh#L10) and [here](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/families/tron/AccountHeaderActions.js) for example)

ghost (changes requested):
Changes could leverage byFamily code generation abstraction.
It looks like snapshot tests should be updated as well

src/renderer/screens/account/AccountHeaderActions.js
IAmMorrow: Yes, and no.

this logic doesn't apply to the whole ethereum family but to the ethereum currency only. So it will always need a "if" somewhere.

Never the less, this piece of code will change with the rebranding, and has to be refactored anyway.

IAmMorrow: Also this file already has some "if currency === ethereum" logic, so it's not really making it any worse

IAmMorrow: Your point remain valid, and in the futur we'll have to work on a system where the whole account UI logic is generic but overloadable both by family or by currency.

ghost (dismissed):

LFBarreto (dismissed):

ghost (dismissed):

ghost (approved):

PR #4247, opened 2021-09-22
LL-7223 add watermark to nightly version

valpinkman:
## 🦒 Context (issues, jira)

[LL-7223]

CleanShot 2021-09-22 at 14 54 52@2x
watermark on the top of the sidebar. 
yes it here. 
i swear. 
look closely. 

## 🖤  Expectations to reach

[LL-7223]: https://ledgerhq.atlassian.net/browse/LL-7223

### DIFF
--- a/src/renderer/Default.js
+++ b/src/renderer/Default.js
@@ -58,6 +58,50 @@ export const TopBannerContainer: ThemedComponent<{}> = styled.div`
   }
 `;
 
+const NightlyLayerR = () => {
+  const children = [];
+  const w = 200;
+  const h = 100;
+  for (let y = 0.5; y < 20; y++) {
+    for (let x = 0.5; x < 20; x++) {
+      children.push(
+        <div
+          style={{
+            position: "absolute",
+            textAlign: "center",
+            top: y * h,
+            left: x * w,
+            transform: "rotate(-45deg)",
+          }}
+        >
+          NIGHTLY
+          <br />
+          {__APP_VERSION__}
+        </div>,
+      );
+    }
+  }
+  return (
+    <div
+      style={{
+        position: "fixed",
+        pointerEvents: "none",
+        opacity: 0.1,
+        color: "#777",
+        width: "100%",
+        height: "100%",
+        top: 0,
+        right: 0,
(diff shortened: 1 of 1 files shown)


achabrolin-ledger (commented):
I don't know what's the purpose of this watermark, but if the goal is to make it obvious to a user of such build that he is using a nightly (thus unstable) version of the app, I feel like as of today the watermark is too subtle (not visible / obvious enough).

In this cas I would rather suggest a sort of red banner (with the version inside why not) in the top left corner of the app in order to make it more obvious

juan-cortes (approved):

achabrolin-ledger:
> I don't know what's the purpose of this watermark, but if the goal is to make it obvious to a user of such build that he is using a nightly (thus unstable) version of the app, I feel like as of today the watermark is too subtle (not visible / obvious enough).
> 
> In this cas I would rather suggest a sort of red banner (with the version inside why not) in the top left corner of the app in order to make it more obvious

We could think of something like that (ugly but visible). We could polish the CSS if need be 🤷‍♂️ 
<details>
 <summary> Example 📷  </summary>


example

</details>

juan-cortes:
@achabrolin-ledger it's also about customer support identifying a user issue when they provide a screenshot we can know if it's from a nightly or not. if we only mark the top left, the user can take a screenshot of the impacted part of the screen and we wont know if it's a nightly or not. I agree maybe it's too subtle but again, we want the users to be able to use the app without hating it, they/we are helping us find bugs and debug

achabrolin-ledger:
> @achabrolin-ledger it's also about customer support identifying a user issue when they provide a screenshot we can know if it's from a nightly or not. if we only mark the top left, the user can take a screenshot of the impacted part of the screen and we wont know if it's a nightly or not. I agree maybe it's too subtle but again, we want the users to be able to use the app without hating it, they/we are helping us find bugs and debug

Ok thanks for the context, I thought it was for internal use only but if it's also used by external end users, makes sense to have it less obvious and displayed on the whole screen 👌

achabrolin-ledger (approved):

PR #4285, opened 2021-10-01
Fix cold launch deep link for bridge

juan-cortes:
## 🦒 Context (issues, jira)

{ MISSING  JIRA TASK }
https://ledger.slack.com/archives/C01REQQBJS3/p1632429876223500

image

In some cases, the loaded settings from disk override the dispatched `deepLinkurl` from launch. This is a proposal for a fix but we could probably find a cleaner one. In fact I guess we can remove the `.once` call in favor of just this variable approach but what if the order of calls is flipped? Better to be safe than sorry.

Here's a totally unnecessary excalidraw sketch of what is happening, in case my super verbose description on slack isn't clear enough, long story short we rely on a setting from the store to handle the deep linking urls and, although we detect the url correctly, we handle the dispatch before we load the initial settings from file, meaning this url is immediately overridden by null and hence doesn't trigger the necessary action.

### This should fix all deeplinking actions from cold launch, not just the bridge.

image

## 🖤  Expectations to reach

### How to test
- Assign this app to open the `ledgerlive://` urls, you might need to uninstall the production build for this.
- With a closed app trigger the bridge from meta mask.
- It should open the LLD app with the bridge modal open like above.

### DIFF
--- a/src/renderer/init.js
+++ b/src/renderer/init.js
@@ -88,13 +88,17 @@ async function init() {
 
   sentry(() => sentryLogsSelector(store.getState()));
 
+  let deepLinkUrl; // Nb In some cases `fetchSettings` runs after this, voiding the deep link.
   ipcRenderer.once("deep-linking", (event, url) => {
     store.dispatch(setDeepLinkUrl(url));
+    deepLinkUrl = url;
   });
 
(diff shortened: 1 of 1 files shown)


MortalKastor (approved):
🚀

PR #4294, opened 2021-10-05
Playwright

nabil-brn:
## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/B2CQA-30

https://user-images.githubusercontent.com/16896559/136051960-ce60f19f-06aa-4360-ad8f-d65532605421.mov

## 🖤  Expectations to reach
- [x] Setup & Conf
- [x] Fixtures (userdata)
- [x] Visual comparison
- [x] Mocks (accounts & device)
- [x] Recorder
- [x] Reporter
- [x] Code coverage
- [ ] CI & artifacts management
- [ ] CodeceptJS POC
- [ ] Migration
- [ ] Documentation


package.json
```diff
@@ -39,7 +39,12 @@
     "spectron": "jest \"tests/specs/.*\\.spec\\.js\" --maxWorkers=4",
     "spectron-inband": "jest \"tests/specs/.*\\.spec\\.js\" --runInBand",
     "spectron-coverage": "nyc report --reporter=text-lcov --temp-dir=tests/coverage | codecov --pipe --disable=gcov",
-    "generate-screenshots": "PWD=$(pwd); cd docker; docker run -v \"$PWD:/github/workspace\" -w \"/github/workspace\" --rm -it $(docker build -q .)"
+    "generate-screenshots": "PWD=$(pwd); cd docker; docker run -v \"$PWD:/github/workspace\" -w \"/github/workspace\" --rm -it $(docker build -q .)",
+    "pretest": "tsc --incremental -p playwright/tsconfig.json",
+    "playwright": "cross-env ALLURE_RESULTS_DIR=playwright/artifacts/allure-results npx playwright test --config=playwright/playwright.config.ts",
+    "playwright-ci": "nyc yarn playwright",
```
ggilchrist-ledger, line 45: Wonder if it might be better to rename this `playwright-code-coverage` or something to make it clearer what it's doing. Unless it's also doing something specific to the CI as well?

playwright/fixtures/common.ts
```diff
@@ -0,0 +1,100 @@
...
+export function generateUUID(): string {
+  return crypto.randomBytes(16).toString('hex');
+}
+
+type TestFixtures = {
+  lang: String;
+  theme: String;
+  userdata: String;
+  env: Object;
+  page: any;
```
ggilchrist-ledger, line 16: Can this be changed to `Page`?

playwright/fixtures/common.ts
```diff
@@ -0,0 +1,100 @@
...
+      fs.copyFileSync(userDataFile, `${userDataPath}/app.json`);
+    }
+
+    // default environment variables
+    env = Object.assign(
+      {
+        MOCK: true,
+        HIDE_DEBUG_MOCK: true,
+        CI: process.env.CI || undefined,
+        SPECTRON_RUN: true,
```
ggilchrist-ledger, line 41: What's `SPECTRON_RUN` doing?

playwright/fixtures/common.ts
```diff
@@ -0,0 +1,100 @@
...
+    const istanbulCLIOutput = path.join('playwright/artifacts/.nyc_output');
+
+    await page.addInitScript(() =>
+    window.addEventListener('beforeunload', () =>
+      (window as any).collectIstanbulCoverage(JSON.stringify((window as any).__coverage__))
+      ),
+    );
+    await fs.promises.mkdir(istanbulCLIOutput, { recursive: true });
+    await page.exposeFunction('collectIstanbulCoverage', (coverageJSON: string) => {
+      if (coverageJSON)
```
ggilchrist-ledger, line 79: Could we not move this to the start of the code coverage section so that we don't run it if it's not needed? We seem to be doing extra things that may not be needed

playwright/models/DeviceAction.ts
```diff
@@ -0,0 +1,61 @@
...
+  async openApp() {
+    await this.page.evaluate(() => {
+      window.mock.events.mockDeviceEvent({ type: "opened" });
+    });
+
+    await this.page.waitForSelector("#deviceAction-loading", { state: "visible" });
+  }
+
+  async genuineCheck() {
+    const result = mockListAppsResult("Bitcoin", "Bitcoin", deviceInfo);
```
ggilchrist-ledger, line 23: Why are we hard coding Bitcoin?

playwright/models/OnboardingPage.ts
```diff
@@ -0,0 +1,40 @@
...
+    ]);
+  }
+
+  async selectDevice(device) {
+    await this.page.click(`button:has-text("${device}")`)
+  }
+
+  async connectDevice() {
+    await this.page.click(
+      'button:has-text("Connect deviceConnect your Nano XIs your device already set up? Connect it to th")',
```
ggilchrist-ledger, line 28: Seems like a bit of a flaky string. Probably ok for now but worth review what elements need better `id`s or `name`s in future

playwright/fixtures/common.ts
valpinkman: it's a env we use inside the app to change the behavior of certain components (like for example, turning off the animation when running the UI tests)

playwright/fixtures/common.ts
```diff
@@ -0,0 +1,100 @@
...
+import * as fs from "fs";
+import * as path from "path";
+import * as crypto from 'crypto';
+
+export function generateUUID(): string {
+  return crypto.randomBytes(16).toString('hex');
+}
+
+type TestFixtures = {
+  lang: String;
```
valpinkman, line 12: all `String` should be `string`

playwright/fixtures/common.ts
valpinkman: we could use `Page` from `@playwright/test`

playwright/models/DeviceAction.ts
valpinkman: agreed, pass an arg to `genuineCheck(coin1, coin2)` so you can pass it to `mockListAppsResult(coin1, coin2, deviceInfo)` or something similar

playwright/models/OnboardingPage.ts
valpinkman: ^ 100% agreed

playwright/specs/account.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+test.describe.parallel("Accounts", () => {
+  for (const currency of currencies) {
+    test(`[${currency}] Add account`, async ({ page }) => {
+      const accountModal = new AccountModal(page);
+      const deviceAction = new DeviceAction(page);
+
+      await test.step(`[${currency}] Open modal`, async () => {
+        await accountModal.open();
+        expect(await page.screenshot()).toMatchSnapshot(`open-modal.png`);
```
ggilchrist-ledger, line 19: We seem to be going in the direction of doing all our asserts as Matching Snapshots. I like this since it means we are checking for styling regressions at the same time as functional tests, but I wonder if test runs will be flakier and slower as a result of all the screenshots. Also do we need to worry about storing lots of image files or anything?

I'm ok with it as it is but just want to make sure we are happy with this approach going forward

playwright/specs/manager.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+test("Manager", async ({ page }) => {
+  const managerPage = new ManagerPage(page);
+  const deviceAction = new DeviceAction(page);
+
+  await test.step("can access manager", async () => {
+    await managerPage.navigate();
+    await deviceAction.manager();
+    expect(await page.screenshot()).toMatchSnapshot({
+      name: "manager-catalog.png",
```
ggilchrist-ledger, line 16: Out of interest how does it know what folder to find the pngs in?

ggilchrist-ledger:
Left some comments/questions @nabil-brn. Looks really good!

playwright/fixtures/common.ts
ggilchrist-ledger: Ah gotcha. I guess we should change the name once we've switched but good for now :)

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+      "target": "es6",
+      "strict": true,
+      "module": "commonjs",
+      "moduleResolution": "Node",
```
valpinkman, line 6:
```suggestion
      "moduleResolution": "node",
```

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+      "target": "es6",
```
valpinkman, line 3:
```suggestion
      "target": "esnext",
```

valpinkman (changes requested):
Last suggestions
- move `playwright/playwright.config.ts` to `./playwright.config.ts` and remove the option from the npm scripts
- rename and use where needed `tsconfig.json` to `tsconfig.playwright.json` or something similar (`tsconfig.test.json`) and use it in npm scripts

playwright/specs/manager.spec.ts
ggilchrist-ledger: Nabil explained it's basically a clever bit of magic from playwright

playwright/models/OnboardingPage.ts
nabil-brn: Yes! You're right, I planned to rework those tests during our migration but why not now 😄 .

To be clear, we should not rely on elements text, because it's flaky ofc and each time the wording changes we'll have false negative errors. It'll also block us from testing languages on Ledger Live.

Moreover, `:has-text` is a pseudo-class that is not a part of [CSS3](https://www.w3.org/TR/selectors-3/) or even [CSS4](https://www.w3.org/TR/selectors-4/) specifications, those are [custom classes](https://playwright.dev/docs/selectors#css-selector) (_Playwright adds custom pseudo-classes like :visible, :text and more._) added by Playwright itself, so we must avoid to use them for the moment.

playwright/models/DeviceAction.ts
nabil-brn: I agree, I just copy/pasted what we did on spectron 😄 .

playwright/fixtures/common.ts
nabil-brn: Thank you guys!

playwright/fixtures/common.ts
nabil-brn: Agreed, I'm waiting for *Plawright + CI* brainstorm to improve this, so when we develop tests locally, we won't do those extra things.

valpinkman:
@live-bot generate-screenshots

valpinkman:
@live-bot generate-screenshots

MortalKastor (dismissed):

valpinkman (dismissed):

valpinkman (approved):

LFBarreto (approved):
:robot:

ghost (approved):

PR #4296, opened 2021-10-06
LL-6449 NFT integration

juan-cortes:
> **NOTE** I'm setting it as not a draft in order to have a build, but this can't be merged until we point back to a non-github-commit version of live-common. It's here to allow @IAmMorrow to integrate with the remaining parts. Also note that the op details in this version are broken for nft operations, I'm looking into it now.

## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-6449

image
image
image
image

## 🖤  Expectations to reach


src/renderer/components/ContextMenu/NFTContextMenu.js
```diff
@@ -0,0 +1,54 @@
...
+export default function NFTContextMenu({ leftClick, children, contract, tokenId }: Props) {
+  const { t } = useTranslation();
+
+  const menuItems = [
+    {
+      key: "opensea",
+      label: t("NFT.viewer.actions.open", { viewer: "Opensea.io" }),
+      Icon: IconOpensea,
+      type: "external",
+      callback: () => openURL(`https://opensea.io/assets/${contract}/${tokenId}`),
```
LFBarreto, line 26: would be best to move these urls into live-common as env variables i think :thinking:

src/renderer/drawers/Drawer.js
```diff
@@ -60,31 +67,52 @@ export const Drawer = () => {
 
   const onRequestClose = useCallback(() => setDrawer(), [setDrawer]);
 
+  const sideDrawerRef = useRef();
+  const onResize = useCallback(e => {
+    // NB Passing this down to the Component as a prop allows us to have sticky elements
+    // by setting a fixed height instead of a percentage of the container, if you are reading
+    // this and have a better solution, treat yourself. -77 due to drawer wrapper padding.s
+    if (!sideDrawerRef.current) return;
+    setHeight(sideDrawerRef.current.clientHeight - 77);
```
LFBarreto, line 76: why not use css for this :thinking: are you loading an iframe inside the drawer or a simple `calc(100vh - 77)` can't work in this case ? :thinking:

src/renderer/drawers/NFTViewerDrawer/ExternalViewerButton.js
```diff
@@ -0,0 +1,120 @@
...
+
+export const ExternalViewerButton = ({ nft }: ExternalViewerButtonProps) => {
+  const { t } = useTranslation();
+
+  const items: DropDownItemType[] = [
+    {
+      key: "opensea",
+      label: t("NFT.viewer.actions.open", { viewer: "Opensea.io" }),
+      icon: <IconOpensea size={16} />,
+      onClick: () => openURL(`https://opensea.io/assets/${nft.collection.contract}/${nft.tokenId}`),
```
LFBarreto, line 58: same as before either factorise it or include this in live-common :+1:

src/renderer/drawers/NFTViewerDrawer/index.js
```diff
@@ -0,0 +1,236 @@
...
+              uppercase
+              pb={5}
+            >
+              {name}
+            </Text>
+          </StickyWrapper>
+          <Skeleton show={show} width={393} minHeight={393}>
+            <Image nft={metadata} full />
+          </Skeleton>
+          <NFTActions>
```
LFBarreto, line 172: remove this if unused

src/renderer/drawers/OperationDetails/index.js
```diff
@@ -124,6 +137,10 @@ const OperationD: React$ComponentType<Props> = (props: Props) => {
   const { extra, hash, date, senders, type, fee, recipients: _recipients } = operation;
   const recipients = _recipients.filter(Boolean);
   const { name } = mainAccount;
+  const operations = useMemo(() => [operation], [operation]);
+  const nfts = nftsFromOperations(operations);
+  const { status, metadata } = useNftMetadata(nfts[0]?.collection.contract, nfts[0]?.tokenId);
```
LFBarreto, line 142: same here seems counterintuitive to get only the first element of nftsFromOperations it should return that element directly if you're only using/expect one

src/renderer/drawers/NFTViewerDrawer/index.js
klambert-ledger: It's just temporary, we're just hiding the send buttons because LLC & LLM isn't ready yet and we want to release the display only. Will be removed for next release 👌

src/renderer/modals/Send/Body.js
```diff
@@ -193,6 +200,44 @@ const Body = ({
...
+      if (account !== nextAccount) {
+        setAccount(nextAccount);
+      }
+      const bridge = getAccountBridge(nextAccount);
+      const standard = nextNft.collection.standard.toLowerCase();
+
+      setTransaction(
+        bridge.updateTransaction(transaction, {
+          tokenIds: [nextNft.tokenId],
+          quantities: standard === "erc1155" ? [BigNumber(1)] : undefined,
```
LFBarreto, line 216: this logic should be handled in the transaction module for `erc1155.transfer` on live-common you'll need to duplicate this on mobile if you keep it here.

src/renderer/drawers/OperationDetails/index.js
klambert-ledger: That's some of the limitation we have because the backend lacks informations...
The general idea is that an NFT collection should have an image, but right now it's not the case. So we are using the first nft of the collection as the image of that collection (+ it contains also the name of the collection).

That's some architecture issue that we can't really resolve right now. 😞

src/renderer/screens/nft/Collections/Row.js
```diff
@@ -0,0 +1,69 @@
...
+          <Text ff="Inter|SemiBold" color="palette.text.shade100" fontSize={4}>
+            <Skeleton width={80} minHeight={24} barHeight={10} show={!tokenName}>
+              {tokenName}
+            </Skeleton>
+          </Text>
+        </Skeleton>
+      </Box>
+      <Skeleton width={42} minHeight={24} barHeight={10} show={show}>
+        <Text ff="Inter|SemiBold" color="palette.text.shade100" fontSize={4}>
+          {nfts?.length}
```
LFBarreto, line 62: you should escape this to a default value maybe

src/renderer/modals/Send/Body.js
gre: agreed. fundamentally i'm not even sure it's a good idea that we have the knowledge, on LLD/LLM side, of "erc1155", it should be abstracted out to "nft" because it's now it looks like in the model.
but too late to change this right now.

gre (approved):
we have stuff identified for improvement but no blocker to me.

src/renderer/components/ContextMenu/NFTContextMenu.js
juan-cortes: Not sure if we need it as an env or not, but it doesn't harm

src/renderer/drawers/Drawer.js
juan-cortes: couldn't make it work with the sticky elements, it was @IAmMorrow who took over afterwards but if you can make it work without this hack, feel free

src/renderer/drawers/OperationDetails/NFTOperationDetails.js
```diff
@@ -0,0 +1,89 @@
...
+import { centerEllipsis } from "~/renderer/styles/helpers";
+
+const NFTOperationDetails = ({ operation }: { operation: Operation }) => {
+  const { t } = useTranslation();
+  const operations = useMemo(() => [operation], [operation]);
+  const nfts = nftsFromOperations(operations);
+  const { status, metadata } = useNftMetadata(nfts[0]?.collection?.contract, nfts[0]?.tokenId);
+  const show = useMemo(() => status !== "loaded", [status]);
+
+  return !nfts[0] ? null : (
```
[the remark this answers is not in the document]
juan-cortes: Because it's a single nft but the hook i was provided returns "nfts plural for operations" hence the hacky memoized operation array and nfts array when I only need one.

src/renderer/screens/nft/Collections/Row.js
juan-cortes: If there are no nfts there would be no collection row, but since the type makes it optional I had to add the ?, if we are here, there are nfts.

klambert-ledger: I just added a 0 as fallback just in case but yeah, that's probably never going to be needed. 👍

src/renderer/modals/Send/Body.js
klambert-ledger: Yep I guess Juan is correct, we will always need this distinction in the LLD/LLM as ERC721 and ERC1155 have a different flow. We could probably skip the ternary though because the `erc721.transfer` module doesn't even use the quantities property anyway 🤷

klambert-ledger: I removed the ternary, I'll add it again if I notice a bug once we tackle the send feature tests.

src/renderer/drawers/Drawer.js
klambert-ledger: Tried to understand why we needed that, couldn't notice a difference with or without all of the additional stuff in the Drawer even when resizing. 
Removed it for now, I'll just cherry pick the commit and fix it properly with the `calc` if I can find what the problem was at some point 🤷‍♂️

LFBarreto (dismissed):
c'est super ça :1st_place_medal:

LFBarreto (approved):

klambert-ledger:
> c'est super ça 🥇

soupeur

PR #4300, opened 2021-10-07
LL-7489 Fix fee & amount values hidden when confirming a Tx with discreet ON

OlivierFreyssinet:
## 🦒 Context (issues, jira)

[LL-7489]

#### Step to reproduce :
* Put the discreet mode on
* Go on Discover
* Do a transaction
* At the moment of the transaction signature, i'm not able to compare values with my HW as they are hidden
Expected behavior : When the user has the discreet mode ON, he should be able to see the values in the "Device" screen to compare values (amount, fees) between the device and the computer.

### Solution:
Just using the `alwaysShowValue` prop that was defined in the `FormattedVal` component to override discreet mode.

I also added this prop to the `CounterValue` component as well to pass it down to the `FormattedVal` component it renders.

Then setting this prop `alwaysShowValue` to true for `CounterValue` and `FormattedVal` components rendered in `StepSummary` and `TransactionSummary` components.
**For the latter (`TransactionSummary`) it might affect other parts of the app where we actually don't want this behavior, I don't think so given what I understand regarding the role of this component but maybe reviewers with better knowledge of the codebase can double check this** 

Screenshot 2021-10-07 at 16 35 53

Screenshot 2021-10-07 at 16 36 27
## 🖤  Expectations to reach

[LL-7489]: https://ledgerhq.atlassian.net/browse/LL-7489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/renderer/components/TransactionConfirm/index.js
+++ b/src/renderer/components/TransactionConfirm/index.js
@@ -73,6 +73,7 @@ const FeesField = ({ account, parentAccount, status, field }: FieldComponentProp
         fontSize={3}
         inline
         showCode
+        alwaysShowValue
       />
     </TransactionConfirmField>
   );


ychen-ledger (commented):
@ofreyssinet-ledger ,  Please find step summary still have value hidden, please find the result as below(MacOs/LLD2.34.3/Paraswap/ETH-> USDT):

https://user-images.githubusercontent.com/85551245/140732256-96618b39-e683-499e-b5bf-fe48d94610c6.mov

ychen-ledger (commented):
Validated in new PR https://github.com/LedgerHQ/ledger-live-desktop/pull/4400#pullrequestreview-802660545

PR #4308, opened 2021-10-11
LL-6731 regions, dates, time and currencies

OlivierFreyssinet:
## 🦒 Context (issues, jira)

[LL-6731]

Added a new field `locale` in the settings that is set when the user changes the region. Its value will be a locale.
`localeSelector` has been modified to use this field while taking into account old existing settings and a new logic to fallback to the language setting.

The existing `region` field in the settings hasn't been removed as it's used for segment analytics we need to handle users' existing saved settings.

- Numbers formatting: Almost all numbers in the app are already formatted using localeSelector
- Currencies formatting: same
- Dates & time formatting: dates are formatted using moment.js so the logic to init/update moment's locale has been changed. (It was previously using the language setting)

- From a user's perspective, the region settings are now completely decoupled from the language settings so they can choose whatever region they want regardless of the language settings.
- All dates, numbers, times and currencies are now formatted using the locale defined by the region settings.
- The description of the region settings has been modified (using the text from the spec) to better explain what this setting does.

## 🖤  Expectations to reach

[LL-6731]: https://ledgerhq.atlassian.net/browse/LL-6731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


LFBarreto (dismissed):
:rocket:

LFBarreto (changes requested):

LFBarreto (dismissed):

LFBarreto (dismissed):

LFBarreto (dismissed):

ghost (dismissed):
Nice, this is long overdue 🚀

nabil-brn (dismissed):
✔️ QA-ed

nabil-brn (approved):

PR #4405, opened 2021-11-09
LL-8076 Update polkadot and live-common

gre:
https://github.com/LedgerHQ/ledger-live-common/releases/tag/v21.15.0


gre:
let's land what we have in the pipe of live-common.

PR #4415, opened 2021-11-12
LL-8112

LFBarreto:
- (Modals): not allow for more than one modal to show at same time
- fix translated "Switch to français" issue

## 🦒 Context (issues, jira)

[LL-8112]

https://user-images.githubusercontent.com/11752937/141437021-06f08d3b-a0c7-457a-8de4-44f2d016ae9e.mp4

## 🖤  Expectations to reach

[LL-8112]: https://ledgerhq.atlassian.net/browse/LL-8112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/renderer/ModalsLayer.js
+++ b/src/renderer/ModalsLayer.js
@@ -29,9 +29,9 @@ const BackDrop: ThemedComponent<{ state: string }> = styled.div.attrs(({ state }
 `;
 
 const ModalsLayer = ({ visibleModals }: *) => {
-  const filteredModals = visibleModals.filter(
-    ({ name, MODAL_SHOW_ONCE }) => !MODAL_SHOW_ONCE || !global.sessionStorage.getItem(name),
-  );
+  const filteredModals = visibleModals
+    .filter(({ name, MODAL_SHOW_ONCE }) => !MODAL_SHOW_ONCE || !global.sessionStorage.getItem(name))
+    .slice(0, 1);
   filteredModals.forEach(
     ({ name, MODAL_SHOW_ONCE }) =>
       MODAL_SHOW_ONCE && global.sessionStorage.setItem(name, Date.now()),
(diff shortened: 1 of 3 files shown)


tlaforge-ledger (approved):

src/renderer/ModalsLayer.js
```diff
@@ -29,9 +29,9 @@ const BackDrop: ThemedComponent<{ state: string }> = styled.div.attrs(({ state }
 `;
 
 const ModalsLayer = ({ visibleModals }: *) => {
-  const filteredModals = visibleModals.filter(
-    ({ name, MODAL_SHOW_ONCE }) => !MODAL_SHOW_ONCE || !global.sessionStorage.getItem(name),
-  );
+  const filteredModals = visibleModals
+    .filter(({ name, MODAL_SHOW_ONCE }) => !MODAL_SHOW_ONCE || !global.sessionStorage.getItem(name))
+    .slice(0, 1);
```
gre, line 34: beware this logic appears to prevent any "wrapped" modal to happen. Are we sure there are not a single case where this would be the case in Live?

valpinkman (approved):

src/renderer/ModalsLayer.js
juan-cortes: Looking for calls to `openModal(` inside the `/modals` folder would reveal nested modals handled by the modals layer. I went over all the matches (7) and the only case that seem to show a stacked modal is:
- Account migration, which we don't currently use https://github.com/LedgerHQ/ledger-live-desktop/blob/master/src/renderer/modals/MigrateAccounts/steps/StepOverview.js#L177 
All other usages are after a `closeModal` call, preventing the stack.

PR #4430, opened 2021-11-17
LL-8157 Exit manager if we enter app, avoid loop

juan-cortes:
## 🦒 Context (issues, jira)
When in the manager and not in a firmware update flow, consider a disconnect as the user interacting with the app some other way, and exit the manager.

## 🖤  Expectations to reach

### Test 1
- Access the manager with a device
- When in the manager, enter an app in the device
- Expected we redirect to the portfolio

### Test 2
- Do the same with a device that has a firmware update
- Disconnect the device while the firmware update modal is open.
- Expected we show the device action to reconnect the device instead of redirecting

-------


amougel:
Overridden by LL-7588

PR #4468, opened 2021-11-25
LL-8207 Improve app size granularity on the Manager

tlaforge-ledger:
## 🦒 Context (issues, jira)
[LL-8207]

1 - New device has more precise values data about storage (apps, firmware allocation, free space, etc...). We don't really need it on UI side for now. So we have to control the precision and the values to show.
2 - We wanted to better handle the rounding of data being less optimistic. When we are showing data about used space, we want to ceil the value and when it's about free space we will floor it to always unsure user will not be surprised thinking he could install something but when I do it, it fails cause missing few bytes on nano device.

before

after

To test it : 
1 - Use a nano SP and go to the manager
2 - Check it works with other devices

## 🖤  Expectations to reach

[LL-8207]: https://ledgerhq.atlassian.net/browse/LL-8207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/renderer/components/ByteSize.js
+++ b/src/renderer/components/ByteSize.js
@@ -6,17 +6,19 @@ import type { DeviceModel } from "@ledgerhq/devices";
 const k = 1024; // 1kb unit
 const sizes = ["bytes", "kbUnit", "mbUnit"];
 
-/** formats a byte value into its correct size in kb or mb unit takling in account the device block size */
+/** formats a byte value into its correct size in kb or mb unit taking in account the device block size */
 const ByteSize = ({
   value,
   deviceModel,
   decimals = 2,
   firmwareVersion,
+  formatFunction,
 }: {
   value: number,
   deviceModel: DeviceModel,
   decimals?: number,
   firmwareVersion: string,
+  formatFunction?: (val: number) => number,
 }) => {
   if (!value) return "–";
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/ByteSize.js
```diff
@@ -6,17 +6,19 @@ import type { DeviceModel } from "@ledgerhq/devices";
 const k = 1024; // 1kb unit
 const sizes = ["bytes", "kbUnit", "mbUnit"];
 
-/** formats a byte value into its correct size in kb or mb unit takling in account the device block size */
+/** formats a byte value into its correct size in kb or mb unit talking in account the device block size */
```
juan-cortes, line 9: Probably "taking" not "takling" nor "talking"

src/renderer/components/ByteSize.js
```diff
@@ -28,13 +30,14 @@ const ByteSize = ({
   const i = Math.floor(Math.log(bytes) / Math.log(k));
 
   const dm = Math.max(0, decimals);
+  const pureSize = parseFloat(bytes / Math.pow(k, i));
```
juan-cortes, line 33: pureSize seems confusing, maybe "displaySize" 🤔

juan-cortes (changes requested):
Please fix ci,  provide screenshots of the before/after, other than that I'm ok with the solution 👍🏼

juan-cortes:
Reset the branch and leave only your changes please 🙏🏼 @tlaforge-ledger

juan-cortes (approved):
lgtm, let's move this on 🚀

PR #4478, opened 2021-11-29
Migrate swap tests to playwright

ggilchrist-ledger:
## 🦒 Context (issues, jira)
Added swap spec to playwright tests

https://user-images.githubusercontent.com/87659842/143889642-7c7e55d6-0156-4191-b5f7-e6bfd3c81c4a.mov


ggilchrist-ledger:
The following files must be from when I merged from `develop` recently and `playwright` must be out of date

src/renderer/modals/Send/steps/StepAmount.js
static/i18n/en/app.json

PR #4484, opened 2021-12-01
[LL-7343] added stake button to eth account

ghost:
## 🦒 Context (issues, jira)
added stake button to eth account

https://ledgerhq.atlassian.net/browse/LL-7343

Based on https://github.com/LedgerHQ/ledger-live-desktop/pull/4240

~Blocked by https://github.com/LedgerHQ/ledger-live-desktop/pull/4496~

<details>
<summary> 🎥  </summary>

https://user-images.githubusercontent.com/87011321/147817580-0dc0936f-cd67-47ae-ab80-41fcae71ff34.mov

</details>

## 🖤  Expectations to reach

### DIFF
--- a/src/renderer/components/Button.js
+++ b/src/renderer/components/Button.js
@@ -349,8 +349,6 @@ class Button extends PureComponent<
     const { isFocused } = this.state;
     const { disabled } = this.props;
     const { onClick, children, isLoading, event, eventProperties, ...rest } = this.props;
-
-    console.log(this.props);
     const isClickDisabled = disabled || isLoading;
     const onClickHandler = e => {
       if (onClick) {
--- a/src/renderer/components/Button.js
+++ b/src/renderer/components/Button.js
@@ -280,6 +280,7 @@ export const Base: ThemedComponent<*> = styled.button.attrs(p => ({
   cursor: ${p => (p.disabled ? "not-allowed" : "pointer")};
   height: ${p => (p.small ? 34 : 40)}px;
   pointer-events: ${p => (p.disabled ? "none" : "")};
+  width: ${p => (p.fullWidth ? "100%" : "auto")};
   outline: none;
 
   ${p => getStyles(p, "default")};
@@ -305,6 +306,7 @@ export type Props = {
   lighterDanger?: boolean,
   disabled?: boolean,
   outline?: boolean,
+  fullWidth?: boolean,
   outlineGrey?: boolean,
   onClick?: Function,
   small?: boolean,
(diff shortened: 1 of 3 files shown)


LFBarreto (dismissed):

OlivierFreyssinet (dismissed):

ofreyssinet-ledger:
LGTM 👌
One thing that I'm curious about is whether there is a particular reason why we're doing Ethereum specific stuff in this generic `AccountHeaderAction` instead of having a family specific `ethereum/AccountHeaderManageActions` like for other coin families such as Algorand or Polkadot?

achabrolin-ledger:
> LGTM 👌 One thing that I'm curious about is whether there is a particular reason why we're doing Ethereum specific stuff in this generic `AccountHeaderAction` instead of having a family specific `ethereum/AccountHeaderManageActions` like for other coin families such as Algorand or Polkadot?

Good question indeed, the answer is that, as it is the case for WalletConnect, this feature is only available for the Ethereum blockchain and not other Ethereum family blockchains (like Binance Smart Chain for example)

ofreyssinet-ledger:
> > LGTM 👌 One thing that I'm curious about is whether there is a particular reason why we're doing Ethereum specific stuff in this generic `AccountHeaderAction` instead of having a family specific `ethereum/AccountHeaderManageActions` like for other coin families such as Algorand or Polkadot?
> 
> Good question indeed, the answer is that, as it is the case for WalletConnect, this feature is only available for the Ethereum blockchain and not other Ethereum family blockchains (like Binance Smart Chain for example)

Ah ok this makes sense, thanks for the explanation. So as I understand we currently have a logic for family specific features but not blockchain specific features?

src/renderer/components/Button.js
```diff
@@ -280,6 +280,7 @@ export const Base: ThemedComponent<*> = styled.button.attrs(p => ({
   cursor: ${p => (p.disabled ? "not-allowed" : "pointer")};
   height: ${p => (p.small ? 34 : 40)}px;
   pointer-events: ${p => (p.disabled ? "none" : "")};
+  width: ${p => (p.fullWidth ? "100%" : "auto")};
```
ggilchrist-ledger, line 283: Out of interest what is this doing? @achabrolin-ledger

achabrolin-ledger: Basically it allow (via the `fullWidth` prop) to specify if the element should take the full with of his parent element (`100%`) or not. In the later case, the width would be calculated automatically by the browser due to the `auto` value

cf. https://developer.mozilla.org/en-US/docs/Web/CSS/width#values

PR #4486, opened 2021-12-01
LL-7771 - v3 Onboarding Quiz

OlivierFreyssinet:
## 🦒 Context (issues, jira)

Run with env var `TEST_ONBOARDING_QUIZZ=1` to test it

[LL-7771]

TODO on UI repo:
- [ ] Fix Radio element buttons shrinking
- [ ] Fix Radio element text align vertical
- [ ] Fix Popin max width & height (too constraining)

https://user-images.githubusercontent.com/91890529/144280639-6b945691-0617-43e3-96d2-d33dd6e85424.mov

## 🖤  Expectations to reach

[LL-7771]: https://ledgerhq.atlassian.net/browse/LL-7771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/renderer/components/ModalQuizz/ModalQuizz.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const AsideRight = (
+    <Answer
+      Illustration={rightSideIllustration}
+      illustrationSize={userMadeAChoice ? 176 : 280}
+      title={rightSideTitle}
+      description={rightSideExplanation}
+    />
+  );
+
+  const [width, height] = [816, 486];
```
elbywan, line 185: I think that these dimensions are not really used by the popin component since there are fixed max width/height values in the styled components theme:

```
popin: {
        min: {
          height: 158,
          width: 462,
        },
        max: {
          height: 522,
          width: 622,
        },
      },
```

src/renderer/components/ModalStepper/CloseButton.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { Box, Button, Icons } from "@ledgerhq/react-ui";
+import styled from "styled-components";
+
+const CloseContainer = styled(Box).attrs(() => ({
+  top: 7,
+  right: 7,
+}))`
+  position: absolute;
```
elbywan, line 9: `position` seems also to be a prop of the `<Box />` component so it could be passed with `.attr` I think

src/renderer/components/ModalStepper/index.tsx
```diff
@@ -51,8 +52,17 @@ const ModalStepper = (props: Props) => {
     else setStepIndex(Math.max(0, stepIndex - 1));
   }, [stepIndex, onClose]);
 
+  const [width, height] = [816, 486];
```
elbywan, line 55: (also here for the popin height/width)

src/renderer/modals/OnboardingQuizz/OnboardingQuizzModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+import StartScreen from "./StartScreen";
+
+export default function QuizzModal() {
+  const [started, setStarted] = useState(false);
+  const { t } = useTranslation();
+  const dispatch = useDispatch();
+  const onClose = useCallback(() => {
+    dispatch(closeModal("MODAL_ONBOARDING_QUIZZ"));
+  }, [dispatch]);
+  // const onCloseLoop = useCallback(() => {
```
elbywan, line 16: There is some commented code block here 🧹

elbywan (commented):
Noice work!! 👍 
I just made a couple of comments about minor stuff.

Also I have some reservations about introducing the `TEST_ONBOARDING_QUIZZ` flag + code that leaks to V2 when it could be avoided.
My understanding is that we wanted to fully separate V3 & V2, @LFBarreto WDYT?

static/i18n/en/app.json
```diff
@@ -7159,7 +7159,7 @@
         "title": "Now game on!",
         "descr": "Answer 3 simple questions to avoid common misconceptions about your hardware wallet.",
         "buttons": {
-          "start": "Let’s take the quiz",
+          "start": "I'm ready",
```
elbywan, line 7162: ⚠️ Impacts V2

src/renderer/components/ModalQuizz/ModalQuizz.tsx
```diff
@@ -0,0 +1,221 @@
...
+  started,
+  StartScreen,
+  dismissable = true,
+}: Props) => {
+  const [stepIndex, setStepIndex] = useState(0);
+  const stepCount = steps.length;
+
+  const { t } = useTranslation();
+
+  const [userChoiceIndex, setUserChoiceIndex] = useState(undefined);
```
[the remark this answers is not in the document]
ofreyssinet-ledger: ah yes, thanks !

src/renderer/components/ModalQuizz/ModalQuizz.tsx
```diff
@@ -0,0 +1,221 @@
...
+      </Radio>
+    </Flex>
+  );
+
+  const rightSideIllustration =
+    (userMadeAChoice
+      ? isCorrectChoice
+        ? CorrectAnswerIllustration
+        : IncorrectAnswerIllustration
+      : Illustration) || Illustration;
```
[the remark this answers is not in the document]
ofreyssinet-ledger: Yes because `CorrectAnswerIllustration` and `IncorrectAnswerIllustration` are optional and if undefined it will fallback to `Illustration`, maybe this will be clearer:

```js
const rightSideIllustration = userMadeAChoice
  ? isCorrectChoice
    ? CorrectAnswerIllustration || Illustration
    : IncorrectAnswerIllustration || Illustration
  : Illustration;
```

src/renderer/components/ModalQuizz/ModalQuizz.tsx
ofreyssinet-ledger: Yes indeed but actually I wanted to change this as there are bigger `popin` components in the [design](https://www.figma.com/file/q1bu6mRiwzIhBEXhszfsts/LLD-%2F-Onboarding?node-id=201%3A1672), i don't think we should go for a static max size like this (in px), maybe make it a percentage ...

static/i18n/en/app.json
ofreyssinet-ledger: It's in the v3 part of the json and the text in the design is actually different from v2, should I just create a new entry for this ?

ofreyssinet-ledger:
Thanks for the review @elbywan ! Actually you made me notice that the scope of the task was a bit larger and it was not only implementation but integration as well (I didn't think the onboarding was ready to integrate the quiz yet).
I'm going to work on a proper integration now and the test flag won't be necessary anymore 👍

src/renderer/components/ModalQuizz/ModalQuizz.tsx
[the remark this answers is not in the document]
elbywan: Ah I see, thanks for explaining! 👍

static/i18n/en/app.json
elbywan: Oh my bad sorry I did not see that it was scoped inside the V3 part 🙇

PR #4506, opened 2021-12-06
Tests stabilization (accounts & settings)

nabil-brn:

- `accounts.spec.ts` is failing on macOS
- `settings.spec.ts` rarely fails but it does, it's due to mouse moves and the scrollbar is showing up sometimes

## 🖤  Expectations to reach
CI must be green ✔️


valpinkman (dismissed):
If this brings back macOS in CI

image

then fuck yeah

PR #4536, opened 2021-12-13
UI tests: Fix flakiness and added firmwareupdate spec

nabil-brn:
## 🦒 Context (issues, jira)

- `accounts.spec.ts` is failing on macOS
- `settings.spec.ts` rarely fails but it does, it's due to mouse moves and the scrollbar is showing up sometimes
- migrated `firmwareupdate.spec.ts`

## 🖤  Expectations to reach

CI must be green ✔️


valpinkman (dismissed):

playwright/models/FirmwareUpdateModal.ts
```diff
@@ -0,0 +1,23 @@
...
+  readonly readyCheckbox: Locator;
+  readonly downloadProgress: Locator;
+  readonly flashProgress: Locator;
+  readonly updateDone: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.page = page;
+    this.readyCheckbox = page.locator('data-test-id=firmware-update-ready-checkbox');
+    this.downloadProgress = page.locator('data-test-id=firmware-update-download-progress');
```
ggilchrist-ledger, line 15: Any need for these three?

playwright/models/Layout.ts
```diff
@@ -24,7 +24,7 @@ export class Layout {
   constructor(page: Page) {
     this.page = page;
 
-    // portfolio
+    // portfolio && accounts
     this.totalBalance = page.locator('data-test-id=total-balance');
 
     // drawer
```
ggilchrist-ledger, line 30: For future we should probably move Drawer into its own PO

src/renderer/modals/DisclaimerModal/index.js
```diff
@@ -114,7 +114,7 @@ class DisclaimerModal extends PureComponent<Props, State> {
             <Box horizontal justifyContent="flex-end" alignItems="center" style={{ flex: 1 }}>
               <Box horizontal alignItems="center" onClick={this.onSeedReady} style={{ flex: 1 }}>
                 <CheckBox
-                  id={"firmware-update-disclaimer-modal-seed-ready-checkbox"}
```
ggilchrist-ledger, line 117: Are you sure this ID isn't getting used for anything else (like styling)?

valpinkman: as a convention, we try to not use IDs for styling, but nice catch anyway, we should indeed check it out (even tho i'm 99% sure we don't need to)

playwright/models/Layout.ts
nabil-brn: Agreed, we'll do it step by step 👍

playwright/models/FirmwareUpdateModal.ts
nabil-brn: They're used by po itself and spec file too.

src/renderer/modals/DisclaimerModal/index.js
nabil-brn: Sure, I looked for the id in the whole code.

ggilchrist-ledger (approved):

valpinkman (approved):

PR #4537, opened 2021-12-13
LL-7773 - Manager v3 rebranding

OlivierFreyssinet:
## 🦒 Context (issues, jira)
[LL-7773]

This doesn't include the rebranding of the firmware update modal/drawer

[Figma for reference](https://www.figma.com/file/7lNG0G4IGF2kphCbef6Gmz/LLD-%2F-Manager?node-id=7%3A3275)

https://user-images.githubusercontent.com/91890529/145873158-a0024138-6f86-47cd-9f79-ffeb3953bc91.mov

## 🖤  Expectations to reach

[LL-7773]: https://ledgerhq.atlassian.net/browse/LL-7773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/renderer/modals/DisclaimerModal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    final: FinalFirmware;
+  };
+  goToNextStep: () => void;
+  onClose: () => void;
+  modelId: DeviceModelId;
+};
+
+type State = any;
+
+class DisclaimerModal extends PureComponent<Props, State> {
```
LFBarreto, line 31: do we need to keep it as a class component? :thinking: 
maybe could be rewritten as a function one with memo

src/renderer/components/TagsTabBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+      {tabs.map((tab, index) => {
+        const isSelected = index === selectedIndex;
+        const color = isSelected ? "palette.neutral.c100" : "palette.neutral.c80";
+        return (
+          <Tag
+            type="opacity"
+            key={index}
+            onClick={() => onSelectedChanged(index)}
+            active={isSelected}
+            style={{ marginLeft: index === 0 ? 0 : "8px" }}
```
LFBarreto, line 24: could you create a PR to fix the tag cursor :thinking: 
also we could use the proxified styled on the ui lib to control margin left inline with no style prop needed
(nb for the cursor you can actually target in styled components props things like "onClick" )

LFBarreto (approved):
LGTM minor issues mostly to be fixed on ui lib

src/renderer/components/TagsTabBar.tsx
ofreyssinet-ledger: Actually someone made similar a tab bar in UI repo since I wrote this code weeks ago so I'll use it instead

src/renderer/modals/DisclaimerModal/index.tsx
ofreyssinet-ledger: Yep indeed, shouldn't we stick to "not touching the logic" unless necessary for rebranding work though ?

ofreyssinet-ledger: In case some changes are made in the develop branches it will be easier to reconcile when we merge everything if the logic is modified as less as possible.

PR #4539, opened 2021-12-14
Implemented intermediate buy screen

VitaliiShpital:

https://user-images.githubusercontent.com/46756926/146033046-657b6421-9de1-4e21-8b27-9faea8c02be8.mp4

## 🖤  Expectations to reach


src/renderer/screens/exchange/index.js
```diff
@@ -1 +1,66 @@
...
+      <Box ff="Inter|SemiBold" fontSize={7} color="palette.text.shade100" id="exchange-title">
+        {t(tabs[activeTabIndex].header, { provider: provider.id })}
+      </Box>
+      <TabBar
+        defaultIndex={activeTabIndex}
+        tabs={tabs.map(tab => t(tab.title))}
+        onIndexChange={setActiveTabIndex}
+      />
+      <Card grow style={{ overflow: "hidden" }}>
+        <Component {...location?.state} {...props} provider={provider} />
```
JunichiSugiura, line 60: Does `undefined` or `null` spread?

src/renderer/screens/exchange/Buy/index.js
```diff
@@ -1,96 +1,102 @@
...
+              cryptoCount={10}
+              onClick={() => toggle("coinify")}
+              isActive={selected === "coinify"}
+            >
+              <CoinifySquare size={48} />
+            </SelectProvider>
+          </Box>
+          <Footer>
+            <ContinueButton primary disabled={!selected} onClick={onContinue}>
+              <Trans i18nKey="common.continue" />
```
JunichiSugiura, line 96: Why using `Trans` where line 74 is utilizing `t()` ?

JunichiSugiura (commented):
I have two questions.

ghost (commented):
Since there is no more history tab in the Buy & Sell screen, you could remove the related code (under `src/renderer/screens/exchange/History.js` as well as in the `static/i18n/en/app.json` file) to reduce dead code.

src/renderer/screens/exchange/index.js
VitaliiShpital: Hey! Thanks for you feedback! 
So if spread of `undefined` or `null` is used in an object then it'll just return empty object.
What I mean is:
`
const test = {...undefined}
// test is {}
`

If spread is used in array then you'll get `TypeError: undefined is not iterable` or `TypeError: null is not iterable`

src/renderer/screens/exchange/Buy/index.js
VitaliiShpital: Good catch! I was looking through the repo and found two different approaches so decided to test each. If I understand correctly <Trans> is still in the code since Component-based times and has to be refactored to use hook approach. 

Fixed, thanks.

VitaliiShpital:
Hey @achabrolin-ledger! Thanks for your feedback!
So I asked @IAmMorrow if I should remove all the history-related code from the repo or just remove this tab from a tablist. So @IAmMorrow told me to remove the tab as `History` will be reworked in the future. 

But yeah, since you think it's better to remove it entirely then ok. Removed it.

src/renderer/screens/exchange/index.js
VitaliiShpital: in any case, I made it more clear. Thanks

IAmMorrow:
Hello boosty team ! Thanks for waiting.

We have a few feedback on the implementation and some bug.

Apparently the Coinify widget has some container issues (it used to be full screen)
image

Users are stuck if they click on coinify, they can't go back and clicking on the menu again doesn't trigger a nav. We didn't provide on the figma any mean to come back to the first step, that's on us. We'll get that fixed.

VitaliiShpital:
Hey @IAmMorrow! Thanks for your feedback. Made a hotfix for Coinify widget's width

src/renderer/screens/exchange/index.js
```diff
@@ -1 +1,66 @@
...
+      <Box ff="Inter|SemiBold" fontSize={7} color="palette.text.shade100" id="exchange-title">
+        {t(tabs[activeTabIndex].header, { provider: provider.id })}
+      </Box>
+      <TabBar
+        defaultIndex={activeTabIndex}
+        tabs={tabs.map(tab => t(tab.title))}
+        onIndexChange={setActiveTabIndex}
+      />
+      <Card grow style={{ overflow: "hidden" }}>
+        <Component {...(location && location.state)} {...props} provider={provider} />
```
LFBarreto, line 60: preferably use explicit props instead of spreading them.
This could cause issues especially if you dont handle null or undefined values in the target component.
Also you will get a lot more props than you actually need this way, and your component might rerender for no apparent reason if its not memoized.
Lastly `state` has already been extracted from location on line 44 so if you want to save code lines hers might aswell use it directly 
```js 
{ ...(state || {}) } 
```

src/renderer/screens/exchange/Buy/SelectProvider.js
```diff
@@ -0,0 +1,59 @@
...
+import type { ThemedComponent } from "~/renderer/styles/StyleProvider";
+
+const Container: ThemedComponent<{}> = styled(Box).attrs(p => ({
+  alignItems: "center",
+  cursor: "pointer",
+  boxShadow: p.isActive ? `0 0 0 4px ${rgba(p.theme.colors.palette.primary.main, 0.25)}` : "",
+}))`
+  border: 1px solid ${p => p.theme.colors.palette.divider};
+  border-radius: 4px;
+  width: calc(50% - 12px);
```
LFBarreto, line 20: preferably use flex layout with `row-gap` to handle 50:50 layout with spacing in between

LFBarreto (changes requested):

src/renderer/screens/exchange/index.js
VitaliiShpital: Hey! Thanks for your feedback! I think I figured it out and fixed this props thing and made everything consistent so it should look obvious now

src/renderer/screens/exchange/Buy/SelectProvider.js
VitaliiShpital: I assume you meant `column-gap`. Yeah, reworked it. Thank you!

src/renderer/screens/exchange/index.js
VitaliiShpital: I've checked that all the target components are expecting `undefined` so it should be fine. Please tell me if you have any other concerns here

LFBarreto (approved):

PR #4559, opened 2021-12-20
LL-7135 - Fix triple dots in NFT gallery list mode

ghost:
## 🦒 Context (issues, jira)

[LL-7135](https://ledgerhq.atlassian.net/browse/LL-7135)

Right click on an NFT in grid mode inside a gallery should show a context menu.
Left click on the triple dots next to an NFT in list mode inside a gallery should show a context menu.

https://user-images.githubusercontent.com/86959861/146784921-d04bc983-7f29-404b-97ac-d8c4199a4148.mov

## 🖤  Expectations to reach

### DIFF
--- a/src/renderer/screens/nft/Gallery/TokensList/Item.js
+++ b/src/renderer/screens/nft/Gallery/TokensList/Item.js
@@ -12,6 +12,7 @@ import Image from "~/renderer/screens/nft/Image";
 import Skeleton from "~/renderer/screens/nft/Skeleton";
 import IconDots from "~/renderer/icons/Dots";
 import { useNftMetadata } from "@ledgerhq/live-common/lib/nft/NftMetadataProvider";
+import NFTContextMenu from "~/renderer/components/ContextMenu/NFTContextMenu";
 import { NFTViewerDrawer } from "~/renderer/drawers/NFTViewerDrawer";
 import { setDrawer } from "~/renderer/drawers/Provider";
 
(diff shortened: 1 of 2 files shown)


src/renderer/screens/nft/Gallery/TokensList/Item.js
```diff
@@ -47,9 +48,10 @@ type Props = {
   tokenId: string,
   id: string,
   mode: "grid" | "list",
+  withContextMenu?: boolean,
 };
 
-const Row = ({ contract, tokenId, id, mode, account }: Props) => {
+const Row = ({ contract, tokenId, id, mode, account, withContextMenu = false }: Props) => {
```
grsoares21, line 54: What do you think of using the isGrid value instead of using a new prop?
Since they have the same value (mode === "grid").

This way we isolate this knowledge about the context menu in this component and the parent doesn't need to know about it. (in grid mode = context menu with right click in the component, not in grid mode = context menu with left click in the three dots)

grsoares21 (changes requested):

src/renderer/screens/nft/Gallery/TokensList/Item.js
klambert-ledger: I actually don't think we should link directly features and design, even though I agree with you that _for the moment_ it's our only use case for this functionality. 

The context menu is actually the ability to right click on the NFT itself not on the triple dots. So we could cleary at some point decide that we want to display an NFT somewhere in grid mode with no context menu on it, and vice-versa, display multiple nfts in list mode that would also be right-clickable to display the links and not only through a left-click on the triple dots.

wdyt ?

grsoares21: Hmmm yeah, I see your point. Ok, let's go for it 👍

grsoares21 (approved):

PR #4591, opened 2021-12-30
LL 8837 - NFT image to pure component

ghost:
## 🦒 Context (issues, jira)
[LL-8837](https://ledgerhq.atlassian.net/browse/LL-8837)

## 🖤  Expectations to reach

### DIFF
--- a/src/renderer/screens/nft/Gallery/TokensList/Item.js
+++ b/src/renderer/screens/nft/Gallery/TokensList/Item.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { useMemo, useCallback } from "react";
+import React, { useMemo, useCallback, memo } from "react";
 import { Trans } from "react-i18next";
 import styled from "styled-components";
 import type { Account } from "@ledgerhq/live-common/lib/types";
@@ -42,6 +42,14 @@ const Dots: ThemedComponent<{}> = styled.div`
   }
 `;
 
+const TitleContainer: ThemedComponent<{}> = styled(Text)`
+  overflow: hidden;
+  text-overflow: ellipsis;
+  display: -webkit-box;
+  -webkit-line-clamp: 1;
+  -webkit-box-orient: vertical;
+`;
+
 type Props = {
   account: Account,
   contract: string,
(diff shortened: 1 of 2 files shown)


src/renderer/screens/nft/Gallery/TokensList/Item.js
```diff
@@ -51,7 +59,7 @@ type Props = {
   withContextMenu?: boolean,
 };
 
-const Row = ({ contract, tokenId, id, mode, account, withContextMenu = false }: Props) => {
+const NftCard = ({ contract, tokenId, id, mode, account, withContextMenu = false }: Props) => {
   const { status, metadata } = useNftMetadata(contract, tokenId);
   const { nftName } = metadata || {};
   const show = useMemo(() => status === "loading", [status]);
```
grsoares21, line 65: Does this useMemo here really makes sense? I doesn't seem to me that this comparison would make much of a difference in permormance / re-renders.

grsoares21 (approved):
Just a small question unrelated to the PR, out of curiosity.

src/renderer/screens/nft/Gallery/TokensList/Item.js
klambert-ledger: The issue this useMemo is trying to solve is that right now, each time we have a new synchronisation, we create a new array of NFTs in the `Account`, which is supposed to preserve the original ref of each of its element if they're identical to the previous synchronisation. 

It's happening here: https://github.com/LedgerHQ/ledger-live-common/blob/348b43649cdad670daafb2abe9d7cdade3251d34/src/families/ethereum/synchronisation.ts#L169
Using this method:
https://github.com/LedgerHQ/ledger-live-common/blob/master/src/bridge/jsHelpers.ts#L121

But it turns out it doesn't work as expected for now (you can test it by clicking the `✅  Synchronized` while being in a gallery, it will probably make you images blink). 
That's why we're trying to switching a PureComponent for the image, and memoizing the NftCard is also here to try to mitigate the issue as, on paper, it should almost never re-render since it's a pretty performance consuming component.

PR #4626, opened 2022-01-11
LL-8867 Add NFT pan and zoom modal

grsoares21:
## 🦒 Context (issues, jira)
[LL-8867]

pan-and-zoom-nft-pr

## 🖤  Expectations to reach

[LL-8867]: https://ledgerhq.atlassian.net/browse/LL-8867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -100,6 +100,7 @@
     "react-lottie": "^1.2.3",
     "react-markdown": "^4.3.0",
     "react-motion": "^0.5.2",
+    "react-prismazoom": "^2.1.1-rc1",
```
LFBarreto, line 103: could you use v3 instead not a fan of locking to rc versions for prod :sweat_smile:

LFBarreto (changes requested):

package.json
LFBarreto: also correct me if i'm wrong but this doesn't seem so hard to do natively on desktop the whole mousewheel to zoom and transform scale :sweat_smile: if you do so move that logic towards the ui lib so we can reuse it

grsoares21: Great point on the version 😅, bumped it to 3.0.0.

Concerning doing it ourselves. I don't think it'd be too hard, but still will take a bit of time. Will keep that in mind to maybe tackle it as technical task when I have some time.

juan-cortes (approved):

LFBarreto (approved):

PR #4659, opened 2022-01-19
LL-9028 Handle new lastSeenDevice event on DeviceActions

juan-cortes:
> There's a less invasive proposal on live-common that will run the firmware check without blocking, here https://github.com/LedgerHQ/ledger-live-common/pull/1637

## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-9028

https://user-images.githubusercontent.com/4631227/150198199-50f384f0-8a8d-4717-850a-a0b802354591.mov

Starting from 21.27.0 of live-common we are emitting (in some cases) a new event in order to update more granular the last seen device information stored by the app. We use this information, alongside the `latestFirmware` available for this device, both for analytics and for handling the display of the firmware update banner.

This means we could now detect an outdated device without the need for the user to go into the manager.
Driving more userflow into the latest version. As you can see in the video, I'm being shown the firmware update banner after going through a receive flow.

## 🖤  Expectations to reach

  - [x] a specific test planned is defined here

### On a device that is on the latest firmware on the selected provider
- On a fresh LLD install without connecting to the manager start a device flow
- Make sure the device is on the dashboard since this is only triggered before the `openApp` case
- After completing the flow (add accounts for instance) open your app.json file
- It is expected that there will be some data for `lastSeenDevice`

### On a device that has an available firmware update on the selected provider
- After starting any device flow, it is expected that the banner will automatically appear

### DIFF
--- a/src/renderer/components/DeviceAction/index.js
+++ b/src/renderer/components/DeviceAction/index.js
@@ -6,7 +6,7 @@ import { connect } from "react-redux";
 import type { Device, Action } from "@ledgerhq/live-common/lib/hw/actions/types";
 import { OutdatedApp, LatestFirmwareVersionRequired } from "@ledgerhq/live-common/lib/errors";
 import { getCurrentDevice } from "~/renderer/reducers/devices";
-import { setPreferredDeviceModel } from "~/renderer/actions/settings";
+import { setPreferredDeviceModel, setLastSeenDeviceInfo } from "~/renderer/actions/settings";
 import { preferredDeviceModelSelector } from "~/renderer/reducers/settings";
 import type { DeviceModelId } from "@ledgerhq/devices";
 import AutoRepair from "~/renderer/components/AutoRepair";
@@ -88,6 +88,7 @@ const DeviceAction = <R, H, P>({
     allowManagerRequestedWording,
     requestQuitApp,
     deviceInfo,
+    latestFirmware,
     repairModalOpened,
     requestOpenApp,
     allowOpeningRequestedWording,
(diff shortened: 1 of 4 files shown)


src/renderer/reducers/settings.js
```diff
@@ -275,7 +275,10 @@ const handlers: Object = {
     { payload }: { payload: { lastSeenDevice: DeviceModelInfo, latestFirmware: any } },
   ) => ({
     ...state,
-    lastSeenDevice: payload.lastSeenDevice,
+    lastSeenDevice: {
+      ...(state.lastSeenDevice || {}),
+      ...payload.lastSeenDevice,
```
LFBarreto, line 280: could you double check that we don't call this dispatch somewhere with potential undefined :grimacing: 
or maybe fail safe it the same way as the line above `|| {}`

LFBarreto (approved):
just a extra check
LGTM :rocket:

src/renderer/reducers/settings.js
juan-cortes, line 280:
```suggestion
      ...(payload.lastSeenDevice || {}),
```

juan-cortes: you're right, won't do any harm to add it

src/renderer/reducers/settings.js
```diff
@@ -275,7 +275,10 @@ const handlers: Object = {
     { payload }: { payload: { lastSeenDevice: DeviceModelInfo, latestFirmware: any } },
   ) => ({
     ...state,
-    lastSeenDevice: payload.lastSeenDevice,
+    lastSeenDevice: {
```
tlaforge-ledger, line 278: I would prefere this one : 

Object.assign({}, state.lastSeenDevice, payload.lastSeenDevice)

If one element is undefined it will not affect the object you are building.
But it you prefere this approach then you can resolve.

src/renderer/components/DeviceAction/index.js
```diff
@@ -123,6 +124,17 @@ const DeviceAction = <R, H, P>({
     }
   }, [dispatch, modelId, preferredDeviceModel]);
 
+  useEffect(() => {
+    if (deviceInfo) {
+      const lastSeenDevice = {
+        modelId: device.modelId,
```
tlaforge-ledger, line 130: is device always set when we have deviceInfo ? Or do device?.modelId needed ?

tlaforge-ledger (changes requested):

src/renderer/components/DeviceAction/index.js
juan-cortes: If we have deviceInfo we should have device as well, it's fine

tlaforge-ledger (approved):

PR #4692, opened 2022-01-30
Adding arabic to lang list

KernelCode:
##  Adding arabic to lang list

Screenshot from 2022-01-30 10-40-50

### DIFF
--- a/src/config/languages.js
+++ b/src/config/languages.js
@@ -21,9 +21,10 @@ export const allLanguages = [
   "sv",
   "tr",
   "zh",
+  "ar",
 ];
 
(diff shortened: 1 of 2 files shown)


juan-cortes (changes requested):
In order for this to be merged it would need to add the language to the experimental languages, not the production stable ones like you propose. So using the `allLanguages` array instead. Then you'd need to enable the experimental flag in the settings in order to use it. 

I believe languages with RTL writing systems will require a lot of layout rework in order to be considered _stable_ for production though, so I don't expect them coming out of experimental in the short term.

PR #4723, opened 2022-02-07
live-common 21.32.2

henrily-ledger:
It's to test algorand fix and build, put as a draft first


haammar-ledger:
Looks like `cryptoassets.md` snapshot needs to be regen'd. @gre please how is that done?

gre:
wait we'll need a 21.32.3 as .2 is no go (ethereum tx issue)

henrily-ledger:
no go wait for 21.33.3

PR #4747, opened 2022-02-16
comment on pr instead of hash for checklist

valpinkman:
## 🦒 Context (issues, jira)

To prevent the comment from being erased, comment on PR number instead of commit hash (lost in a rebase or force push)

## 🖤  Expectations to reach

flagging is better than a very long PR. so prefer Option 1 if Option 2 takes more than a sprint -->


elbywan (approved):
👌

PR #4748, opened 2022-02-16
LIVE-883 LIVE-885 Support tools/swap/plugins on manager app list

juan-cortes:
## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LIVE-883
https://ledgerhq.atlassian.net/browse/LIVE-885

In order to continue improving the support of apps on the manager and offering specific rendering depending on the application type this PR adds a specific action for the exchange app, specific redirect for tools as well as a minor redesign of the apps table to account for the Learn More options that are now not limited to currency apps not supported in Live. This has been validated with design and we are all very happy with it as a temporary UI until we redo de manager.

image
image

## 🖤  Expectations to reach

  - [x] a specific test planned is defined here
  - [x] this PR is covered by automatic UI test

## How to test this
- Access the manager and look for the rendering changes for the different types of apps
- Look for installed and uninstalled versions of these types
- Supported currency apps (bitcoin) should show an Add Account CTA if installed
- Not supported currency apps (cardano) should show a Learn More CTA that redirects to the crypto assets page or specific page for currency
- Tools (recovery) should show a Learn More CTA that redirects to generic support url
- Swap apps (Exchange) should show a Swap CTA when installed that redirects to the swap feature
- Plugins (Paraswap) should show a Go to App CTA when installed that redirects to discover.
(Probably good to test in both themes)


src/config/urls.js
```diff
@@ -64,6 +64,7 @@ export const urls = {
     "https://support.ledger.com/hc/en-us/articles/360021039173-Choose-network-fees?docs=true",
   recipientAddressInfo:
     "https://support.ledger.com/hc/en-us/articles/4404389453841-Receive-crypto-assets?docs=true",
+  managerAppLearnMore: "https://support.ledger.com/hc/en-us/categories/4404376139409?docs=true",
```
haammar-ledger, line 67: Is the link correct? It's valid but it directs me to the generic welcome page of Ledger Docs.

ghost (approved):
LGTM apart from a question on the support link.

src/config/urls.js
juan-cortes: It's what I got, yes 🤷🏼

src/renderer/screens/manager/AppsList/AppActions.js
```diff
@@ -169,27 +222,26 @@ const AppActions: React$ComponentType<Props> = React.memo(
...
-              ) : (
+              ) : hasSpecificAction ? (
                 <Button
-                  color={canAddAccount ? "palette.primary.main" : "palette.text.shade40"}
-                  inverted
+                  color={"palette.primary.main"}
                   style={{ display: "flex", backgroundColor: "rgba(0,0,0,0)" }}
                   fontSize={3}
-                  onClick={onNavigateToPlatform}
+                  onClick={() => onNavigateTo(type)}
```
ofreyssinet-ledger, line 230: Could this this simply be `onClick={onNavigateTo}` with `onNavigateTo` being declared like this and using `type` directly instead of as an argument ?
```ts
const onNavigateTo = useCallback(() => {
    /* here do what you have to do and use `app.type` directly*/
}, [app.name, history, app.type]);
```
This arrow func declared in the props will create unnecessary renders no ?

OlivierFreyssinet (approved):

src/renderer/screens/manager/AppsList/AppActions.js
juan-cortes: You are right :)

PR #4754, opened 2022-02-18
bugfix/LIVE 964 Market remove live-compatible filter for now

LFBarreto:
- (Market): remove filter by live compatible for now

## 🖤  Expectations to reach

### DIFF
--- a/src/renderer/Default.js
+++ b/src/renderer/Default.js
@@ -175,9 +175,6 @@ export default function Default() {
                         <Route path="/settings" render={props => <Settings {...props} />} />
                         <Route path="/accounts" render={props => <Accounts {...props} />} />
                         <Route path="/card" render={props => <Card {...props} />} />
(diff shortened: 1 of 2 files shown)


src/renderer/Default.js
```diff
@@ -220,6 +217,9 @@ export default function Default() {
                           render={props => <MarketCoinScreen {...props} />}
                         />
                         <Route path="/market" render={props => <Market {...props} />} />
+                        <FeatureToggle feature="learn">
+                          <Route path="/learn" render={props => <Learn {...props} />} />
+                        </FeatureToggle>
```
ofreyssinet-ledger, line 222: is this supposed to be in this PR?

OlivierFreyssinet (approved):

src/renderer/Default.js
LFBarreto: it fixes a regression on feature flag :+1:

PR #4647, opened 2022-01-17
LL-8941 Release USB hold after 5 seconds

juan-cortes:
image

## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-8941

## 💻  Description / Demo (image or video)

This is just a bump of the singleton transport in order to benefit from the USB connection release introduced in 2.23.0
When we manage to get this merged it would allow https://github.com/MetaMask/metamask-extension or any third party application relying on the WebHID transport (or the node-hid ones for that matter) to access a device even after LLD takes a hold of the connection.

## 🖤  Expectations to reach

PR must pass CI, merge develop if conflicts, do not force push. Thanks!

- **on QA**: at least one of these two checkboxes must be checked:
  - [x] a specific test planned is defined here
  - [ ] this PR is covered by automatic UI test
- **on delivery**: at least one of these two checkboxes must be checked: 
  - [ ] Option 1: **no impact**: The changes of this PR have ZERO impact on the userland (invisible for users)
  - [ ] Option 2: **atomic delivery**: the changes is atomic and complete (no partial delivery)

PR must pass CI, merge develop if conflicts, do not force push. Thanks!

## 🖤  How to test
First a foreword. The singleton connection ownership issue was only present as far as I know in Mac at this point, meaning that testing should (feel free to correct me) focus mostly on this. It is important to understand that there's a 5 second timeout before we release the control over the USB connection, 5 seconds after the last APDU exchange.

### I'm too quick case
- Launch LLD
- Access a device flow, any flow that requires APDU exchanges (manager, receive, send, add accounts)
- After your last interaction quickly switch to Chrome on https://repl.ledger.tools/ and try to connect to the device
- If done under 5 seconds it's expected to get an error message `NotAllowedError: Failed to open the device.`
- Try connecting after a few second, it is expected the connection succeeds.

### Normal case
- Launch LLD
- Access the manager
- Enter an app in your device
- It is expected that the LLD shows a screen saying another application is open in your device
- Allow a few seconds for the timeout to be reached.
- Switch to Chrome on https://repl.ledger.tools/ and try to connect to the device.
- It should connect

PR #4764, opened 2022-02-21
LIVE-590 POC Testing autodisconnect mechanism

juan-cortes:
> **This can't be merged. It needs first the ledgerjs merge, then the live-common changes and bump, then this pr updated, but this is the only way to test it.**

## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LIVE-590

For context, this is a continuation of the initial task that attempted to release the hold from the singleton transport after a given amount of time. The problem that we initially had with that approach, problem that was identified by our wonderful QA @alalmi-ledger , was that disconnecting after a hard timeout that only relied on apdu exchanges to reset the timeout didn't account for the fact that we have some long running tasks inside `withDevice` calls.

`withDevice` will open a connection to a device and will keep that connection open unless unmounted or specifically disconnected. In certain flows such as adding accounts or sending transactions we would reach the timeout before completing the flow, triggering a disconnect and breaking said flow.

In https://github.com/LedgerHQ/ledgerjs/pull/797 we introduced the concept of freezing the auto-disconnect temporarily while are still inside a `withDevice` call. This pr leverages that functionality, while relying on a commit for live-common that actually uses that new exposed static method [setAllowAutoDisconnect](https://github.com/LedgerHQ/ledger-live-common/commit/b09ed14715d79c52dd440ef900168e415c10e951) to enable/disable the feature.

## How to test this
Same as the original PR, the context of this is still making sure that we don't keep control of the HID connection after we are done with the device. https://github.com/LedgerHQ/ledger-live-desktop/pull/4647 Just be sure that the failing tests (long add accounts on Cosmos, or a complex send) also work.

The device should be connectable after 5 seconds when we are no longer interacting with it from https://repl.ledger.tools/ on chrome.

In particular, check what happens with flows if we leave the device idle for more than 5 seconds after the action is completed. I noticed the add accounts flow will throw an error when we disconnect even though we are no longer interacting with the device. Perhaps other flows have this same behavior and we'd need to adapt the UX to it (ie, if the action is done, ignore the error basically)

## 🖤  Expectations to reach


chabroA:
@juan-cortes Hey 👋 Since this PR is not ready to be merged, wouldn't it make sense to convert it to draft, until all the requirements and updates you mentioned are done?

cf. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

juan-cortes:
Closing for now until we have the live-common version making use of it

PR #4794, opened 2022-02-28
Cosmos JS experimental

gre:
branch is used to build a WIP of https://github.com/LedgerHQ/ledger-live-common/compare/cosmos-js

**this is still a WIP**


ghost (changes requested):
Unexpected UI snapshots diff in `tests/specs/swap.spec.ts-snapshots` :confused: with weird padding at the top.
This has nothing to do with that PR, it happened before and seems to still happen randomly.
I think it will screw up UI tests on other PRs if merged...

PR #4828, opened 2022-03-10
Feat  LIVE-1630 lld update ctas style

LFBarreto:
## 🦒 Context (issues, jira)
[LIVE-1630]

- Updated market buy/swap ctas style
- Updated style of account page

## 💻  Description / Demo (image or video)

https://user-images.githubusercontent.com/11752937/157708489-c877db1b-73f9-4303-9a52-25fcfacbcd2d.mp4

## 🖤  Expectations to reach

PR must pass CI, rebase develop if conflicts. Thanks!

- **on QA**: at least one of these two checkboxes must be checked:
  - [ ] a specific test planned is defined on Jira
  - [ ] this PR is covered by automatic UI test
- **on delivery**: at least one of these two checkboxes must be checked: 
  - [ ] Option 1: **no impact**: The changes of this PR have ZERO impact on the userland (invisible for users)
  - [ ] Option 2: **atomic delivery**: the changes is atomic and complete (no partial delivery)

[LIVE-1630]: https://ledgerhq.atlassian.net/browse/LIVE-1630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4846, opened 2022-03-16
feat(account): improve the empty state by displaying options in the header

imqdee:
## 🦒 Context (issues, jira)

Until now, the header in the account page for an empty account wasn't
rendered. Because of this, walletconnect button wasn't rendered, you needed
some tokens to see it appears. This commit fixes this, from now, walletconnect
button and receive/buy buttons would be rendered for any empty accounts.
As it is not required to own tokens to use walletconnect, this modification makes
sense to me. Regarding the receive/buy buttons, I think it's better to have them
always available, even for empty accounts (we missed a business opportunity by
hidding the buy button in this case).

## 🖤  Expectations to reach

## 🧪 How to test

**Scenario 1**
- Ensure you have an empty Polygon or BSC account
- Go to the account page of your empty account
- The buy and the receive buttons must be displayed in the header

**Scenario 2**
- Ensure you have an empty Ethereum account
- Go to the account page of your empty account
- The buy, receive, and walletconnect buttons must be displayed in the header

### DIFF
--- a/src/renderer/screens/account/AccountHeaderActions.js
+++ b/src/renderer/screens/account/AccountHeaderActions.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { useCallback } from "react";
+import React, { useCallback, useMemo } from "react";
 import { compose } from "redux";
 import { connect } from "react-redux";
 import { withTranslation, Trans } from "react-i18next";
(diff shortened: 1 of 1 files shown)


thomasrogerlux (approved):
Makes sense

LFBarreto (changes requested):
Same will conflict with #4828
once released rebase and fix issues

PR #4848, opened 2022-03-17
LIVE-590 Remove device connection hold after timeout

juan-cortes:
## 🦒 Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LIVE-590

After literally two months of iterations and releases this is the first time the PR for testing this feature is actually pointing at released versions of the dependencies. The code for live-common `v21.35.0` https://github.com/LedgerHQ/ledger-live-common/pull/1735 made it to the last version and the changes at the transport level made it to `v6.26.0` https://github.com/LedgerHQ/ledgerjs/pull/797 (possible earlier, but they're definitely there)

Without this PR the functionality of the USB hold should remain unchanged. This PR introduces a locking/unlocking mechanism that is triggered when we start/end a `withDevice` code block. From my testing this allows us to not be able to connect while a long running task is happening such as an add account flow but as soon as the flow finishes +5s the hold is released and we no longer prevent users from connecting.

This was mostly targeting metamask users on mac in the first place.

## 🖤  Expectations to reach

## How to test this (copied from https://github.com/LedgerHQ/ledger-live-desktop/pull/4764)

Same as the original PR, the context of this is still making sure that we don't keep control of the HID connection after we are done with the device. Just be sure that the failing tests (long add accounts on Cosmos, or a complex send) also work. The device should be connectable after 5 seconds when we are no longer interacting with it from https://repl.ledger.tools/ on chrome.

In particular, check what happens with flows if we leave the device idle for more than 5 seconds after the action is completed. I noticed the add accounts flow will throw an error when we disconnect even though we are no longer interacting with the device. Perhaps other flows have this same behavior and we'd need to adapt the UX to it (ie, if the action is done, ignore the error basically)


tlaforge-ledger (approved):

thomasrogerlux (approved):
Code looks good

PR #4849, opened 2022-03-17
Polygon: Uses balance from backend even if no txs is in explorer

gre:
## 🦒 Context (issues, jira)

LIVE-1102

fixes a workaround where some Polygon accounts wouldn't even see their balance when they receive only "internal transactions" that our explorer don't support.

https://github.com/LedgerHQ/ledger-live-build/actions/runs/2000520750

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@ledgerhq/hw-transport-http": "6.24.1",
     "@ledgerhq/hw-transport-node-hid-singleton": "6.24.1",
     "@ledgerhq/ledger-core": "6.14.5",
-    "@ledgerhq/live-common": "https://github.com/LedgerHQ/ledger-live-common.git#develop",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "@ledgerhq/hw-transport-http": "6.24.1",
     "@ledgerhq/hw-transport-node-hid-singleton": "6.24.1",
     "@ledgerhq/ledger-core": "6.14.5",
-    "@ledgerhq/live-common": "https://github.com/LedgerHQ/ledger-live-common.git#develop",
+    "@ledgerhq/live-common": "https://github.com/LedgerHQ/ledger-live-common.git#4e52a29ac60f8f95247a09b61e7454ecc08044cc",
```
thomasrogerlux, line 52: I think this crash on develop

PR #4881, opened 2022-03-31
[LIVE-1804] Add User-agent as http header

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LIVE-1804


thomasrogerlux (approved):
Code looks good

thomasrogerlux:
Keep in mind checking node env equal production will match the `production` env but also `staging` env, I don't know if we want that

ghost (approved):

PR #4946, opened 2022-04-26
Optimize lodash usage (do not import the root)

gre:
This fixes the convention we had everywhere in the code in the usage of lodash: we must not import _ from "lodash" but use the functional imports.

if this pass the CI, it shall be good.


juan-cortes (approved):
It passes CI, so ✔️

thomasrogerlux (approved):

